PR #121, opened 2016-10-06
Load GLEW package only for MVSC or MINGW.

nmellado:


Zouch:
30 whiplashes if that does not work on windows !

PR #133, opened 2016-11-21
Use gl3w for every platform.

Zouch:
Needs to be bulletproofed on mac.


louen:
So, we be mergin' soon ?

louen:
To merge this we need to use External project to retrieve gl3w and not rely on EXECUTE commands on CMake for it to work on windows.

PR #141, opened 2017-02-10
Add a warning as error option

### LINKED ISSUE #125: Consider using -Werror to enforce warnings
Warning-as-error is good practice !

louen:
Fix the only warning in main code
Fixes #125


louen:
I believe this would work in Windows and MacOS, but that would be nice to check.

PR #142, opened 2017-02-23
Val assets

louen:
This PR includes several improvement to file loading, and fixes nasty bugs with gizmos.

Here's a breakdown of all the changes
* Gizmo fixes ( 5f793b7 and ccb1ff1) fixed two bugs with gizmos.
The first one appeared in release builds due to Eigen optimizations, and the second one prevents the gizmo render objects from being picked, thus avoiding a crash.

* File loading from plugins ( 639a441)
The behaviour of the file loading function has been modified, so that if assimp fails to load the scene file it still gets passed on to the other plugins. This allows to have specialized files meant to be understood by a specific plugin with its own loading facility. Because of that the list of extensions in the "open file" dialog box has been disabled. A much better solution would be to query each plugin for its allowed extensions.
Moreover, the generic "file manager" has been modified so that it can write to an arbitrary stream, not just a 'fstream`.

* Point-triangle distance query overhaul and mesh utilities :
a24d7a5 added a few basic mesh utility functions. They are not used in the engine yet but I needed them in my own plugins.
5730761 and  6cfb13e refactored the point-triangle distance query function with a cleaner output structure.

* Explicit plugin load and ignore (3e7b66a)
This  commit adds two options to the command line :  `-l` and `-i` which modify plugin loading behavior
If `-l` options are given, only the specified plugins (whose name is after `-l` ) are loaded. Otherwise all plugins are loaded (the default behaviour is unchanged).
`-i` allows to prevent a plugin to be loaded (it has precedence over `-l`)


louen:
Tell @mauriziokovacic  about it :stuck_out_tongue:

PR #145, opened 2017-03-02
Cmake improvements

nmellado:


CMakeLists.txt
```diff
@@ -17,6 +17,20 @@ option(RADIUM_WARNINGS_AS_ERRORS    "Treat compiler warning as errors" OFF)
...
   set( CMAKE_BUILD_TYPE Debug )
 endif()
+
+# guard against in-source builds (source: Eigen)
+if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
+  message(FATAL_ERROR "In-source builds not allowed. Please make a new directory (called a build directory) and run CMake from there. You may need to remove CMakeCache.txt. ")
+endif()
+
+# guard against bad build-type strings (source: Eigen)
+string(TOLOWER "${CMAKE_BUILD_TYPE}" cmake_build_type_tolower)
```
louen: There is a related piece of code in CompileFlags.cmake (which sets the build type to debug if it is not in the three allowed. )

CMakeLists.txt
```diff
@@ -17,6 +17,20 @@ option(RADIUM_WARNINGS_AS_ERRORS    "Treat compiler warning as errors" OFF)
 if ( NOT CMAKE_BUILD_TYPE )
   set( CMAKE_BUILD_TYPE Debug )
 endif()
+
+# guard against in-source builds (source: Eigen)
```
louen: Is there a reason why we don't want in-source builds ?

CMakeLists.txt
```diff
@@ -122,3 +136,39 @@ add_subdirectory(Applications)
...
+################################################################################
+
+find_package(Doxygen)
+if(DOXYGEN_FOUND)
+  # get changeset id
+  find_package(Git)
+  if(GIT_FOUND)
+    execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      OUTPUT_VARIABLE GIT_CHANGESET)
```
louen: This variable is actually pretty useful to have even outside of doxygen context (e.g. for Version.cpp)

Can you get it out of the if() ?

CMakeLists.txt
louen: OK, sold :)

PR #148, opened 2017-03-03
Add Docs directory in Doxygen configuration, so the markdown pages ar…

nmellado:
…e now automatically included


nmellado:
Please do not delete this branch, we can use it for all documentation related commits that are expected to come in the upcoming weeks to set the doc configuration.

louen:
I think we should first update the docs in the docs folder.

nmellado:
No problem

PR #149, opened 2017-03-07
String test and asset loading fix

louen:
This PR includes two commits :
* a unit test  for `Core::StringUtils` and a fix for the bug it helped to catch
* A fix to asset loading which would give a default name to unnamed meshes in assimp, which could cause problems for skinning setup.

### DIFF
--- a/src/Tests/CoreTests/String/StringTest.hpp
+++ b/src/Tests/CoreTests/String/StringTest.hpp
@@ -1,0 +1,80 @@
+#ifndef RADIUM_STRINGTESTS_HPP_
+#define RADIUM_STRINGTESTS_HPP_
+
+
+#include <Tests/CoreTests/Tests.hpp>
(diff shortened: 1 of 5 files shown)


src/Tests/CoreTests/String/StringTest.hpp
```diff
@@ -0,0 +1,80 @@
...
+            int v1 = sprintf(buffer, format, 42, this, 3.14);
+            int v2 = stringPrintf( str, format, 42, this, 3.14);
+            RA_UNIT_TEST( v1 == v2 && str == std::string(buffer), "String printf");
+
+
+
+
+
+
+
```
nmellado: A specific reason for these white lines ?

louen: Not really

louen:
Can I haz PR now ?

Issue #154, opened 2017-03-08
Clear the Engine


hoshiryu:
It would be nice to have a clean way to clear the Engine.
For example, if one wants to reset the Engine without having to relaunch the app (like a "new scene") or reload a complete scene.
Thus, destruction signals have to be managed correctly ( see issue #153 ), as for the component unregistration from the systems.

louen:
As I said to @hoshiryu , the implementation of this function should be nothing but removing all entities from the engine, and leave the destruction order handle resource deletion  correctly.

Issue #153, opened 2017-03-08
Deletion Order Problem


hoshiryu:
When deleting an Entity through the EntityManager, the signal "fireEntityDestroyed" is fired by the EntityManager, thus removing the ItemEntry in the EntityTreeModel (and its children).
When the Entity's Components are destroyed (which happens at the end of the Entity destructor), the signal "fireRenderObjectRemoved" is fired by their destructor through the RenderObjectManager, thus trying to remove the corresponding ItemEntry from the EntityTreeModel while they are not present anymore.
This triggers an assert in ItemModel::removeItem().

Who should fire the signals? in which order?

hoshiryu:
Another issue
When deleting an Entity while one of its RenderObjects has a gizmo attached, the SignalManager locks its mutex before processing the corresponding signals.
One of these signals is "selectionChanged" of the SelectionManager, which the MainWindow binds to its slot "onSelectionChanged", then triggerring the signal "selectedItem", which the MainWindow binds to the GizmoManager slot "setEditable" finally calling GizmoManager::spawnGizmo().
But this method resets the gizmo, which deletes its RenderObject, thus calling the SignalManager to process the adequate signals.
And we got an infinite mutex lock !!!

PR #157, opened 2017-03-09
Delete objects

louen:
This PR has two commits : one is a minor fix to Cmake messages, and the other implements correct object deletion order (#153) and a clear scene button and method(#154)


src/Core/Index/IndexMap.hpp
```diff
@@ -45,6 +45,11 @@ class IndexMap {
     inline Index insert( const T& obj );                    // Insert a object in the IndexMap. Return an invalid index if the object is not inserted.
     inline bool  insert( const T& obj, Index& idx );        // Insert a object in the IndexMap. Return true if the object is inserted, and store it in idx.
 
+    template<typename... Args>
+    inline Index emplace( const Args&&... args );             // Construct an object in place in the IndexMap. Return an invalid index if the object is not inserted.
+    template<typename... Args>
+    inline bool  emplace(  Index& idx, const Args&&... args ); // Construct an object in place in the IndexMap. Return true if the object is inserted, and store it in idx.
+
```
louen: This was required to use the map with `std::unique_ptr`

src/Core/Index/IndexMap.inl
```diff
@@ -116,7 +145,7 @@ inline bool IndexMap< T >::remove( const uint i ) {
 /// ACCESS
 /// ===============================================================================
 template < typename T >
-inline T IndexMap< T >::at( const Index& idx ) const {
+inline const T& IndexMap< T >::at( const Index& idx ) const {
     typename std::deque< Index >::const_iterator it = std::find( m_index.begin(), m_index.end(), idx );
     CORE_ASSERT( it != m_index.end(), "Index not found" );
     return m_data.at( it - m_index.begin() );
```
nmellado: Would be cleaner to use std::distance here, no ?
http://en.cppreference.com/w/cpp/iterator/distance

src/Core/Index/IndexMap.hpp
nmellado: ok.

src/Core/Index/IndexMap.inl
```diff
@@ -301,6 +330,17 @@ inline T& IndexMap< T >::operator[]( const uint i ) {
...
+inline const T& IndexMap< T >::operator[]( const Index& idx ) const {
+    return at( idx );
+}
+
+
+
+template < typename T >
+inline const T& IndexMap< T >::operator[]( const uint i ) const{
+    return at( i );
+}
```
nmellado: Why do we need a const uint version of this function ?
I sounds better to me to stick to Index, by adding an implicit constructor or cast operator.
By this mean, we know that Index will always be used, without any overhead, which is better for testing and monitoring.

src/Engine/Managers/EntityManager/EntityManager.cpp
```diff
@@ -128,5 +113,18 @@ namespace Ra
                 m_entities[i]->swapTransformBuffers();
             }
         }
+
+        void EntityManager::deleteEntities()
+        {
+            std::vector<uint> indices;
```
nmellado: suggestion, add:
`indices.reserve(m_entitites.size();)`

src/Engine/RadiumEngine.hpp
```diff
@@ -10,6 +10,7 @@
 #include <vector>
 
 #include <Core/Log/Log.hpp>
+#include <Core/Utils/Singleton.hpp>
```
nmellado: Why is this needed ?

louen: Actually it's not needed.

louen:
As for the  comments to the Index class I wholeheartedly agree with them, but it would be more sensible to plan for its refactoring in a separate issue.

PR #160, opened 2017-03-10
Fix openmp parralelism for linear blending

hoshiryu:


nmellado:
Right, indeed...
So, what do we do ? Shall we use the exact same strategy in both places ?

louen:
Exactly. See the implementation of `DualQuaternionSkinning.inl`

PR #161, opened 2017-03-10
Plugin action

hoshiryu:
Waiting for private plugins in @louen repo to handle the feature?


Plugins/Skinning/src/SkinningPlugin.cpp
```diff
@@ -98,6 +158,33 @@ namespace SkinningPlugin
         CORE_ASSERT( m_current, "should be disabled" );
         CORE_ASSERT( newType >= 0 && newType < 3, "Invalid Skinning Type" );
         m_current->setSkinningType( SkinningComponent::SkinningType( newType ) );
+        switch (newType) {
+        case 0:
+        {
+            m_actionLBS->setIcon(QIcon("../../Assets/Images/LB_on.png"));
```
nmellado: Would be safer to use the Qt resource system to avoid explicit path to icons in the applications.
http://doc.qt.io/qt-5/resources.html

Plugins/Skinning/src/SkinningComponent.hpp
```diff
@@ -1,7 +1,13 @@
...
 #define SKINPLUGIN_SKINNING_COMPONENT_HPP_
 
-#include <SkinningPlugin.hpp>
+#include <Core/CoreMacros.hpp>
+/// Defines the correct macro to export dll symbols.
+#if defined  Skinning_EXPORTS
+    #define SKIN_PLUGIN_API DLL_EXPORT
+#else
+    #define SKIN_PLUGIN_API DLL_IMPORT
+#endif
```
louen: Why is this here ? This should be in SkinningPlugin.hpp ...

hoshiryu: Gonna refactor it

louen:
Now rebased on master at #170

Issue #167, opened 2017-03-16
Allow asserts with streams interface


louen:
I would like to be able to do

` CORE_ASSERT( i == 0, "Variable i="<<i<<"is wrong");`

This is currently not possible with current implementation of CORE_ASSERT.

PR #168, opened 2017-03-17
Assert rework

### LINKED ISSUE #137: Add a "Release with asserts" configuration
For when I don't want to be in slow debug but still have an assert hit when bad things happen.
We could customize RelWithDebInfo so that it still triggers the asserts. We need to redesign a few macros in CoreMacros and investigate the ON_DEBUG() usage.

louen:
This branch includes fixes to windows build and a rework of asserts to fix #137 and #167

### DIFF
--- a/src/Core/CoreMacros.hpp
+++ b/src/Core/CoreMacros.hpp
@@ -17,6 +17,7 @@
 #include <cassert>
 #include <cstdio>
 #include <iostream>
+#include <sstream>
 
 // ----------------------------------------------------------------------------
 // Compiler identification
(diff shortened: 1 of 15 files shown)


src/Core/CoreMacros.hpp
```diff
@@ -246,58 +254,78 @@ typedef double Scalar;
...
-#define CORE_ERROR_IF( EXP, DESC )          \
-    MACRO_START                             \
-    if( UNLIKELY(!(EXP)) ) {                \
-        CORE_ERROR( DESC );                 \
-    }                                       \
+#define CORE_ERROR_IF( EXP, DESC ) \
+    MACRO_START                    \
+    if( UNLIKELY(!(EXP)) ) {       \
+        CORE_ERROR( DESC );        \
+    }                              \
```
louen: Actually this could directly call REPORT_FAIL with the correct expression. Fixing that soon

louen:
The `ON_ASSERT` macro is commonly use in grabbing the return code of some functions and asserting if the return code is wrong.

```c++
ON_ASSERT( int retcode = ) printf("lalala %u", 42);
CORE_ASSERT( retcode >= 0, "error in printf");
```
This use case is tricky to get right without the `ON_ASSERT` : you can't call the function inside the assert macro (because you change the behaviour of the code if asserts are disabled) and if you don't wrap the return code in `ON_ASSERT` you get  "unused variable" warnings when asserts are disabled.

The macro to check whether asserts are active or not is `CORE_ENABLE_ASSERT`  (this is actually a new thing from the change-set, before it used to be `CORE_DEBUG` as asserts were tied to the debug build config).
```c++
#ifdef CORE_ENABLE_ASSERT
/* code */
#endif 
```

nmellado:
Like I said, I got the point ;).
Ok for `CORE_ENABLE_ASSERT`. 
The PR sounds valid for me.

PR #171, opened 2017-03-24
Generalize the texture class to 2D and 3D

louen:
This commit simply make Tex3D a generic N-D texture with implementation for bilinear (2D) interpolation as well as the existing trilinear (it would be simple to extend it to 4D if necessary but it's not yet provided).

A template wizard could probably make a generic  template-recursive version instead of using explicit specializations.


nmellado:
Tested on linux plateform only ?

louen:
Windows too iirc. why ?

nmellado:
Just checking.

PR #174, opened 2017-03-28
Remove dynamic_cast and replace by fetch in system's map for componen…

hoshiryu:
…ts ;

This was here to have a more fluid GUI regarding selection of object parts.
For example, after posing a character, one needed to specifically select the SkinningComponent from the tree view to be able to change the Skinning method, then select the animation skeleton again to change the pose.

### DIFF
--- a/Plugins/Skinning/src/SkinningComponent.hpp
+++ b/Plugins/Skinning/src/SkinningComponent.hpp
@@ -15,14 +15,8 @@
 
 namespace SkinningPlugin
 {
-    class SkinningPluginC;
-}
-
-namespace SkinningPlugin
-{
     class SKIN_PLUGIN_API SkinningComponent : public Ra::Engine::Component
     {
-        friend class SkinningPlugin::SkinningPluginC;
     public:
 
         enum SkinningType
--- a/Plugins/Skinning/src/SkinningComponent.hpp
+++ b/Plugins/Skinning/src/SkinningComponent.hpp
@@ -13,11 +13,16 @@
 #include <Engine/Component/Component.hpp>
 #include <Engine/Managers/ComponentMessenger/ComponentMessenger.hpp>
 
+namespace SkinningPlugin
+{
+    class SkinningPluginC;
+}
 
 namespace SkinningPlugin
 {
     class SKIN_PLUGIN_API SkinningComponent : public Ra::Engine::Component
     {
+        friend class SkinningPlugin::SkinningPluginC;
     public:
 
         enum SkinningType
--- a/Plugins/Skinning/src/SkinningPlugin.cpp
+++ b/Plugins/Skinning/src/SkinningPlugin.cpp
@@ -73,11 +73,10 @@ namespace SkinningPlugin
         Ra::Engine::ItemEntry it = m_selectionManager->currentItem();
         if (it.m_entity)
         {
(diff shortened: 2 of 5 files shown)


hoshiryu:
In the case one needs a Component from another Plugin, a way to get it is to use the ComponentMessenger.
Also, in the case the Plugin creates multiple Components for an entity (of either the same class or different classes), it's up to the Plugin and its System to deal with them.

Plugins/Skinning/src/SkinningComponent.hpp
```diff
@@ -13,11 +13,16 @@
...
+namespace SkinningPlugin
+{
+    class SkinningPluginC;
+}
 
 namespace SkinningPlugin
 {
     class SKIN_PLUGIN_API SkinningComponent : public Ra::Engine::Component
     {
+        friend class SkinningPlugin::SkinningPluginC;
```
louen: Why is that friend declaration necessary ?

Plugins/Skinning/src/SkinningPlugin.cpp
```diff
@@ -73,18 +73,17 @@ namespace SkinningPlugin
         Ra::Engine::ItemEntry it = m_selectionManager->currentItem();
         if (it.m_entity)
         {
-            const std::vector< std::unique_ptr<Ra::Engine::Component> >& comps = it.m_entity->getComponents();
-            size_t i = 0;
-            for ( ; i<comps.size(); ++i)
+            bool found = false;
+            for (auto ec : m_system->m_components)
```
louen: suggestion : use`std::find_if()` for this kind of code ! :)

Plugins/Skinning/src/SkinningComponent.hpp
hoshiryu: because I need to access a protected member from SkinningPluginC

louen: Protected member of the  component ? Which one ? and why not make it public then ?

hoshiryu: Ok, this one was not supposed to be there but in the SkinningSystem

louen: Maybe it would make sense for the system class to give access to this data then (in the form of const ptrs) ?

Plugins/Skinning/src/SkinningSystem.hpp
```diff
@@ -17,9 +17,15 @@
...
 {
+    class SkinningPluginC;
+}
+
+namespace SkinningPlugin
+{
 
     class SKIN_PLUGIN_API SkinningSystem :  public Ra::Engine::System
     {
+        friend class SkinningPlugin::SkinningPluginC;
```
louen: I still think it would be better to have an accessor function. I don't like the idea of the "plugin" class (which is related to UI) being used in the System class which is related to the Engine.

nmellado:
@louen, are the changes introduced in 061434a ok ?

louen:
Yes :+1:

louen:
I'm going to squash the 3 commits into one.

nmellado:
ok.

nmellado:
Then we can squash #176

PR #176, opened 2017-04-02
Expe mathias

MathiasPaulin:
Here is (at last) the correct integration of glbinding with radium.
Lot of difficulties related to QT/OpenGL/glbinding (same as for glew for those who know these nightmares resulting from Qt being quite intrusive with OpenGL code).
But, all works fine on MacOs. Need to be compiled and tested on Linux (might be straightforward) and windows (I have no idea of what will be wrong but there are several parts I did'nt bother cause I've no mean to compile on windows ...

It will be a good thing if the merge with master could be done before the end of the week so that our interns could continue OpenGL Stuff refactoring.


cmake/FindRadium.cmake
```diff
@@ -66,19 +70,33 @@ IF ( RADIUM_ROOT_DIR )
...
         )
 
+#    FIND_LIBRARY ( GLBINDING_LIBRARIES
+#            NAMES glbinding glbinding32.lib glbinding.lib libglbinding.a
+#            PATHS ${RADIUM_ROOT_DIR}/${CMAKE_BUILD_TYPE}/3rdPartyLibraries/lib
+#            ${RADIUM_ROOT_DIR}/lib
+#            )
+
     SET ( Radium_FOUND TRUE )
+    find_package(GlBinding REQUIRED)
```
nmellado: With linux, this lines cause: 
```
CMake Error at cmake/FindRadium.cmake:80 (find_package):
  By not providing "FindGlBinding.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "GlBinding", but CMake did not find one.

  Could not find a package configuration file provided by "GlBinding" with
  any of the following names:

    GlBindingConfig.cmake
    glbinding-config.cmake

```

Assimp and GLBinding does not seems to be included the same way (no `find_package` for assimp). Is there a specific reason for that ?

nmellado:
Works for me on Linux.

nmellado:
Works for me on Debian in Release and Debug mode.

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
```diff
@@ -1,20 +1,24 @@
+#include <Engine/Renderer/OpenGL/OpenGL.hpp>
 #include <GuiBase/Viewer/Gizmo/GizmoManager.hpp>
 
 #include <Engine/Renderer/Camera/Camera.hpp>
 #include <Engine/Managers/SystemDisplay/SystemDisplay.hpp>
 
 
-#include <GuiBase/Viewer/Viewer.hpp>
+//#include <GuiBase/Viewer/Viewer.hpp>
+
```
louen: Can we remove these commented includes ?

src/GuiBase/Viewer/CameraInterface.hpp
```diff
@@ -67,6 +68,9 @@ namespace Ra
 
             virtual void update( Scalar dt ) {}
 
+
+            static const Engine::Camera& getCameraFromViewer(QObject *v);
+
```
louen: I'm not really happy about this but this is better than the previous solution.

src/Engine/Renderer/RenderObject/RenderObject.hpp
```diff
@@ -132,7 +134,7 @@ namespace Ra
             std::shared_ptr<Mesh> m_mesh;
 
             // FIXME(Charly): Remove this
-            RenderParameters m_renderParameters;
+            //RenderParameters m_renderParameters;
```
louen: I think we should follow charly here and remove this, shouldn't we ?

src/Engine/Renderer/Mesh/Mesh.hpp
```diff
@@ -69,19 +71,40 @@ namespace Ra
...
+                RM_QUADS                                                      = 0x0007, // decimal value: 7
+                RM_QUAD_STRIP                                                 = 0x0008, // decimal value: 8
+                RM_POLYGON                                                    = 0x0009, // decimal value: 9
+                RM_LINES_ADJACENCY                                            = 0x000A, // decimal value: 10
+                RM_LINE_STRIP_ADJACENCY                                       = 0x000B, // decimal value: 11
+                RM_TRIANGLES_ADJACENCY                                        = 0x000C, // decimal value: 12
+                RM_TRIANGLE_STRIP_ADJACENCY                                   = 0x000D, // decimal value: 13
+                RM_PATCHES                                                    = 0x000E, // decimal value: 14
+            };
+
```
louen: Is there a reason why we can't use the opengl constants here ?

src/Engine/Renderer/Mesh/Mesh.cpp
```diff
@@ -111,7 +111,7 @@ namespace Ra {
             GLenum type = GL_FLOAT;
 #endif
             constexpr GLuint size = VecArray::Vector::RowsAtCompileTime;
-            constexpr GLboolean normalized  = GL_FALSE;
+             GLboolean normalized  = GL_FALSE;
```
louen: why has the constexpr dissapeared here ?

src/Engine/RadiumEngine.cpp
```diff
@@ -9,19 +9,27 @@
...
+
+#include <Engine/Managers/EntityManager/EntityManager.hpp>
+#include <Engine/Managers/SignalManager/SignalManager.hpp>
+#include <Engine/Renderer/RenderObject/RenderObjectManager.hpp>
+
 #include <Engine/FrameInfo.hpp>
 #include <Engine/System/System.hpp>
 #include <Engine/Entity/Entity.hpp>
-#include <Engine/Component/Component.hpp>
+//#include <Engine/Component/Component.hpp>
```
louen: If this is commented out, why not remove it ?

cmake/submoduleGlBinding.cmake
```diff
@@ -0,0 +1,79 @@
+
+# here is defined the way we want to import glbinding
+# TODO : verify with VERBOSE=1 on build time that CXX_FLAGS are correct
```
louen: Can you add the TODO as issues so that we can track them ?

Plugins/Skinning/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 project(SkinningPlugin)
 
 cmake_minimum_required(VERSION 2.8.11)
+if (APPLE)
+    cmake_policy(SET CMP0042 NEW)
+endif(APPLE)
+
```
louen: what is that for ?

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
MathiasPaulin: Done

src/GuiBase/Viewer/CameraInterface.hpp
MathiasPaulin: I'm also not really happy about that but it does the work ... Need a more indepth refactoring to be able to access the current camera from different location in the software. But is this really a need or an ease ?

src/Engine/Renderer/RenderObject/RenderObject.hpp
MathiasPaulin: done

src/Engine/Renderer/Mesh/Mesh.cpp
MathiasPaulin: This is due to a lack of constexpr in glbinding. In glbinding GLboolean is declared as a class (and not an unsigned int as in gl.h) with two static objects GL_TRUE and GL_FALSE. As their is no constructor declared as constexpr in glbinding/gl/boolean.h, we canot declare normalized as a constexpr here. An what is the benefit of declaring a local (temporery) variable as a constexpr ?

src/Engine/RadiumEngine.cpp
MathiasPaulin: done

Plugins/Skinning/CMakeLists.txt
MathiasPaulin: need this to remove cmake warnings and allow the system to find the corect path for dynamically linked modules and libraries.
-- from cmake documentation
CMP0042
MACOSX_RPATH is enabled by default.
CMake 2.8.12 and newer has support for using @rpath in a target’s install name. This was enabled by setting the target property MACOSX_RPATH. The @rpath in an install name is a more flexible and powerful mechanism than @executable_path or @loader_path for locating shared libraries.
CMake 3.0 and later prefer this property to be ON by default. Projects wanting @rpath in a target’s install name may remove any setting of the INSTALL_NAME_DIR and CMAKE_INSTALL_NAME_DIR variables.
This policy was introduced in CMake version 3.0. CMake version 3.0.2 warns when the policy is not set and uses OLD behavior. Use the cmake_policy command to set it to OLD or NEW explicitly.

cmake/submoduleGlBinding.cmake
MathiasPaulin: OK

louen:
OK, so everyone agrees to squash this into one commit ?

nmellado:
Agree

PR #183, opened 2017-04-10
Fix FindRadium for out of repo plugin development

nmellado:


cmake/FindRadium.cmake
```diff
@@ -30,45 +30,47 @@ ENDIF(NOT RADIUM_ROOT_DIR)
...
+
+    IF (NOT GLBINDING_LIBRARIES)
+        FIND_LIBRARY ( GLBINDING_LIBRARIES
+            NAMES glbinding
+            PATHS ${RADIUM_ROOT_DIR}/${CMAKE_BUILD_TYPE}/3rdPartyLibraries/lib
+            )
+    ENDIF (NOT GLBINDING_LIBRARIES)
+
     # TODO (Mathias) : verify if cmake recommand this (append the include dir of module dependencis in the include dir of the module header)
+    # TODO (Nico) : this is unconsistent with the lib behavior: deps *.so are not added in RADIUM_LIBRARIES
```
nmellado: We need to chose what to do here

louen:
I think the right thing to do would be to split FindRadium in two
* the part that finds the root directory of the radium engine. This should not really change across updates of the engine and thus would not necessitate to be upgraded in plugins repositories.
* the part that defines the directories and include dirs, which could be included by plugins just like we do with CompileFlags

cmake/FindRadium.cmake
MathiasPaulin: Does transitive linking works on systems other than MacOs ?
Yep, the current find radium is inconsistant wrt lib and include behavior. 
This is because we need include directory of dependencies when compiling something that includes radium and its dependencies but transitive linking allows to not explicitely link with dependencies libraries when linking with radium libraries (already linked with dependencies).

Issue #173, opened 2017-03-27
Investigate possible use of Eigen::Ref


louen:
Typical use case : 

```c++
// library code
void foo(const Vector3& v) {  /* v = input argument */ }
void bar(Vector3& v) { /* v = in/out or output argument */}

// client code uses a Matrix3
Matrix3 M = /* ...*/
foo(M.col(1)); // <- compiles but creates a temporary Vector3
bar(M.col(2));// <- fails to compile : 
```
The `col()` function here does not return a `Vector3` but returns a temporary object implicitely convertible to a `Vector3`.
Because it's a temporary (i.e. a r-value) one cannot take a non-const reference to it (hence the failure to call `bar()`).
The case of `foo()` is a little more tricky but it seems that the compiler silently generates a temporary copy of the `Vector3`,  which is suboptimal. See it live here: http://ideone.com/gzPiKj

Fortunately the Eigen folk have a facility for this use case : `Eigen::Ref` which solves both problems 
if we
* replace pass-by-reference arguments (const references to some vector type `VectorD`) to `const Eigen::Ref<const VectorD>&`
* replace output by-reference arguments (non-const refs to VectorD) by `Eigen::Ref<VectorD>`

As explained by folks on the Eigen IRC channel the first one might be useless in practice, however the second use-case is more plausible (it happened to me at least...), so we might want to have a look at our use of output-by-reference arguments and put `Ref`s instead.

Here's the IRC log on #eigen  around this topic for reference:

```irc
14:10 < louen> say I have a function which takes a non-const ref to a vector
14:10 < louen> like
14:10 < louen> something like `void f( Vector3& v) { v[1] = 0;}`
14:11 < louen> is there a way I can use that function with an appropriate size 3x1 block of a matrix ?
14:11 < louen> e.g a col() of a Matrix3 ?
14:13 <@ChriSopht> louen: you can declare your function as void f(Eigen::Ref<Vector3> v) { v.y() = 0;}
14:14 <@ChriSopht> this will accept .col() of a matrix, if the matrix is column-major (and has the correct number of rows, and the correct scalar)
14:15 < louen> oh that's great
14:15 < louen> didn't know that feature
14:15 <@ChriSopht> alternative, with C++11, if you want to accept anything, is to use template<typename Derived> void f(Eigen::MatrixBase<Derived>&& v) {...}
14:18 < louen> yes but here it's a function for vector3s (it uses cross product) 
[...]

PR #184, opened 2017-04-11
Gizmos hide

louen:
This PR contains two commits 
* one that uses eigen::ref for some functions in Algebra
* one that replaces the current behaviour of gizmos with a more sensible one.

### DIFF
--- a/src/Core/Math/LinearAlgebra.hpp
+++ b/src/Core/Math/LinearAlgebra.hpp
@@ -156,7 +156,7 @@ namespace Ra
             inline bool checkRange( const Vector_& v, const Scalar& min, const Scalar& max );
 
             /// Get two vectors orthogonal to a given vector.
-            inline void getOrthogonalVectors( const Vector3& fx, Vector3& fy, Vector3& fz );
+            inline void getOrthogonalVectors( const Vector3& fx, Eigen::Ref<Vector3> fy, Eigen::Ref<Vector3> fz );
 
             /// Get the angle between two vectors. Works for types where the cross product is
             /// defined (i.e. 2D and 3D vectors).
@@ -164,7 +164,7 @@ namespace Ra
             inline Scalar angle( const Vector_& v1, const Vector_& v2 );
 
             /// @return the projection of point on the plane define by plane and planeNormal
-            inline Vector3 projectOnPlane(const Vector3& planePos, const Vector3 planeNormal, const Vector3& point);
+            inline Vector3 projectOnPlane(const Vector3& planePos, const Vector3& planeNormal, const Vector3& point);
 
             /// Get the cotangent of the angle between two vectors. Works for vector types where
             /// dot and cross product is defined (2D or 3D vectors).
--- a/src/Core/Math/LinearAlgebra.inl
+++ b/src/Core/Math/LinearAlgebra.inl
@@ -63,19 +63,19 @@ namespace Core
     }
 
     template<typename Vector_>
-    Scalar Vector::cotan(const Vector_& v1, const Vector_& v2)
+    Scalar Vector::cotan( const Vector_& v1, const Vector_& v2 )
     {
         return  v1.dot(v2) / v1.cross(v2).norm();
     }
 
     template<typename Vector_>
-    Scalar Vector::cos(const Vector_& v1, const Vector_& v2)
+    Scalar Vector::cos( const Vector_& v1, const Vector_& v2 )
     {
         return  (v1.dot(v2)) / std::sqrt( v1.normSquared() * v2.normSquared());
     }
 
     template <typename Vector_>
-    Scalar Vector::getNormAndNormalize( Vector_& v)
+    Scalar Vector::getNormAndNormalize( Vector_& v )
     {
         const Scalar norm = v.norm();
         v /= norm;
@@ -83,7 +83,7 @@ namespace Core
     }
 
     template <typename Vector_>
-    Scalar Vector::getNormAndNormalizeSafe( Vector_& v)
+    Scalar Vector::getNormAndNormalizeSafe( Vector_& v )
     {
         const Scalar norm = v.norm();
         if (norm > 0 ) {v /= norm;}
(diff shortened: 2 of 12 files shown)


src/Core/Math/LinearAlgebra.hpp
```diff
@@ -156,15 +156,15 @@ namespace Ra
...
+            inline void getOrthogonalVectors( const Vector3& fx, Eigen::Ref<Vector3> fy, Eigen::Ref<Vector3> fz );
 
             /// Get the angle between two vectors. Works for types where the cross product is
             /// defined (i.e. 2D and 3D vectors).
             template<typename Vector_>
             inline Scalar angle( const Vector_& v1, const Vector_& v2 );
 
             /// @return the projection of point on the plane define by plane and planeNormal
-            inline Vector3 projectOnPlane(const Vector3& planePos, const Vector3 planeNormal, const Vector3& point);
+            inline Vector3 projectOnPlane(const Vector3& planePos, const Vector3& planeNormal, const Vector3& point);
```
nmellado: Why not using const Ref<>, as in [Eigen documentation](https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html#title1):
```
inline Vector3 projectOnPlane(const Eigen::Ref<Vector3> planePos, const Eigen::Ref<Vector3> planeNormal, const Eigen::Ref<Vector3> point);
```

src/Core/Math/LinearAlgebra.inl
```diff
@@ -75,15 +75,15 @@ namespace Core
     }
 
     template <typename Vector_>
-    Scalar Vector::getNormAndNormalize(Vector_& v)
+    Scalar Vector::getNormAndNormalize( Vector_& v)
```
nmellado: If you really want to commit that change, you also want to add the space before the closing bracket.

src/Core/Math/LinearAlgebra.inl
```diff
@@ -75,15 +75,15 @@ namespace Core
...
     {
         const Scalar norm = v.norm();
         v /= norm;
         return norm;
     }
 
     template <typename Vector_>
-    Scalar Vector::getNormAndNormalizeSafe(Vector_& v)
+    Scalar Vector::getNormAndNormalizeSafe( Vector_& v)
```
nmellado: Same as line 78

src/Engine/Renderer/RenderObject/Primitives/DrawPrimitives.cpp
```diff
@@ -464,7 +464,7 @@ namespace Ra {
                 std::vector<uint> indices;
                 indices.reserve(pointCount * 2 - 2);
 
-                Scalar dt = 1.0 / Scalar(pointCount - 1);
+                Scalar dt = 1.0f / Scalar(pointCount - 1);
```
nmellado: I believe the correct way to do it is `Scalar(1) / / Scalar(pointCount - 1)`.
With your version, you implicitly assume that Scalar is float.

src/GuiBase/Viewer/Gizmo/Gizmo.hpp
```diff
@@ -41,9 +44,17 @@ namespace Ra
                 }
             }
 
+            // shows or hides the gizmos drawables.
+            void show(bool on)
+            {
+                for (auto ro: m_renderObjects)
+                {
+                    Ra::Engine::RadiumEngine::getInstance()->getRenderObjectManager()->getRenderObject(ro)->setVisible(on);
```
nmellado: I suggest to temporary store the pointer to the RenderObjectManager out of the loop (code readability)

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
```diff
@@ -86,43 +90,56 @@ namespace Ra
...
+            {
+                currentGizmo()->selectConstraint( drawableId );
+            }
+        }
+
+        Gizmo* GizmoManager::currentGizmo()
+        {
+            return ( m_currentGizmoType == NONE )?
+                   nullptr :
+                   m_gizmos[ m_currentGizmoType - 1].get();
```
nmellado: Need to say in the documentation that invalid gizmos are set to nullptr.

src/GuiBase/Viewer/Gizmo/GizmoManager.hpp
```diff
@@ -57,13 +62,17 @@ namespace Ra
             /// Retrieve the transform from the editable and update the gizmos.
             void updateValues() override;
 
+
```
nmellado: Are these empty line wanted (with line 22) ?

src/Core/Math/LinearAlgebra.hpp
louen: Oh, but have you read what Eigen people said in the issue about eigen const ref (#173) ?
> 18:45 <@ChriSopht> louen: especially for Vector3d it may be overkill
So I'm following suit here :)

nmellado: Then he says:

> 18:48 <@ChriSopht> use Ref<const ...> when it is semantically necessary 

Again it depends if you plan to take benefit from `Eigen::Ref` or not. Do you believe the non-const function will be called will `mymatrix.col()` while the const function will always be called with an actual `Vector3d` ?

Another option would be to provide both prototypes (i.e. `const Vector3d&` and `const Eigen::Ref<Vector3>`), but I'm not sure the compiler would manage do disambiguate between both cases.

louen: For non-const refs, we must use `Eigen::Ref` if we ever want to use something like a matrix column as an argument, otherwise the compiler refuses it.
In the case of const ref the compiler accept it but it will generate a copy (or it could optimize it).
The question is, do we want to live with this copy in this specific use case (which is very rare) or modify all our code accepting `const Vector3&` ?

nmellado: > For non-const refs, we must use Eigen::Ref if we ever want to use something like a matrix column as an argument, otherwise the compiler refuses it.

I do agree, Eigen::Ref is the proper solution for that.

> In the case of const ref the compiler accept it but it will generate a copy (or it could optimize it).
> The question is, do we want to live with this copy in this specific use case (which is very rare) 

As stated by Eigen guys:
> 18:47 <@ChriSopht> even copying a 100d vector does not cost that much. as always avoid premature optimization

having a copy (especially on such rare case, as  you said) is not critical, and might be optimized by the compiler (as you said). So, I would recommend to use const Ref, to have consistent API between const and non-const functions, and simpler coding rules. In general, the overhead should be almost 0.

> or modify all our code accepting const Vector3&

I'm not sure these practical refactoring consideration should influence or code design.

louen:
I can haz PR now ? :cat:

nmellado:
Looks amazing.

PR #193, opened 2017-05-18
Fixbuildchain

MathiasPaulin:
Small fixes in cmake files to built radium from in distribution source to out distribution source


nmellado:
This PR is deprecated and has been replaced by #195.

Issue #194, opened 2017-05-18
acecpted pull request Show/hide gizmos instead of delete/creating them break simple app test0


MathiasPaulin:
The compilation is OK, but, the change in the constructor of the gizmo manager (file  src/GuiBase/Viewer/Gizmo/GizmoManager.cpp, adding the creation of several gizmos (Translate, Rotate, nullptr) make the application test0 segfault.

When creating such gizmos, the call Engine::SystemEntity::uiCmp() try to get the instance of the singleton SystemEntity that is not yet created (and then equals to nullptr) when this is called by the simple app test0. So ... the results is ... BAD_ACCESS --> SEGFAULT.
Authors of pull request MUST test the 2 example applications and all the "standard" plugins in each build type : Debug, RelWithDebInfo, Release before submitting pull request. While waiting for a working continuous integration procedure (OK, not tomorrow), reviewers of pull request must also verify this on their system!!!

I'll try to fix this in the frame my fixBuildSystem pull request (even if thes are two different things) , so, wait a few before validating it.

louen:
This is why we need continuous integration :)

MathiasPaulin:
Yep, this is mentioned that in my message ... But thereis several technical problems for that. it's on the road but ... not tomorrow as I said.

nmellado:
Test0 does not crash anymore on VS.
Is this fixed ?

louen:
Can we close this issue ?

hoshiryu:
Ping: is this issue still topical?

nmellado:
@MathiasPaulin , what do you think ?

MathiasPaulin:
we can close this issue.

PR #195, opened 2017-05-22
Fix mvsc + buildchain

nmellado:
This branch includes changes from previous PR #193 + changes for Visual Studio 2017 compliance.

I'm starting this new PR to have a common thread with all changes at the same place.
Contributions are welcome.

In its current state:
 - Linux platform: 
    - compiles in Release, RelWithDbg, Debug
    - main app and test0 runs in Release, Debug
 - Windows VS:
    - compiles in Release and Debug
    - apps crashes for all configurations. 
 - Windows mingwin (QtCreator)
    - compiles in Release and Debug
    - apps run for all configurations 
       - iff glbinding and assimp libs are copied (manually) to the bin dir. FIXME
 - MacOS
    - compiles in Release, RelWithDbg, Debug
    - main app runs for all configurations iff the cmake process is run twice. FIXME

Changes:
 - output build files into Bundle-{COMPILER_ID} folder
 - always compiles and use 3rdparty in release mode


MathiasPaulin:
this PR compiles fine on MacOs in all build mode (Debug, Release, RelWithDebInfo).
there is still one building problem (i did'nt find where yet) when running an application for the first time. Even if all dependencies are correctly compiled and "installed", at the first run of an application (main-app or test0), the application fail to dynamically load libglbinding.dylib. By running cmake && make again (that surprisingly just do the linking of all lilbs and executable), everything works fine after that.
main-app run fine in each build mode.
Test0 run fine but is buggy and crash at exit. This is due to the problem described in issue #194

nmellado:
Thanks @MathiasPaulin for the tests.
I've done some testing on test0, it seems to me that the crash is due to GL, I got a crash when trying to access `viewer->context()` (see dbd53e9).

My guess is a GlBinding/QtGL fight. What do you think ?

MathiasPaulin:
@nmellado Concerning your problem when accessing viewer->context() at line 36 of Applications/Test0/main.cpp, the fact that this pointer is nullptr is safe and everything is OK. At this point, the OpenGL context was not yet initialized (see doc at http://doc.qt.io/qt-5/qopenglwidget.html#context). You can access the context only once the InitializedGL was called after the window mapping. This is common to all version of Qt and is a reason why you cannot call OpenGL stuff in the constructor of a Qglwidget or QOpenGlWidget. I moved your CORE_ASSERT just after _viewer->show at line 114 of Applications/Test0/main.cpp. The only concerns that remains (on MacOs at least) is the problem of unloading shaders and the hacky^H^H^H^H^HdefaultConfig that is still incomplete ... and make the application crash at exit.

nmellado:
I've copied this CORE_ASSERT line from the Main app. Anyway, even without it, test0 is crashing when qt is trying to display the main window.

Applications/Test0/minimalapp.hpp
```diff
@@ -0,0 +1,50 @@
+#include <QApplication>
```
louen: Maybe we should update this class to follow the rest of the codebase and use `m_` as a prefix for members. (Remember that the standard frowns upon using underscores as the first character of an identifier...)

CMakeLists.txt
```diff
@@ -91,19 +92,14 @@ include(SubModules)
...
 # ----------------------------------------------------------------------------------------------------------------------
 #   Applications
 #   Set Base directories for dependencies
-# use Eigen version included in the Radium distribution, not the system-wide installed if available
-
-#set (EIGEN3_DIR ${RADIUM_SUBMODULES_INSTALL_DIRECTORY})
-#set (GLBINDING_DIR ${RADIUM_SUBMODULES_INSTALL_DIRECTORY})
 
+set( OMIT_MAIN_APP False CACHE BOOL "Choose to build or not the main radium application" )
 add_subdirectory(Applications)
```
louen: I thought this cmake option disappeared. Why did it pop back ?

Plugins/Animation/CMakeLists.txt
```diff
@@ -67,6 +67,7 @@ set(Qt5_LIBRARIES ${Qt5Core_LIBRARIES} ${Qt5Widgets_LIBRARIES} ${Qt5OpenGL_LIBRA
 target_link_libraries( Animation
                        ${Qt5_LIBRARIES}
                        ${RADIUM_LIBRARIES}
+					   ${GLBINDING_LIBRARIES}           # Radium dep
```
louen: Y U no use spaces for indentation ?

Plugins/Dummy/CMakeLists.txt
```diff
@@ -1,49 +1,32 @@
 project(DummyPlugin)
 
 cmake_minimum_required(VERSION 2.8.11)
+if (APPLE)
+    cmake_policy(SET CMP0042 NEW)
+endif(APPLE)
```
louen: What is this cmake_policy for ? It would be nice to have a little  comment explaining why.

Plugins/Dummy/CMakeLists.txt
```diff
@@ -77,6 +60,7 @@ set(Qt5_LIBRARIES ${Qt5Core_LIBRARIES} ${Qt5Widgets_LIBRARIES} ${Qt5OpenGL_LIBRA
 target_link_libraries( Dummy
                        ${Qt5_LIBRARIES}
                        ${RADIUM_LIBRARIES}
+				       ${GLBINDING_LIBRARIES}           # Radium dep
```
louen: Space ! Spaaaaaace ! Spaaaaace !

Plugins/FancyMesh/CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ set(Qt5_LIBRARIES ${Qt5Core_LIBRARIES} ${Qt5Widgets_LIBRARIES} ${Qt5OpenGL_LIBRA
 target_link_libraries( FancyMesh
                        ${Qt5_LIBRARIES}
                        ${RADIUM_LIBRARIES}
+				       ${GLBINDING_LIBRARIES}           # Radium dep
```
louen: See above.

Plugins/Skinning/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ set(Qt5_LIBRARIES ${Qt5Core_LIBRARIES} ${Qt5Widgets_LIBRARIES} ${Qt5OpenGL_LIBRA
 target_link_libraries( Skinning
                        ${Qt5_LIBRARIES}
                        ${RADIUM_LIBRARIES}
+				       ${GLBINDING_LIBRARIES}           # Radium dep
```
louen: Id.

README.md
```diff
@@ -48,60 +66,62 @@ $ cmake ..
...
 
-The compiled application can be found in `bin`. Default plugins DLL are compiled in
-`Plugins/bin` by default.
+#### Supported versions of MSVC
+Since Radium requires:
+* the C++11/C++14 advanced features such as `constexpr`, 
+* cmake built-in support
 
-### Building on Windows with Visual Studio
+you will need a recent MSVC (2015 minimum).
```
louen: So this needs to be updated to 2017 as well

README.md
```diff
@@ -48,60 +66,62 @@ $ cmake ..
...
-
-#### Building
+Other dependencies (Eigen, Assimp and glbinding) are included as a submodule in the git repository.
 
-As long as cmake run smoothly the engine should compile.
+#### Getting started with Visual Studio
 
-### Run
+Thanks to the integrated support of CMake in Visual Studio, you don't need a VS solution to build your project: open the Radium folder (via *File* > *Open* > *Folder ...* or `devenv.exe <foldername>`).
+VS should run cmake, generate the target builds (Debug and Release by default).
```
louen: No RelWithDeb on visual studio ?

README.md
```diff
@@ -48,60 +66,62 @@ $ cmake ..
...
-You can now run Cmake in Qt Creator. It should locate Qt libs automatically. 
-It should also find GLEW provided you followed the previous step. 
-If not, you can always set `GLEW_INCLUDE_DIR` and `GLEW_LIBRARY` manually to their correct locations.
+Right click on CMakeList.txt > build > all.
 
-### Running
+#### Run
 
-If you run the program within QtCreator the path should be correctly set, and the executable should run fine.
+Assimp and GlBinding libs need to be manually copied in the bin directory (FIXME).
```
louen: I would suggest "DLLs" instead of "libs" (in windows world :  `.lib` = static library)

src/Engine/Renderer/RenderTechnique/RenderTechnique.hpp
```diff
@@ -30,7 +30,7 @@ namespace Ra
             RA_ENGINE_API void updateGL();
             bool shaderIsDirty;
 
-            ShaderConfiguration getBasicConfiguration() const;
+			RA_ENGINE_API ShaderConfiguration getBasicConfiguration() const;
```
louen: Space ? Space !

src/Engine/Renderer/RenderTechnique/ShaderConfiguration.cpp
```diff
@@ -1,10 +1,154 @@
 #include <Engine/Renderer/RenderTechnique/ShaderConfiguration.hpp>
 
+/**
+ * @todo : make default shader configuration/default shader programm usable outside of the main radium distribution or
+ * find a way to ensure that "plain.glsl.*" is always reachable (and change the name of this.
+ * @see TranslateGizmo and RotateGizmo that explicitely try to load the shaer "Plain" that is main-app dependant!
+ */
```
louen: You can reference issue #194  in this comment I think.

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
```diff
@@ -12,6 +12,8 @@ namespace Ra
 {
     namespace Gui
     {
+        /* TODO : Mathias -- Beuurk. Creating gizmos by default implies that all applications developped on top of the engine will have them.
+         * This is not a good idea. Applications mus be able to define and create their their own gizmos */
```
louen: Same here, reference GH issue number

Plugins/Dummy/CMakeLists.txt
MathiasPaulin: CMake 2.8.12 and newer has support for using @rpath in a target’s install name. This was enabled by setting the target property MACOSX_RPATH. The @rpath in an install name is a more flexible and powerful mechanism than @executable_path or @loader_path for locating shared libraries.

CMake 3.0 and later prefer this property to be ON by default. Projects wanting @rpath in a target’s install name may remove any setting of the INSTALL_NAME_DIR and CMAKE_INSTALL_NAME_DIR variables.

This policy was introduced in CMake version 3.0. CMake version 3.0.2 warns when the policy is not set and uses OLD behavior. Use the cmake_policy command to set it to OLD or NEW explicitly.

Plugins/Animation/CMakeLists.txt
MathiasPaulin: Because I prefer tabs :-p
Sirously, this comes when switching between different configurations and indentation recommandation on several tools on several platforms ... I'll try to fix this for my next PR

src/Engine/Renderer/RenderTechnique/RenderTechnique.hpp
MathiasPaulin: Bananas .... Bananas ...

Issue #198, opened 2017-06-20
ShaderConfiguration - DefaultShader does not exist


Jeckhys:
A default shader configuration has been introduced since this commit : 

https://github.com/AGGA-IRIT/Radium-Engine/commit/59b54d242b56f21f9711b120219aa5d95da67b7a#diff-2a234226723ed7563d8e4a151b87864e

The shader source files do not exist (no "DefaultShader.*.glsl").

We should either create DefaultShader.*.glsl or replace "DefaultShader.*.glsl", in ShaderConfiguration.cpp at lines 146 & 147, with "Default.*.glsl"

By the way, this should be "Shaders/DefaultShader.*.glsl".

nmellado:
As you can see here, this shader is directly included in the source code: https://github.com/AGGA-IRIT/Radium-Engine/commit/59b54d242b56f21f9711b120219aa5d95da67b7a#diff-2a234226723ed7563d8e4a151b87864e

hoshiryu:
Ping: is this issue still topical?

nmellado:
Fixed by #199

PR #199, opened 2017-06-21
globjects integration

Jeckhys:
Here is globjects integration, it includes : 

- `Framebuffer`
- `Texture`
- `Shader`
- `Program`

classes from globjects.

Old `FBO` & `ShaderObject` classes have been deleted.

There are also some memory leaks which seem to come from globjects methods `setUniform()` and `setName()` (called in `ShaderProgram`). It would be nice to have a review about it since I've not been able to find the real source of these leaks for now (I'm still looking for it...)


nmellado:
Does this fixes #198  ?

PR #200, opened 2017-06-28
OpenMesh integration

Syn4ck:
This PR includes the addition of the OpenMesh submodule and prepares the deletion of the DCEL structure in favor of OpenMesh structures.

- Notice the DCEL class is still functional and is just marked as deprecated. The objective is to propagate OpenMesh structures by encouraging replacement of DCELs in all Radium branches. Currently, OpenMesh structures are not used at all.

- For an efficient use and integration of OpenMesh structures I refer you to the OpenMesh documentation [https://www.openmesh.org/media/Documentations/OpenMesh-Doc-Latest/index.html](url)

- Please check the interagtion of OpenMesh submodule and signal any problems with classes naming and coding rules conformity.

- On the sidelines, a test has been added to check the correspondence between a DCEL and an OpenMesh structure.


src/Core/Mesh/DCEL/Face.hpp
```diff
@@ -20,7 +20,8 @@ namespace Core {
 * Chapter 2, Paragraph 2.2, Page 29
 *
 */
-class RA_CORE_API Face : public IndexedObject  {
+
+class [[deprecated]] RA_CORE_API Face : public IndexedObject  {
```
louen: Has anyone checked that VS supports the [[deprecated]] flag ? If not it might require a macro in CoreMacros.

Syn4ck:
This PR is now update and works with last master version. 
The [[deprecated]] flag seems supported by Visual since VS 2015 (https://docs.microsoft.com/fr-fr/cpp/cpp/attributes2)

Syn4ck:
This PR is now update and works with last master version. 
The [[deprecated]] flag seems supported by Visual since VS 2015 (https://docs.microsoft.com/fr-fr/cpp/cpp/attributes2). However, I'm not sure to use it well...

louen:
You know, why not delete the DCEL we have directly ? Dead code tell no tales. Let's keep just one version that works.

Syn4ck:
I didn't delete it because it's currently used by src\Core\Animation\Skinning\RotationCenterSkinning.cpp and src\Core\Algorithm\Subdivision\FullEdgeOperation.cpp (and maybe in some plugins). These files should certainly be updated, but I leave this job to people working on them.

cmake/submoduleOpenMesh.cmake
```diff
@@ -0,0 +1,55 @@
+set (CMAKE_CXX_FLAGS "-fexceptions")
```
hoshiryu: Here is a big issue: this line completely overrides the CMAKE_CXX_FLAGS for all Radium.

The thing here is that OpenMesh requires exceptions to be enabled when it is compiled.
However, since some exceptions management is done in hpp files (see OpenMesh/Core/IO/StoreRestore.hh), when compiling OpenMesh's integration within Radium, a "exception handling disabled" error is thrown.

Hence, we must decide on whether enabling exceptions OR not using OpenMesh (or finding some kind of a hack).

dlyr:
This PR is included in #257

PR #204, opened 2017-07-12
Fix wrong search dir for libradiumCore

dlyr:


louen:
Actually FindRadium is also wrong on Windows for the libraries location. I will push my changes to this branch.

cmake/FindRadium.cmake
```diff
@@ -59,45 +64,45 @@ IF ( RADIUM_ROOT_DIR )
     ############################################################################
     # Get dependencies if not already specified
     IF(NOT EIGEN3_INCLUDE_DIR)
-        set(EIGEN3_INCLUDE_DIR ${RADIUM_BUNDLE_DIRECTORY}/3rdPartyLibraries/include)
+        set(EIGEN3_INCLUDE_DIR "${RADIUM_3RDPARTY_DIR}/include")
```
louen: @nmellado  I would like you to confirm this is the right solution.

nmellado:
I've added to the PR a new behavior for the file `Version.cpp`.
It is now generated once to avoid systematic call to the linker each time cmake is called.
Since the file is produced by cmake, it is generated in the build dir, not in the source dir. 
Consequently, removing the build dir will lead to the re-generation of this file.

Note that there is a dependency between this file and libRadiumCore, which allows to recompile the core library each time the file is modified.

Minor thing, the current git changeset is now added in `Version.cpp` and printed when starting Radium main app.

hoshiryu:
Regarding variables names, I think that `RADIUM_INCLUDE_DIR` should be renamed into `RADIUM_INCLUDE_DIRS` since there are multiple paths and they don't just apply to Radium but also to the 3rdPartyLibs.
Also, `RADIUM_ROOT_DIR`, `RADIUM_INCLUDE_DIR` and `RADIUM_PLUGIN_OUTPUT_PATH` are cited in the top-of-file comment, while other set variables are not (`RADIUM_BUNDLE_DIRECTORY`, `EIGEN3_INCLUDE_DIR`, ...). I think we should either unset them or "declare" them in the top-of-file comment.

MathiasPaulin:
I agree with @hoshiryu. When several directories are set in one variable, cmake guidelines recommend to name the variable XXX_INCLUDE_DIRS or XXX_LIBRARIES ...
I've modified this.
I also removed out some debug messages in the CMakeFiles and removed unnecesarry duplicate
For the last version of @nmellado, it is important that the file src/Core/Utils/Version.cpp is removed before compiling. When I checkout this PR, this file remained in my source directories and linker generates duplicate symbol errors.
I've commit some minor fixes and approved this PR.

nmellado:
> For the last version of @nmellado, it is important that the file src/Core/Utils/Version.cpp is removed before compiling. When I checkout this PR, this file remained in my source directories and linker generates duplicate symbol errors.

Yes indeed.
Merging then.

PR #205, opened 2017-07-12
Convert HandleData into Skeleton in Asset function

louen:
This PR moves the skeleton creation to a function in the Asset namespace instead of the Animation plugin.


hoshiryu:
Actually, are changing the IO management via the PR #209.
It would be nice of you to review the PR #209 in order to validate the architecture, and then update your present PR accordingly once the PR #209 is accepted.

PR #209, opened 2017-07-18
Radium IO

Jeckhys:
This PR includes the new Radium IO system, and also add the AssimpFileLoader (default ON) and PbrtFileLoader (default OFF) which is not yet done.


louen:
Can I have a bit of a high-level explanation of the changes ? 
the changelist is big and I would like to review it but without any directions this is a bit daunting...

hoshiryu:
As we discussed with @MathiasPaulin, @dlyr and @nmellado, the thing here is to move all the file loading stuff in an "IO" module, and having an interface for file loaders to implement.
Hence, only the module IO would depend on file loader stuff such as assimp, pbrt or others.

louen:
Does it change any of the current file loading interfaces for client code (e.g. `FileData`, `GeometryData` and `HandleData`) ?

hoshiryu:
No, these are just moved into IO

Applications/MainApplication/Gui/MainWindow.cpp
```diff
@@ -4,7 +4,7 @@
 #include <QFileDialog>
 #include <QToolButton>
 
-#include <Core/Utils/File/OBJFileManager.hpp>
+#include <Core/File/deprecated/OBJFileManager.hpp>
```
nmellado: I would personally prefer to keep the obj loader as a simple loader, very specialized but which can be useful I you don't want to use assimp.

hoshiryu: The thing is it has to be adapted to the FileLoader interface, that's why it is in Core/File/deprecated instead of IO.

CMakeLists.txt
```diff
@@ -118,6 +120,7 @@ add_subdirectory(${RADIUM_SRC_DIR})
 # ----------------------------------------------------------------------------------------------------------------------
 #   Applications
 #   Set Base directories for dependencies
+set( OMIT_MAIN_APP False CACHE BOOL "Choose to build or not the main radium application" )
```
louen: This was  deleted a few changelists ago. Why put it back ?

Applications/MainApplication/Gui/MainWindow.cpp
louen: The best way to do it is to file an issue so that we don't forget to put it back on once this pr is merged

src/Core/File/AnimationData.inl
```diff
@@ -1,22 +1,36 @@
...
 }
 
+inline AnimationTime& AnimationData::getTime() {
+    return m_time;
+}
+
+inline void AnimationData::setTime( const AnimationTime& time ) {
+    m_time = time;
+}
+
```
louen: Why is there the need for a setter *and* a getter of non-const ref ?

src/Engine/Renderer/RenderTechnique/RenderTechnique.hpp
```diff
@@ -19,6 +19,7 @@ namespace Ra
     namespace Engine
     {
         // NOTE(Charly): Would it be interesting to provide more stuff here ?
+        // TODO(Val) : we need proper memory management of these objects.
```
louen: This has to go away !

src/IO/AssimpLoader/AssimpAnimationDataLoader.cpp
```diff
@@ -1,29 +1,28 @@
...
+#include <IO/AssimpLoader/AssimpWrapper.hpp>
 
 namespace Ra {
-namespace Asset {
-
+namespace IO {
 
 /// CONSTRUCTOR
-AssimpAnimationDataLoader::AssimpAnimationDataLoader( const bool VERBOSE_MODE ) : DataLoader< AnimationData >( VERBOSE_MODE ) { }
+AssimpAnimationDataLoader::AssimpAnimationDataLoader( const bool VERBOSE_MODE ) : DataLoader< Asset::AnimationData >( VERBOSE_MODE ) { }
```
louen: So did the asset namespace disappear or not ? shouldn't it be `IO::` ?

CMakeLists.txt
dlyr: Seems like a divergence between two branches ... 
but I think it is good to keep it

nmellado: What was the rational behind this deletion ?

Applications/MainApplication/Gui/MainWindow.cpp
nmellado: Done, see #211 .

src/Core/File/AnimationData.inl
dlyr: do you mean we can remove  `inline AnimationTime& AnimationData::getTime()`  ? (or have` inline const AnimationTime& AnimationData::getTime()` )

nmellado: +1

CMakeLists.txt
dlyr: something like cosmic ray

nmellado: ok, so we keep it.
I believe that it should be extended to something allow to build only the lib, or the libs + apps, libs+examples, etc..

I suggest to change this variable to `BUILD_APPS̀`, and update behavior accordingly. 
Good ?

louen: Fine by me

src/Core/File/AnimationData.inl
louen: One or the other, but not both.

dlyr: vote for `inline const AnimationTime& AnimationData::getTime()`

nmellado:
Now works on windows with visual.

MathiasPaulin:
Now works on MacOs

src/Engine/Renderer/RenderTechnique/RenderTechnique.hpp
dlyr: What has to go away ? the comment or whole class ?

src/IO/AssimpLoader/AssimpAnimationDataLoader.cpp
dlyr: Seems like IO is for file IO stuff, and Asset is for storing data from IO stuff ...

src/Engine/Renderer/RenderTechnique/RenderTechnique.hpp
MathiasPaulin: I think that the raw pointer const ShaderProgram* shader must become shared_ptr<ShaderProgram> to have proper memory management. The class might remains but bettter memory management must be fixed.

src/IO/AssimpLoader/AssimpAnimationDataLoader.cpp
MathiasPaulin: I agree with @dlyr

PR #210, opened 2017-07-18
Fix build Radium with double precision

hoshiryu:
When compiling Radium with double precision floating point, glm wrappers cause an ambiguity.
Since the RadiumToGlm wrappers are used only to set uniforms for shader programs, I just enforced them to take single precision vectors/matrices as input.

Shall we have both single and double precision versions of these wrappers?
What about the GlmToRadium wrappers (that are currently not used)?

### DIFF
--- a/src/Core/Math/GlmAdapters.hpp
+++ b/src/Core/Math/GlmAdapters.hpp
@@ -124,9 +124,19 @@ namespace Ra
         // It may be useful because globjects's getUniform() method returns a glm structure and Radium
         // only deals with Eigen's structures.
 
+        inline void fromGlm( const glm::vec2& v, Vector2& out )
+        {
+            out << v[0], v[1];
+        }
+
         inline Vector2 fromGlm( const glm::vec2& v )
         {
             return Vector2( v[0], v[1] );
+        }
+
+        inline void fromGlm( const glm::vec3& v, Vector3& out )
+        {
+            out << v[0], v[1], v[2];
         }
 
         inline Vector3 fromGlm( const glm::vec3& v )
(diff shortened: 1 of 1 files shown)


MathiasPaulin:
I think this is not the way to solve this problem. In Radium, all is Dependent upon the "Scalar" typedef, defined in CoreMacros.hpp according to the compilation option CORE_USE_DOUBLE.
Nevertheless, when converting a Vector from Ra::Core to a glm::vec* for globjects, I think that we can always transform to floats or, in the underlying type of glm::vec2 (in shaders, uniforms are declared as float vectors whit the underlying type configurable using some #option in GLSL).

So, I recommand rather to write the fonction as, eg, 
inline glm::vec2 toGlm( const Vector2& v )
          {
              return glm::vec2( static_cast<glm::vec2::value_type>(v( 0 )), static_cast<glm::vec2::value_type>(v( 1 )) );
          }
and so on for others.
Note that glm::vec2::value_type can be defined by the user at compile time (so, we can always set it at the same value than Scalar ...

nmellado:
I do agree with mathias about the compile-time cast.
FYI, Eigen also provide a way to cast a matrix directly `Eigen::Matrix::cast<float>()`.

For the same used for the cast type, I would prefer something like `GlmAdapters::GlmScalar`.

Since we are on this class, what about adding in-place copy for glm->eigen adapters, such as:
```
inline fromGlm( const glm::vec2& v, Vector2& out )
{
   out << v[0], v[1];
}

inline Vector2 fromGlm( const glm::vec2& v )
{
   Vector2 out;
   fromGlm(v,out);
   return out;
}
```

PR #213, opened 2017-07-20
Rebase val_fixes. I hope I haven't mess up anything

dlyr:
Also added method to get the bone index of a ro

Add slerp function

Fix for Qt 5.7 and lower

Added continuous projection on polyline

Add a max threads option to main app

Corrected a typo in task queue and changed the assert to a warning for
unmet dependencies

Add animation time output from animation  component

Fix task queue assert, added better message

UI to print timings and task graph

Added a macro to tell Release and RelDeb builds

Make fitCamera a method of mainwindow(+button)

Added end of frame signal and anim time display

Add function to extract the edge list of a mesh

Fix animation keyframes access

Remove the BVH (for frustum culling) in ROManager

not used BUT was causing a memory leak due to bad management

Added lerp function

Add stepping functionality to gizmos

VSA fix from Maurizio ;

Add color scattering function from Maurizio ;

Added specific folder for  exported frames

Added a frame counter on skinning data

Fix swing test precision error in release

Fix windows build (enable RTTI)

Note: this is a temporary fix. It would be better to avoid using dynamic
cast in globjets if possible.

Fix GUI according to tab widget and renderer menu ;

update submodule, remove old comment


dlyr:
Ok, I thought this was introduced by val_fixes ;)

louen:
It's a pity we lost the commit history though...

dlyr:
How I think it should be better ;) I can do it again with the commit history !

PR #222, opened 2017-08-21
Add new example demonstrating use with GLFW and nanoGUI

nmellado:


MathiasPaulin:
I'm trying to compile this PR on MacOs ... some difficulties due to nanogui (which compile only with Apple-clang (C++11) on MacOs, not gnu-gcc due to some objective C++ features used by nanogui). As soon as I success in compiling all with CLANG, I'll try to compile nanogui with clang and Radium with gnu-gcc. For the moment, I have a conflict in the std::string ABI between C++11 libc++ from clang (used by nanogui) and C++14 libc++ from gcc (used by radium). 
As soon as everything is fixed, this PR could be approved.

nmellado:
An easiest way would also to have this example working only on linux :)
The issues are not directly linked to Radium's code, it just demonstrate it works.

Except of course if you plan to use these libraries for your projects

MathiasPaulin:
I agree that we could be not so tightly coupled with Qt. Specifically on MacOs wher there is still a retina problem with QT and the main app (on Retina display, there is a (known) issue in the OpenGL viewport size that fill only 1/4 of the window. by using Qt in an other way that in main-app or by being able to use Radium without Qt on MacOs should be a good idea.
More, I like nanogui for its simplicity, efficiency and  ...
So, I want to have the ability to use these libraries on MacOs ...

nmellado:
Fair enough..

MathiasPaulin:
Well, As it is functional also on MacOs, this PR could be merged. 
The only question I have is the use of nanogui AND glfw. At execution time, I got this :
`objc[46779]: Class GLFWLayoutListener is implemented in both /opt/local/lib/libglfw.3.dylib (0x10465bae0) and /opt/local/lib/libnanogui.dylib (0x1045e3fa8). One of the two will be used. Which one is undefined.`
and a 5 more similar messages.

nmellado:
Hum, according to the documentation, 

> NanoGUI builds on GLFW for cross-platform OpenGL context creation and event handling ...

does that mean that GLFW is included in nanogui binaries ?
I can try to play around on linux if you want.

MathiasPaulin:
Yes, nanogui is built using several submodules (and links to those) among which we found eigen, glfw, nanovg, ...
I did not inspect why they link statically with those but if this is the case, no need to link the demo application with glfw library ...

MathiasPaulin:
Well .. Reading some issues on nanogui, I found this answer from Wenzel Jackob.

> And realistically, the encouraged answer here is to follow the approach described in the documentation using NanoGUI as a submodule.

I'll update the PR in order to use nanogui as if it was a submodule (but without adding really the submodule).
I.e, I'll create a submodulenanogui.cmake file on the model of others submodule but will not add nanogui as a submodule as it is necessary just for the small GLFW/nanogui example.

As soon as this is done, I'll ask you to eventually validate the PR.

nmellado:
I've added nanogui as submodule, that is compiled/used only if glwf can be found on the system.
Tested only on linux, please test and fix for other systems.

louen:
So do we leave this open ?

nmellado:
Update: This application needs to be moved to https://github.com/STORM-IRIT/Radium-NanoGUIApp

nmellado:
Working example is now live at: https://github.com/STORM-IRIT/Radium-NanoGUIApp
Can someone can try to fetch and compile it, so we can close this PR ?

nmellado:
Closing, this code has been moved to https://github.com/STORM-IRIT/Radium-NanoGUIApp.

Issue #228, opened 2017-09-08
Loader issue: faces are missing


nmellado:
Some files are only partially loaded : [io_crash_models.zip](https://github.com/AGGA-IRIT/Radium-Engine/files/1288271/io_crash_models.zip)

The bug has been introduced by accepting PR #227 .

PR #229, opened 2017-09-11
Mesh feature tracking plugin

hoshiryu:
Adding class FeaturePickingManager to handle selection of vertices, edges and triangles on meshes.
Adding a new Plugin aiming at displaying the selected feature informations, as well as a 3D sphere at the feature location.
Also adding a doc file to explain how this works.

This PR also adds a "pickable" flag in RenderObjects so that UI stuff cannot be selected.
However, there are a few things that may be inconvenient:
 - the signal QItemSelectionManager::currentChanged() have to be manually emitted when selecting a feature, due to QitemSelectionManager::setCurrentIndex() not emitting it (still unkown reason) ;
 - when interating with the vertex/triangle index widgets in the "Tracking" tab, the main window calls SelectionManager::setCurrentEntry() for plugins to be told about the new feature selection ;
 - the MainApplication's m_viewer has to be set before loading plugins since it owns the FeaturePickingManager;


nmellado:
@hoshiryu , is the PR ready for review ?

hoshiryu:
Yes!
It's acceptance also depends on discussion about the inconvenient things.

nmellado:
I've a question regarding the `SelectionManager`.

Is there a specific reason this is not a singleton ?

louen:
> Is there a specific reason this is not a singleton ?
Because we don't need to access it globally ?

nmellado:
It is. 
It's now part of the `PluginContext`, which looks like a global access, no ?

hoshiryu:
This should go in an issue regarding the `PluginContext` and singletons

louen:
AFAIK the plug-in context is explicitly passed to all plug-ins at startup.
It could be a singleton, but I think it is good to minimize the amount of global access variables.

nmellado:
I've done some testing:
* the radius of the sphere showing the selection must adapt to the size of the model (e.g. relatively to the model bounding-box size), 
* the selection is not working with point cloud, even in vertex mode.

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+    Scalar MeshFeatureTrackingComponent::getFeatureScale() const
+    {
+        if (m_data.m_featureType == Ra::Engine::Renderer::RO)
+        {
+            return 1.0;
+        }
+        auto ro = Ra::Engine::RadiumEngine::getInstance()->getRenderObjectManager()->getRenderObject(m_data.m_roIdx);
+        return ro->getAabb().diagonal().squaredNorm() / 2000;
+#if 0
```
hoshiryu: The sphere radius is already computed according to the aabb diagonal

hoshiryu:
This PR is about **Mesh** Feature selection/tracking, so it's not intended to work on point clouds.
Moreover, the feature selection is performed through raycasting on the mesh triangles, preventing direct adaptation to point clouds.
A better approach would be to use OpenGL picking through shaders, so that even points in point clouds could be selected.

nmellado:
> A better approach would be to use OpenGL picking through shaders, so that even points in point clouds could be selected.

We already have that in Radium, no ?

nmellado:
These commits add support for vertex picking on point clouds + compute sphere size linearly wrt bb diagonal size.
@hoshiryu could you check the sphere size is ok on you side, and fix it if needed.

hoshiryu:
It's ok for me.
But unfortunately, when solving the conflicts through github, it merged master with issue #228 .

PR #231, opened 2017-09-19
Fix missing faces when loading quad meshes

### LINKED ISSUE #228: Loader issue: faces are missing

nmellado:
Due to unwanted side effect when disabling automatic triangulation in assimp loader.
Fixes #228

### DIFF
--- a/src/IO/AssimpLoader/AssimpFileLoader.cpp
+++ b/src/IO/AssimpLoader/AssimpFileLoader.cpp
@@ -50,6 +50,7 @@ namespace Ra {
 
             const aiScene *scene = m_importer.ReadFile( fileData->getFileName(),
                                                         aiProcess_GenSmoothNormals      |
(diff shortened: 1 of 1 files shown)


hoshiryu:
There is another thing to fix: after 1 animation step, the mesh is flat rendered.

hoshiryu:
Ok just found out why the mesh is flat rendered.
This is due to the activation of `RADIUM_WITH_TEXTURE`, which makes the `AssimpGeometryDataLoader` to load duplicate vertices.

nmellado:
Ok, so what do we do ? You just need to disable textures and it's good, right ?

PR #232, opened 2017-09-20
Line Rendering of Point-clouds

nmellado:
PR is not ready for merging yet.

In its current state, the codes forces to use `TempRenderer`, which implements the implicit brush technique. It crashes when meshes are loaded:
```
ERROR: OpenGL error (/export/home/cezanne/nmellado/git/Radium-Engine/src/Engine/Renderer/Mesh/Mesh.cpp:52, glDrawElements( static_cast<GLenum >(m_renderMode), m_numElements, GL_UNSIGNED_INT, (void*)0 )) :  Invalid operation : The specified operation is not allowed in the current state. The offending command is ignored and has no other side effect than to set the error flag.
(GL_INVALID_OPERATION : 0xGL_INVALID_OPERATION).
```

## TODO
* Rename `TempRenderer`, for instance as ImplicitBrushRenderer`
* Allow to use both meshes and point clouds (fix crash). This requires to evaluate the APSS reconstruction only for point-clouds. 
Suggestion: add a general mechanism to produce `position` and `normal` render outputs, for each type of objects.
 * Add ImplicitBrushRenderer to the list of default renderers in MainApplication


nmellado:
@dlyr I'm not sure we will have time to do the edits to merge this PR, as it might require some efforts to catch-up with `master-v1`.
Also, this implementation might be deprecated with the changes that will be introduced in Radium V2.

What do you think ?
I'm not happy with closing this issue, however being realistic I'm not sure we will have time to process it.

PR #235, opened 2017-09-26
Normalize skinning weights at load to ensure they sum to 1

hoshiryu:
Some models we have do not provide normalized weights.
While model files should contain normalized weights, it is safer to ensure it at load.


nmellado:
Could someone review this PR please ?

Issue #236, opened 2017-09-26
Animation/Skinning of textured objects


hoshiryu:
The actual animation/skinning mecanism recomputes a vertex normal from its one-ring.
While working fine for non textured objects, it fails to give the correct normal for textured objects, where vertices are duplicated for each face.
Even though there are util functions providing the computation of a duplicate table, it would be better not to discard the one that is computed when loading the file.

hoshiryu:
The thing is that when texture loading is enabled, then the duplicates are also loaded, in the sense that `duplicateTable[i] == i` so that edges and faces are loaded correctly w.r.t. texture coordinates.
Thus, the duplicateTable loses all meaning and cannot be used anymore to correctly compute the normals after editing the mesh (vertices for skinning, faces for decimation, ...)

A possible work around could be to always compute the duplicate table as if not loading duplicates, and use the corresponding flag of GeometryData to determine which index to put in edges and faces.

dlyr:
Why the normals are recomputed ?
For animation the normals should be transformed along with the vertices right ?
Moreover I think if a topological mesh is needed, it should be created at load time, with an association table from topoMesh -> viewMesh ... having a duplicateTable is misleading I think.

nmellado:
I do agree with @dlyr, it is not clear to me why the normals need to be recomputed. 

Regarding topological meshes, we have the work by Xavier submitted in another PR.

hoshiryu:
I agree that in the case of simple geometric skinning methods (LBS, DQS, CoR) the normals can be transformed along with the vertices since we know the applied transformation.
However, in some cases (implicit skinning, mesh manipulation) the normals has to be recomputed w.r.t. the neighborhood of the vertices.

Regarding topological meshes, since the aforementioned PR is not ready for merge by now, we have to manage things somehow, thus the use of a duplicate table.

nmellado:
Fixed by #239.

PR #239, opened 2017-10-08
Normal computation using duplicateTable

hoshiryu:
This PR solves issue #236 and does some code cleanup.

Regarding duplicate tables:
* duplicate tables are now all `std::vector<uint>` ;
* the one in `GeometryData` is meant to deal with vertices indices in the loaded `FileData`, i.e. `table[i] == i` when loading duplicates ;
* the one added in `FancyMeshComponent` is meant to deal with vertices indices in the final mesh, i.e. `table[i] == i` when NOT loading duplicates, and be used whenever needed by other components.

Regarding normals computation, an utility function has been added to take into account vertex duplicates across faces.

Regarding the code cleanup:
* add missing include ;
* remove useless spaces, align/beautify code, typos ;
* remove useless data copy in `FancyMeshComponent::handleMeshLoading`;
* transpose modifications in `FancyMeshComponent::handleMeshLoading` to `RenderObject::createFancyFromAsset` ;
* allow to set `GeometryData`'s duplicate table in-place ;
* update edges and faces data w.r.t. duplicate table only if needed ;
* simplify `AnimationComponent::createWeightMatrix` .


Plugins/FancyMesh/src/FancyMeshComponent.cpp
```diff
@@ -211,6 +218,11 @@ namespace FancyMeshPlugin
         return &(getMesh());
     }
 
+    const std::vector<uint>* FancyMeshComponent::getDuplicateTableOutput() const
```
nmellado: I don't like the idea of returning a pointer like that (allows calls to delete). What about returning a const reference ?

Plugins/FancyMesh/src/FancyMeshComponent.hpp
```diff
@@ -72,6 +73,9 @@ namespace FancyMeshPlugin
         const Ra::Core::Index* roIndexRead() const;
 
     private:
+        // the duplicate table for vertices, according to mesh data.
+        // Note: if duplicates have NOT been loaded, this is a 1-1 correspondance, i.e. m_duplicateTable[i] == i .
```
nmellado: For the sake of clarity, can you explain what I need to do if I want to iterate over the original mesh vertices (assuming I don't care about the loading options).
I guess this should be something like:

``` c++
for (uint id : *( component->getDuplicateTableOutput())) {
   auto & v = component->getVerticesRw()[id];
   // do stuff
}
```

If I want to skip the duplicates, I guess I need to do:

``` c++
for ( auto & v : *(component->getVerticesRw())) {
   // do stuff
}
```

2 questions:
 1. Can you please confirm this ?
 2. What about providing convenience iterators for this, instead of requiring the user to  work with the duplicate table ? This would make the API simpler to use and to understand for users. 
Bonus: in case we know in advance duplicates will not be loaded, both iterators could by-pass the duplicate table, since we know it's encoding the identity in that case.

src/Core/Geometry/Normal/Normal.cpp
```diff
@@ -35,12 +35,50 @@ void uniformNormal( const VectorArray< Vector3 >& p, const VectorArray< Triangle
...
+    }
+}
+
+
+void uniformNormal( const VectorArray< Vector3 >& p, const VectorArray< Triangle >& T,
+                    const std::vector<uint>& duplicateTable, VectorArray< Vector3 >& normal ) {
+    const uint N = p.size();
+    normal.clear();
+    normal.resize( N, Vector3::Zero() );
 //#pragma omp parallel for
```
nmellado: Why is this commented out ?

src/Engine/Renderer/RenderObject/RenderObject.cpp
```diff
@@ -106,7 +106,8 @@ namespace Ra
             for (size_t i = 0; i < asset->getVerticesSize(); ++i)
             {
                 mesh.m_vertices.push_back(T * asset->getVertices()[i]);
-                mesh.m_normals.push_back((N * asset->getNormals()[i]).normalized());
+                if(asset->hasNormals())
+                    mesh.m_normals.push_back((N * asset->getNormals()[i]).normalized());
```
nmellado: Could you please reserve std::vectors capacity before calling push_back in for loops ?

src/IO/AssimpLoader/AssimpGeometryDataLoader.cpp
```diff
@@ -204,55 +204,54 @@ namespace Ra {
...
 
         /// EDGE
         void AssimpGeometryDataLoader::fetchEdges( const aiMesh& mesh, Asset::GeometryData& data ) const {
             const uint size = mesh.mNumFaces;
-            auto &edge = data.getEdges(); edge.resize(size);
+            auto &edge = data.getEdges();
+            edge.resize( size );
             #pragma omp parallel for
-            for( int i = 0; i < int(size); ++i ) {
+            for( uint i = 0; i < size; ++i ) {
```
nmellado: Using unsigned loop counters require OpenMP 3.x.
https://stackoverflow.com/questions/2820621/why-arent-unsigned-openmp-index-variables-allowed

Since we use this multiple time in the code, please check in `cmake` that OpenMP version is 3.x or above.

Plugins/FancyMesh/src/FancyMeshComponent.cpp
hoshiryu: This has to be so because of the way the ComponentMessenger works.
Better ask @louen about that

src/Core/Geometry/Normal/Normal.cpp
hoshiryu: I guess we can get rid of it together with the `#pragma omp critical` comments, because it doesn't seem to be faster.
( same for within the above method )

nmellado: Ok, so parallelization doesn't improve performances here.

src/IO/AssimpLoader/AssimpGeometryDataLoader.cpp
nmellado: Fixed by 9cecd62.

nmellado:
I'm approving the changes, even if the use of the duplicate table is not perfect and must be improved.
I suggest we accept the PR in its current state, and add an issue explaining what are the pro/cons of the current implementation, and what should be improved.

From my point of view, the main problem is that currently accessing normals and positions require access to the duplicate table, while it's not the case for other attributes. This might be very very confusing for users trying to access mesh vertices attributes.

nmellado:
@MathiasPaulin can you test this is working on macos ? We updated OpenMP configuration, so we need to check it is working on all plateforms.

MathiasPaulin:
Tested on MacOsX, both with apple Clang (no OpenMP support) and with gcc7 (OpenMp support).
Compile and run fine.
Lots of warning due to eigen ...

nmellado:
Ok great.
Eigen will is updated in another concurrent PR.

PR #245, opened 2017-11-16
MacosX, Retina screens,  hdpi scaling solved

MathiasPaulin:
On MacosX with Retina screens, this commit solve the OpenGL hdpi scaling problem (view port only half the size on each dimension)

Note that on my test machine, this results in (very) poor performances when rendering on a retina screen (2fps instead of 600 for the default grid rendering, same for different scenes), due to the underlying QT and openGL interaction model used in the main app.
The application can be dragged from retina to non retina but, as a resize event is not always associated to this drag (I've not success in identifying why), the user as to manually generate such an event (by resizing manually the window.

To solve this performance and resizing problem, next direction will be try to remove the dependency between the application and QOpenGLWidget (in the same way the integration of glBinding with Qt is demonstrated as it not exhibit this problem).

### DIFF
--- a/Applications/MainApplication/main.cpp
+++ b/Applications/MainApplication/main.cpp
@@ -6,6 +6,8 @@
 
 int main( int argc, char** argv )
 {
+    QApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+
     Ra::MainApplication app( argc, argv );
 
     const uint& fpsMax = app.m_targetFPS;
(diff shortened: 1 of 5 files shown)


nmellado:
I've side effect on my hdpi screen, Linux OS: it is not possible to reduce the size of the main app window, it goes out of the screen.
Impossible to maximize the window also.

MathiasPaulin:
You can remove this PR. The problem is solved in a better way in PR #255

Issue #251, opened 2017-11-29
Problem when cloning : submodule init failed


MathiasPaulin:
As everytime a big set of merges are done on the master branch, I remove my local radium repository and clone it again from github. This allow to test "fresh download" regularly.
This morning, after clonning, I've difficulties in initialising submodules.
Here is the output of the submodule init stuff : 

```
magohamoth2:Radium-Engine mathias$ git submodule init
Submodule '3rdPartyLibraries/Assimp' (https://github.com/assimp/assimp.git) registered for path '3rdPartyLibraries/Assimp'
Submodule '3rdPartyLibraries/Eigen' (https://github.com/RLovelett/eigen.git) registered for path '3rdPartyLibraries/Eigen'
Submodule '3rdPartyLibraries/OpenGL/glbinding' (https://github.com/cginternals/glbinding.git) registered for path '3rdPartyLibraries/OpenGL/glbinding'
Submodule '3rdPartyLibraries/OpenGL/glm' (https://github.com/g-truc/glm.git) registered for path '3rdPartyLibraries/OpenGL/glm'
Submodule '3rdPartyLibraries/OpenGL/globjects' (https://github.com/cginternals/globjects.git) registered for path '3rdPartyLibraries/OpenGL/globjects'
Submodule '3rdPartyLibraries/Pbrt' (https://github.com/mmp/pbrt-v3) registered for path '3rdPartyLibraries/Pbrt'
Submodule '3rdPartyLibraries/tinyply' (https://github.com/nmellado/tinyply.git) registered for path '3rdPartyLibraries/tinyply'
magohamoth2:Radium-Engine mathias$ git submodule update
Cloning into '/Users/mathias/Professionnel/SVN-IRIT-paulin/paulin/Developpement/Radium/Radium-Engine/3rdPartyLibraries/Assimp'...
Cloning into '/Users/mathias/Professionnel/SVN-IRIT-paulin/paulin/Developpement/Radium/Radium-Engine/3rdPartyLibraries/Eigen'...
Cloning into '/Users/mathias/Professionnel/SVN-IRIT-paulin/paulin/Developpement/Radium/Radium-Engine/3rdPartyLibraries/OpenGL/glbinding'...
Cloning into '/Users/mathias/Professionnel/SVN-IRIT-paulin/paulin/Developpement/Radium/Radium-Engine/3rdPartyLibraries/OpenGL/glm'...
[...]

PR #252, opened 2017-11-29
Hotfix issue #251

nmellado:
Update globject repos to master


nmellado:
@dlyr, are you sure it's valid ?
I don't want you to feel bad because you miss the review process twice ;)

MathiasPaulin:
I've clone the repo, switched to the branch and, from now, all is fine. I compile, then try to make a new branch (that is not working from master) and will make my final review after that ...

dlyr:
No i'm not ... but I don't want to test ... (actually it seems that the commit id of globjets is ok  	globjects @ 275f159 ) .. ok I have tested and it checkouts and compiles smoothly.

PR #254, opened 2017-11-29
Add TopologicalMesh to use OpenMesh

dlyr:
This PR add the creation of TopologicalMesh, that use OpenMesh, with a converter from/to basic Mesh.
The converter needs improvement to manager vertices attributs, but is yet working for "simple" vertex.
It also add a command line application to show how to convert and use TopologicalMesh and Mesh.

By the way something wet wrong during rebase, 
0d78a46 and d8eb7ee should be one commit with "adding command line application example"
-> FIXED (I rebase again ... git is magic, but I had to force push ...)


Applications/SimpleSubdivideExample/main.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+    else{        
+        Ra::Core::TriangleMesh mesh;
+        Ra::Core::TopologicalMesh topologicalMesh;
+        Ra::Core::OBJFileManager obj;
+        if(a.inputFilename.empty()){
+            mesh  = Ra::Core::MeshUtils::makeBox();
+        }
+        else{
+            obj.load(a.inputFilename, mesh);
```
hoshiryu: Note here that under the current implementation of the OBJFileManager, the mesh is considered triangle-only.
Hence, if the mesh has quadrangular (or other polygon types) faces, the loaded TriangleMesh might endup with holes.

src/Core/Mesh/Wrapper/Convert.cpp
```diff
@@ -132,6 +132,53 @@ void convert( const Dcel& dcel, TriangleMesh& mesh ) {
 }
 
 
+/// Particular conversion for a progressive mesh
+/// since some faces doesn't exist anymore
+void convertPM( const Dcel& dcel, TriangleMesh& mesh )
+{
+
+    // TODO !!!
```
hoshiryu: ?

src/Tests/CoreTests/Geometry/GeometryTests.hpp
```diff
@@ -36,7 +36,7 @@ namespace RaTests
 
             // Test line queries.
             RA_UNIT_TEST(  pointToLineSq(a, a, b-a) == 0.f, "distance from A to AB" );
-            RA_UNIT_TEST(  pointToLineSq(b, a, b-a) == 0.f, "distance from B to AB" );
+            RA_UNIT_TEST(  areApproxEqual(pointToLineSq(b, a, b-a), 0.f), "distance from B to AB" );
```
hoshiryu: Why changing this test only and not the first one too?

Applications/SimpleSubdivideExample/main.cpp
dlyr: Do not change the app right ? do we need/want to handle quad mesh ? 
I could use Assimp here, but simpler to just call OBJFileManager

src/Core/Mesh/Wrapper/Convert.cpp
dlyr: (don't know ...)

src/Tests/CoreTests/Geometry/GeometryTests.hpp
dlyr: right, I will fix that (i'm only responsible for the rebase/subdivider app, I do not rewrite nor read all the code ;))

dlyr: btw it should be equal or approx equal ?
if approxEqual I should change the segment tests also

Applications/SimpleSubdivideExample/main.cpp
hoshiryu: Yeah, just wanted to mention that won't work for quad meshes as is.

PR #257, opened 2017-12-05
Add TopologicalMesh using OpenMesh v2

dlyr:
replace david_openmesh.
It is rebase to master, and take comment from last PR into account
Please test windows compilation ;)


src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
```diff
@@ -0,0 +1,121 @@
...
+                v._normal = Core::Vector3(n[0], n[1], n[2]);
+
+                int vi;
+                vMap::iterator vtr = vertexHandles.find(v);
+                if(vtr == vertexHandles.end()){
+                    vi = vertexIndex++;
+                    vertexHandles.insert( vtr, vMap::value_type(v, vi));
+                    //meshVertices.push_back(v);
+                    out.m_vertices.push_back(v._vertex);
+                    out.m_normals.push_back(v._normal);
```
hoshiryu: Is it possible to allocate beforehand and modify here? would be faster

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
```diff
@@ -0,0 +1,121 @@
...
+                }
+                else{
+                    vi = vtr->second;
+                }
+                indices[i] = vi;
+                i ++;
+            }
+            out.m_triangles.push_back(Triangle(indices[0],indices[1],indices[2]));
```
hoshiryu: Same here

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+        uint num_halfedge = in.m_triangles.size() * 3;
+        for(unsigned int i=0; i<num_halfedge; i++){
+            Vector3 p = in.m_vertices[in.m_triangles[i/3][i%3]];
+            Vector3 n = in.m_normals[in.m_triangles[i/3][i%3]];
+
+            vMap::iterator vtr = vertexHandles.find(p);
+            TopologicalMesh::VertexHandle vh;
+            if(vtr == vertexHandles.end()){
+
```
hoshiryu: useless empty line

src/Tests/CoreTests/TopologicalMesh/ConvertTest.hpp
```diff
@@ -0,0 +1,86 @@
...
+            TriangleMesh mesh;
+            TopologicalMesh topologicalMesh;
+
+            //Test for close mesh
+            mesh  = Ra::Core::MeshUtils::makeBox();
+            Ra::Core::MeshConverter::convert(mesh, topologicalMesh);
+            Ra::Core::MeshConverter::convert(topologicalMesh, newMesh);
+            RA_UNIT_TEST( isSameMesh(mesh,newMesh), "Conversion to topological box mesh failed" );
+
+            //Test for mesh with bounderies
```
hoshiryu: typo: boundaries

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
dlyr: Not sure to catch the point here, could you commit what you want ?

nmellado: Just adding out.m_vertices.reserve(....) before the loop should be ok

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
nmellado: For line 67, it's better to use emplace_back

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
dlyr: ok, but there the size I can have is a upper bound (since I do not know the final number of vertices at the end).
Also remember this procedure will eventually be rewritten soon ;)

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
louen: You can actually `emplace_back({indices[0], indices[1], indices[2]})`, I think

src/Core/CoreMacros.hpp
```diff
@@ -223,7 +223,16 @@
...
+#   undef DEPRECATED
+# if defined(__GNUC__)
+#  if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) >= 40500 /* Test for GCC >= 4.5.0 */
+#    define DEPRECATED(msg) __attribute__ ((deprecated(msg)))
+#  else
+#    define DEPRECATED(msg) __attribute__ ((deprecated))
+#  endif
+# elif defined(__clang__)
+#  define DEPRECATED(msg) __attribute__ ((deprecated(msg)))
+# endif
```
louen: I believe that if we are to change this we should start using the C++14 `[[deprecated]]` attribute (as you did in the code below) and remove these macros.

src/Core/Mesh/DCEL/Face.hpp
```diff
@@ -41,6 +42,8 @@ class RA_CORE_API Face : public IndexedObject  {
     HalfEdge_ptr m_he; // The halfedge reference
 };
 
+
+
```
louen: these two empty lines make me sad

src/Core/Mesh/DCEL/HalfEdge.hpp
```diff
@@ -30,7 +31,7 @@ class RA_CORE_API HalfEdge : public IndexedObject {
               const HalfEdge_ptr& prev  = nullptr,
               const HalfEdge_ptr& twin  = nullptr,
               const Face_ptr&     f     = nullptr,
-              const Index&        index = Index::INVALID_IDX() ); // Build a complete Halfedge
+              const Index&        index = Index::INVALID_IDX() ); // Build a complete Halfedg
```
louen: Introducing a typo in a comment. How dare you !

src/Core/Mesh/Wrapper/Convert.cpp
```diff
@@ -132,6 +132,53 @@ void convert( const Dcel& dcel, TriangleMesh& mesh ) {
 }
 
 
+/// Particular conversion for a progressive mesh
```
louen: What is this code ?

src/Core/Mesh/Wrapper/TopologicalMeshConvert.hpp
```diff
@@ -0,0 +1,20 @@
+#ifndef MESHCONVERTER_H
+#define MESHCONVERTER_H
+
+#include <Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp>
+#include <Core/Mesh/TriangleMesh.hpp>
+
+namespace Ra {
+namespace Core {
+
+    //! Adapter class to convert between vortex::Mesh and vortex::TopologicalMesh
```
louen: `vortex::`   should be `Core::`

src/Tests/CoreTests/Geometry/GeometryTests.hpp
```diff
@@ -17,34 +17,34 @@ namespace RaTests
...
 
             Vector3 y,z;
             Ra::Core::Vector::getOrthogonalVectors(b-a, y, z);
 
             // Test line queries.
-            RA_UNIT_TEST(  pointToLineSq(a, a, b-a) == 0.f, "distance from A to AB" );
-            RA_UNIT_TEST(  pointToLineSq(b, a, b-a) == 0.f, "distance from B to AB" );
-            RA_UNIT_TEST(  areApproxEqual(pointToLineSq(na, a, b-a), 0.f), "point on the line.");
-            RA_UNIT_TEST(  areApproxEqual(pointToLineSq(nb, a, b-a), 0.f), "point on the line.");
+            RA_UNIT_TEST(  areApproxEqual(pointToLineSq(a, a, b-a), Scalar(0.)), "distance from A to AB" );
```
louen: What is the rationale behind these changes ? I think it makes sense to ask for the distance from A to B to be *exactly* 0. Were the tests failing on your machine ?

src/Tests/CoreTests/TopologicalMesh/ConvertTest.hpp
```diff
@@ -0,0 +1,86 @@
+#ifndef RADIUM_CONVERT_TESTS_HPP_
```
louen: A great unit test ! Good work !

src/Core/Mesh/Wrapper/Convert.cpp
dlyr: don't know, it was here from @Syn4ck, and I comment it ...

src/Core/Mesh/DCEL/HalfEdge.hpp
dlyr: (sometimes I hit keyboard without notice ...)

src/Tests/CoreTests/Geometry/GeometryTests.hpp
dlyr: nope, actually I don't care ... just one was change to use areApproxEqual, the other not.
Since there is no special case in pointToLineSq that test extremities, and float computation could have roundup errors, I think it is better to keep areApproxEqual

src/Core/Mesh/Wrapper/Convert.cpp
louen: `ProgressiveMesh` looks like something from Céline or Anahid's ?

src/Core/CoreMacros.hpp
dlyr: The point here is I don't want to change the macro, juste make it compatible with OpenMesh one.
Let this modification for next  PR.

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
dlyr: (outdated)

src/Core/Mesh/Wrapper/Convert.cpp
dlyr: Yes ... do you prefer to remove the commented code ?

src/Tests/CoreTests/Geometry/GeometryTests.hpp
louen: Well, it's more about our *specification* of what the result of the algorithm should be, because this unit test is implicitly a specification of the algorithm.

Do we want to make sure that the distance between a segment and one of its extremities is always *exaclty* equal to zero ? I think it makes sense (just like the distance from a point to itself should be exactly zero). If we decide this, the implementation *has* to make it so that the result cannot suffer from roundup errors (which is the case for the current implementation).

src/Core/Mesh/Wrapper/Convert.cpp
louen: Lets make sure this code isn't useful to anyone. If it is let them keep it somewhere else, then delete it. Let's not add dead code (its' already enough to maintain the living one :) )

src/Core/CoreMacros.hpp
louen: Yes, we should add an issue to remove them.

src/Tests/CoreTests/Geometry/GeometryTests.hpp
dlyr: from the code of pointToLineSq
                `return (dir.cross(q - a)).squaredNorm() / dir.squaredNorm();`
so `a` must return 0, `b` don't.
Is it more readable to have always approxEqual ? don't know. I let you choose and commit

src/Core/Mesh/Wrapper/Convert.cpp
dlyr: OK I delete it, if it is useful, it must have been somewhere else !

PR #259, opened 2017-12-05
doc to use radium-as-submodule of your own project

dlyr:
for review to check everything is correct ;)

### DIFF
--- a/Docs/radium-as-submodule.md
+++ b/Docs/radium-as-submodule.md
@@ -11,7 +11,7 @@ in the external directory.
 You can clone Radium, or add it as a submodule of your git project.
 For instance if you want Radium in `myproject/external/Radium-Engine`
 
-```
+```bash
 mkdir external
 git submodule add https://github.com/STORM-IRIT/Radium-Engine/ 
 external/Radium-Engine --recursive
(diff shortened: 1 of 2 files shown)


dlyr:
Could someone test validate this doc and comment or accept ?

Docs/radium-as-submodule.md
```diff
@@ -0,0 +1,90 @@
...
+
+You can also fork https://github.com/STORM-IRIT/Radium-AppExample 
+It already contains Radium as submodule and a build-radium.sh script
+in the external directory.
+
+## git submodule, or just clone
+
+You can clone Radium, or add it as a submodule of your git project
+git submodule radium in my own project
+For instance Radium in ./myproject/external/Radium-Engine
```
hoshiryu: I don't understand this sentence!

hoshiryu:
Some comments:
 - l.12: "." at the end of sentence.
 - l.33: "Add a line" -> "Add the following line"
 - l.40: "add ... that need" -> "Add the following line in the CMakeLists.txt of your applications that need Radium as a library:" + linebreak + "find_package(...)"
 - l.41: "If needed ..." -> "Setting the RADIUM_ROOT variable before the previous line might be needed to help cmake find the Radium library:" + linebreak + "set(RADIUM_ROOT ...)"
 - l.43: "def ... vars" -> "define ... variables:"
 - l.52: "access ..." -> "and give access ...through the following commands: "
 - l.78: "radium..." -> "Radium (in release and/or debug) using the following command (for Linux):"
Note: If Radium is set as a git submodule, then cmake might compile it w.r.t. the "ExternalProject_Add"
command.

I think we might also check that the findRadium.cmake clears the variables it doesn't declare as output, e.g. local variables to build the output ones.

Docs/radium-as-submodule.md
dlyr: which part ?
You can clone Radium -> `git clone https://github.com/STORM-IRIT/Radium-Engine.git`
or add it as a submodule of your git project
-> `git submodule add  https://github.com/STORM-IRIT/Radium-Engine.git external/Radium-Engine`
line 12 could be removed ...
I will reformulate

Docs/radium-as-submodule.md
```diff
@@ -88,3 +87,5 @@ mkdir build-radium-release
 cmake ../Radium-Engine -DCMAKE_BUILD_TYPE=Release
 make -j8
 ```
+Or add Radium to your project cmake process with `ExternalProject_Add`.
+
```
dlyr: @hoshiryu Do you have a valid example for this one ?
I prefer to compile radium on its own, as it happens to delete my build dir, but don't want to recompile radium. Both approaches are ok.

hoshiryu: Well, it's the same process as the one we use for Radium's external libraries.
If your project uses Radium only as an external lib, you only need to compile Radium once.
Maybe a Radium "install" might be needed, as we do for the external libs.

dlyr: could you add the valid example command to the doc ?
with external project in radium, assimp for instance, if I clean radium build dir, then next compilation will compile assimp right ?

hoshiryu: If you use make clean to clean your build dir it's fine, but if you `rm -rf build` you would have to rebuild it again.
A way to do it might be to check if the libs are already compile and call the ExternalProject_add command only if they are not.

Docs/radium-as-submodule.md
```diff
@@ -0,0 +1,131 @@
...
+            -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+            -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
+            -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+	# skip the install step	
+	INSTALL_COMMAND cmake -E echo "Skipping install step."
+)
+```
+Then you have to setup manually all the variables that would have been setted up by `find_package`:
+```
+    ${RADIUM_INCLUDE_DIR}
```
louen: Would be nice to add typical values for these variables (e.g. `RADIUM_INCLUDE_DIR`)  should point to the `src` folder of Radium Root, etc.

PR #260, opened 2017-12-06
Qt settings

MathiasPaulin:
This PR :

- Make keymapping configuration persistant among different main-app run
- Rename "Test0" Application as "HelloRadium"
- Delete unused (and obsolete) "OpenGLExperiment"
- Allows all Qt-based application to use the Viewer class.
    This make the "HelloRadium" application work again as it crashes because of the gizmomanager default construction/destruction by the viewer.
  - Removed gizmmomanager initialization from Viewer contructor
  - Add createGizmoManager method to the viewer
  - Modified main_app to create the gizmoManager after creating the viewer
  - Removed unneeded debug messages spread accros the code to find the bug.


src/GuiBase/Viewer/Viewer.cpp
```diff
@@ -190,31 +196,33 @@ namespace Ra
...
-        // It seems that this is not the case on Linux, where the (QWindow) window report a size of wxh and the
-        // initial viewport of size of 100x100. (at least on my Nvidia OpenGL 4.4 drivers)
-        // So we set here the initial viewport to the size of the window.
-        // Surprisingly, on MacOs, the wiewport follow the size of the window (I presume, as a consequence
-        // of the activation of the context by m_context->makeCurrent(this);). Not on Linux ... is it a Qt bunctionnality ?
-
-//        int qtViewport[4];
-//        glGetIntegerv(GL_VIEWPORT, qtViewport);
-//        LOG( logDEBUG ) << "Qt Viewport (Viewer/Qwindow): " << qtViewport[0] << '-' << qtViewport[1] << '+' << qtViewport[2] << '-' << qtViewport[3];
+            // FIXME (Mathias) : try to understand the reasons of so different functionalities and how is managed the vieport size on MAcOs
```
dlyr: Is it fixed or not ?
if yes, please remove these comments.
If not, maybe add an issue and just here the issue id

MathiasPaulin: We have tested event management on different plateform for different applications and all seems fixed. We can remove the comment, this is just a remainder in the case another Qt event problem arises in the future ... I'll commit the new comment with just the issue number ...

PR #263, opened 2017-12-07
[hotfix] Renderers crash

nmellado:
Fixes two problems:
* GL error when switching between to renderers at runtime (missing call to makeCurrent) 1672156
* Crash when starting main application:
  Renderers in plugins were added while GL was not initialized.
  This is fixed by allowing deferred initialization of renderers that are added before GL initialization.

### DIFF
--- a/Applications/MainApplication/Gui/MainWindow.cpp
+++ b/Applications/MainApplication/Gui/MainWindow.cpp
@@ -372,11 +372,13 @@ namespace Ra
     {
         // always restore displaytexture to 0 before switch to keep coherent renderer state
         m_displayedTextureCombo->setCurrentIndex(0);
-        m_viewer->changeRenderer(m_currentRendererCombo->currentIndex());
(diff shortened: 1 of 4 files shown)


nmellado:
Can one of the render guys have a look too ?

MathiasPaulin:
I've got a look, seems fine for me. I just complete weak guard for multi-threaded cases  and make some cosmetic style changes in viewer.cpp. Even if I've not test the change of renderer (neither native nor plugin) because I don't have one, this PR could be accepted if all is fine for other reviewers

nmellado:
To test it, you can modify one ok the plugins packaged with radium so it adds a ForwardRenderer.
I will test tomorrow on windows

nmellado:
Your changes looks good to me.
The commit track is a bit messy however (sh*t happens ;) ).

As all the changes are coherent, shall we squash and merge ?

PR #267, opened 2017-12-13
[hotfix + feature] Fix Qt signaling system for OpenGL initilisation + Allow Extension of OpenGL biotope by plugins

MathiasPaulin:
- [hotfix] HelloRadium works fine now on Linux and MacOs : fix problem due to wrong order of initialisation and confusion of Qt event. Small rewrinting of the MainApplication to follow this fix. These modifications were also usefull for the introduced feature.
- [Feature] Extend the Plugin Interface to notify the plugins once OpenGL is initialised. This allow plugins to interact with the Radium OpenGl system (adding shaders, configuring the context, ...)


MathiasPaulin:
[hotfix] was validated on Linux and MacOs
[feature], only on MacOs

nmellado:
FYI: both MainApp and HelloRadium currently crashes on Windows with this PR.
Before, MainApp was working, and HelloRadium was not crashing (but not really working: black screen).

I'm investigating.

PR #269, opened 2017-12-22
Picking enhancement

hoshiryu:
This PR:
 - unifies picking management. Picking of vertices (mesh or point cloud), and mesh edges and triangles is now performed using OpenGL rendering only.
 - removes "FeatureTracking" stuff from RadiumPluginInterface and MainApplication/MainWindow, which are moved to the MeshFeatureTracking Plugin (also working for point clouds).
 - provides a mesh painting plugin using the picking of a collection of vertices/edges/triangles within a screen-space circle.
 - enables picking when moving the mouse whil holding the corresponding button (way better paint experience than clicking all the time).
 - renders Point Clouds by calling glDrawElement( GL_POINTS ).
 - fix the loading of objects with no material (e.g. point clouds).
 - fixes some minor bugs.

Tested on Linux, meshes and point clouds.

Note: picking has not been tested on Line meshes
Note: as the current implementation goes, when using multipicking, the same feature appears in the PickingResult as many times as there are pixels it renders on.


hoshiryu:
This PR is ready for merge!

nmellado:
Great.
Can you explain how to test it ?
I just want to test how it behaves on HDPI screens.

hoshiryu:
You can see the doc for it ;)
There are 2 plugins:
* the Feature tracking plugin that allows to select one feature at a time (vertex, edge or triangle) by holding the corresponding key (V, E or T) while picking the RO (using the mouse right button)
* the PaintMesh plugin that allows to paint on the ROs of the scene. For this, one can use single picking as above or activate multipicking by pressing key C, which draws a mouse-tracking screen circle, and then pick the features as above.

PR #273, opened 2018-01-29
Material subsystem3

dlyr:
Rebase version of material_subsystem2


MathiasPaulin:
src/CMakeLists.txt has redundant set_property ... I've fixed it. I test with my external plugin that depends on this branch and I'll approve the PR.

src/Engine/Renderer/Material/MaterialConverters.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * Instruction on how to extend the material system
+ */
+///////////////////////////////////////////////
+////        Material converter system       ///
+///////////////////////////////////////////////
+
+namespace Ra {
+  namespace Engine {
+    namespace MaterialConverterSystem {
+
```
hoshiryu: Why "System"?

src/Engine/Renderer/Material/MaterialConverters.hpp
```diff
@@ -0,0 +1,83 @@
...
+      using AssetMaterialPtr = const Ra::Asset::MaterialData *;
+      using RadiumMaterialPtr = Ra::Engine::Material *;
+      using ConverterFunction = std::function<RadiumMaterialPtr(AssetMaterialPtr)>;
+
+      /** register a new material converter
+      *  @return true if converter added, false else (e.g, a converter with the same name exists)
+      */
+      bool registerMaterialConverter(const std::string &name, ConverterFunction converter);
+
```
hoshiryu: Why not using the Manager pattern or put these methods directly in the RadiumEngine?

MathiasPaulin: Manager pattern rely on the implementation of a singleton pattern that is not really robust in this case. More, using the manager pattern will require to have access to then "Engine" Object and thus, might restrict the usage of material converters for Radium rendering applications.

src/Engine/Renderer/Material/MaterialConverters.hpp
MathiasPaulin: Change to EngineMaterialConverters (to be consistant with EngineRenderTechniques)

hoshiryu:
- Can we make the setter/getter for the material consistant in GeometryData (both pointer or both const &)?
- Regarding the use of the reset material along the setter/getter in RenderTechnique, may we add it everywhere a shared_ptr is used the same way?

MathiasPaulin:
In geometryData, seters and getters are not all consistant.
eg, for normals ...
`
           inline const Vector3Array &getNormals() const;

            // Copy data from normalList. In-place setting with getNormals is preferred.
            template<typename Container>

            inline void setNormals(const Container &normalList);
 `
Having a setter with a pointer means transferring ownership and a getter with a const ref allow to access a polymorphic object being sure it exist ...

If we want to be consistant for all setter/getter, this will require a big refactoring.

Perhaps the word "set" is not the good one. I can rename it to "store" ???

dlyr:
Is there any "TLDR" instruction to update code from old system to the new one.
I have some
                    ro->getRenderTechnique()->material->m_ns = Scalar(1);
                    ro->getRenderTechnique()->material->m_kd = color;
                    ro->getRenderTechnique()->material->m_ks = color2;
What should I use in place of this ?

MathiasPaulin:
As you use only BlinnPhongMaterial, you can use :
` auto mat = (BlinnPhongMaterial*)ro->getRenderTechnique()->getMaterial().get)
  mat-m_ns = ....
` 
or with a static_cast ....

MathiasPaulin:
Globally, using a concrete material through an abstract pointer is not a good idea. First, as you are sure that you use only blinnPhongMaterial, (static)(const)cast the ptr to what you really use.
Or wait until we implement introspection on Materials ....

dlyr:
Ok, I end up with  `static_cast<Ra::Engine::BlinnPhongMaterial*>ro->getRenderTechnique()->getMaterial().get()` which compiles (but not tested yet).

PR #276, opened 2018-02-02
Refactor index class

nmellado:
Uniformize Index use.
Set Index as constexpr class.

### DIFF
--- a/src/Core/Mesh/MeshTypes.hpp
+++ b/src/Core/Mesh/MeshTypes.hpp
@@ -3,6 +3,7 @@
 
 #include <Core/RaCore.hpp>
 #include <Core/Math/LinearAlgebra.hpp>
+#include <Core/Index/Index.hpp>
 
 namespace Ra
 {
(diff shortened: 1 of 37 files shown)


nmellado:
This PR is ready for review.

nmellado:
Fair enough

hoshiryu:
For the Index class methods, why do we have some defined in the .hpp and others in the .inl ?

src/Core/Mesh/MeshTypes.hpp
```diff
@@ -13,11 +14,9 @@ namespace Ra
         typedef Vector2ui Edge;
         typedef Vector3ui Triangle;
 
-        typedef uint TriangleIdx;
-        typedef uint VertexIdx;
-        typedef uint HalfEdgeIdx;
-
-        enum { InvalidIdx = uint( -1 ) };
+        using TriangleIdx = Index;
```
hoshiryu: why `using` for these and `typedef` for the two above?

MathiasPaulin: The question could be more general. Do we have to use the (old-style) typedef or the more C++ clean "using" for type aliasing ?
I prefer the second ... I canchange all the typedef to using if tis is OK ...

nmellado: Yes, please.

PR #277, opened 2018-02-03
Init an unitialized member value.

dlyr:
use 7, but could be set to whatever you want (always better than nothing ;)


nmellado:
@dlyr do you agree with this change ?

dlyr:
@nmellado perfect !

PR #278, opened 2018-02-03
Window factory

dlyr:
Add MainWindowFactory class to extend MainApplication.

MainApplication constructor take a factory to delegate the creation of MainWindow. This allows the creation of MainWindow derivate class without duplicating the MainApplication.
Next step is to provide a MainWindowInterface and move MainApplication in GuiBase.

(this PR also incorporate #277 )

### DIFF
--- a/Applications/MainApplication/MainApplication.cpp
+++ b/Applications/MainApplication/MainApplication.cpp
@@ -45,7 +45,7 @@
 
 namespace Ra
 {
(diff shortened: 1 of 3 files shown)


Zouch:
Makes sense

dlyr:
Thanks, maybe we can wait for #269 to be merged before accept this one (I think some conflict will arise ...)

PR #280, opened 2018-02-03
Guibase refactor

dlyr:
Move some base class to GuiBase (so that it makes easy to create other application ...)
todo : have two base mainapplication, one with very low level stuff, the other with more facilities but also more predefined behaviors.
To be merged after #278 if accepted


dlyr:
Well, it's "Qt" style, and it's on the application side (not in the GuiBase)

> Macro Documentation
qApp
A global pointer referring to the unique application object. It is equivalent to QCoreApplication::instance(), but cast as a QApplication pointer, so only valid when the unique application object is a QApplication.

It just redef the macro to have a ptr to *your* application rather than raw `QApplication`

dlyr:
Wait for at least @louen review before merge ?!

dlyr:
Seems I have some bugs (SelectionManager stuff)

nmellado:
Two PR are waiting for this one. 
Shall we accept them by the time you fix this one ?

dlyr:
Let's go
The PB I found is the fix (as suggested by clang) for `void SelectionManager::select` actually mask the correct `select` method. So one fix could just be to remove select.
Also in my other application I have trouble with picking manager, but do not affect main-app.
I would not have time today to deeply test it, so go one with other PR, I will check this tomorrow.

hoshiryu:
It seems to me that the current "BaseApplication" is not that basic, especially when looking at the MainApplication which just adds nothing to it.
I would rather set a class more like the MinimalApp to be the "BaseApplication" and have the "MainApplication" extending it with all the stuff we have for it.
Moreover, it may remove the GuiBase dependency on IO.

dlyr:
Well Base is not Basic to me, but have what a new "Qt" App need to have (I have this for the other application I develop, and override some behaviors.

MathiasPaulin:
Well, some questions ...
Applications/MainApplication/MainApplication.cpp only contains  #include <MainApplication.hpp>. Do we really need this MainApplication.cpp file ?
I agree with @hoshiryu, perhaps Base application is not so a base (It is not basic but seems to me be more than a Base). If this classe is call BaseApplication, it seems to me that it must be inherited to get some "Base" for an specific Application without redefining everything. So, this class might be rewritten to allow inheritance for extension ... I can do this in the next few days if you agree. 
What we could find in the "BaseApplication" : 
- Command line options : some seems to be general ("r", "m", "p", "l"), other are more application specifi ("f", "i", "n"). the Base clase could define a parser with the first, the derived class will add the last . 
- File loader might be added only by derived applications that want to use one (and so perhaps the loadfile method that could be virtual and default to nothing
- Setup Scene is purely application dependant (and ther is still a #FIXME in this method"
- addBasicShaders is purely application dependant

With this rewrinting of BaseApplication, the MainApplication.hpp/cpp could define a class that inherits BaseApplication and that add the specific part of RadiumMainApp (to be rename to something else by the way).

dlyr:
Ok, so as far as understand : 
Here the discussion is out of the scope of the PR (this PR do not have change BaseApplication at all ! it is the same as before !)
For my current project, I need to share this base application so it may goes to libGuiBase, this is the proposal of this PR.
We can correct some "mistakes" by the same occasion, but a deep modification should happens in a forthcoming PR e.g. moving setupScene to the MainWindow or viewer for instance.
So before accepting this PR I propose to:
- add virtual to the BaseApplication dtr (but it's not necessary since we never considered a derived class as a BaseApplication when considering data deletion ;) but it will prevent misuse of this class). The mainApp MACRO is defined in MainApplication.hpp such that it give access to MainApplication rather than BaseApplication.
- remove mainapplication.cpp
++d

nmellado:
As @dlyr suggested, this PR does not aim at providing a proper Base class for Radium apps, but rather to move some part of the code from Application to GuiBase.
As such, I suggest we work and evaluate this PR for what it does, and we add the BaseApplication to the Radium V1 project roadmap.

In any case, this PR merge  (as the others currently under review) need to be post-poned until the CI integration is properly set.

So please review this PR according to its current goal.

nmellado:
Looks like the dllexport/dllimport macro is missing on windows.

dlyr:
I was not aware of this one ... last commit should work ;)

dlyr:
Yeah it pass CI tests ;)

nmellado:
@hoshiryu @MathiasPaulin are you ok with this PR (see your previous comments).
If yes, we can merge this PR, and create and issue to discuss how to design a proposer BaseApplication class.

hoshiryu:
Ok for me

PR #291, opened 2018-02-18
At least one thread for taskQueue, change assert in SystemDisplay.

dlyr:
And a minor typo in comment.

### DIFF
--- a/src/Engine/Managers/SystemDisplay/SystemDisplay.cpp
+++ b/src/Engine/Managers/SystemDisplay/SystemDisplay.cpp
@@ -15,8 +15,6 @@ namespace Ra
         UiComponent* SystemEntity::uiCmp()
         {
             SystemEntity *instance=getInstance();
-            CORE_ASSERT(instance->getNumComponents() == 2 || instance->getNumComponents() == 3,
-                        "This entity should have only two or three components");
             CORE_ASSERT(instance->getTransformAsMatrix() ==
                         Core::Matrix4::Identity(), "Transform has been changed !");
             return static_cast<UiComponent*>(instance->getComponent("UI"));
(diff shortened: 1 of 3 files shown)


src/Engine/Managers/SystemDisplay/SystemDisplay.cpp
```diff
@@ -15,6 +15,13 @@ namespace Ra
         UiComponent* SystemEntity::uiCmp()
         {
             SystemEntity *instance=getInstance();
+            ///\todo Wow, A plugin add a component to systemEntity.
```
nmellado: I don't understand, is this a todo ? What should be done ?

src/Engine/Managers/SystemDisplay/SystemDisplay.cpp
```diff
@@ -25,11 +32,13 @@ namespace Ra
 #ifndef RA_DISABLE_DEBUG_DISPLAY
         DebugComponent* SystemEntity::dbgCmp()
         {
-            CORE_ASSERT(getInstance()->getNumComponents() == 2,  
-                        "This entity should have two components only");
-            CORE_ASSERT(getInstance()->getTransformAsMatrix() == 
+            SystemEntity *instance=getInstance();
+            /// And here also ;) of course
+            CORE_ASSERT(instance->getNumComponents() == 2 || instance->getNumComponents() == 3,
```
nmellado: Why 2 or 3 ? What changed here ?

hoshiryu: This is due to the MeshFeatureTrackingPlugin which adds the drawn sphere to the SystemEntity.
Since there was supposed to be only 2 components, I changed this to "2 or 3".
This might not be the best way to do it though.

dlyr: Yeah in fact it's more a reminder that something is "badly designed" than a todo.
Either one (i.e. Plugins) can add some component in SystemEntity so we do not need to check the number of components, either it is not allowed at all (but allow for "my nice plugin only" is not a good idea I think).
But since I'm not the author of theses modifications, I added the todo as a note for later. 
Feel free to change comment, remove the assert, change the way systemEntity work ;) but please at least accept the core_assert test modification so that we can use helper function again (right now if you try  RA_DISPLAY_SPHERE for instance, you have a assert false ! ;) )

nmellado: Sure, ok.
I don't really mind the practical changes in the code, but I would like we keep the code understandable and self contained.
For me, the comments are right now more confusing than helpful.

So please update the comment, such that it explains what is the problem and, if any, clearly state what are the todos.

dlyr:
I have fixed this issue ;)

PR #294, opened 2018-02-20
Topological mesh v2

dlyr:
So here is the PR for "feedback". It is not functionnal at all since I have commented out lot's of compil error generating lines.

One idea is to cleanup Meshes data structure
The other one is to have all the data in TriangleMesh (as discussed, it will be renamed mesh)
To this end it TriangleMesh have a VertexData array, that could be extended, and support things like normals, texcoord, weight.
Automatic vbo intsanciation and GPU communication will come after

### DIFF
--- a/src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
+++ b/src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
@@ -1,0 +1,14 @@
+#include <Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp>
+
+namespace Ra{
+namespace Core{
+
+Scalar dot(TopoVector3 a, TopoVector3 b){
+    return a.dot(b);
+
+}
+TopoVector3 cross(TopoVector3 a, TopoVector3 b){
+    return a.cross(b);
+}
+}
+}
--- a/src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
+++ b/src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
@@ -3,11 +3,11 @@
 namespace Ra{
 namespace Core{
 
-Scalar dot(TopoVector3 a, TopoVector3 b){
+Scalar dot(const TopoVector3 &a, const TopoVector3 &b){
     return a.dot(b);
 
 }
-TopoVector3 cross(TopoVector3 a, TopoVector3 b){
+TopoVector3 cross(const TopoVector3 &a, const TopoVector3 &b){
     return a.cross(b);
 }
 }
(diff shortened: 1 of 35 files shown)


dlyr:
So here is the PR for "feedback". It is not functionnal at all since I have commented out lot's of compil error generating lines.

One idea is to cleanup Meshes data structure
The other one is to have all the data in TriangleMesh (as discussed, it will be renamed mesh)
To this end it TriangleMesh have a VertexData array, that could be extended, and support things like normals, texcoord, weight.
Automatic vbo intsanciation and GPU communication will come after

dlyr:
Another solution would be to have a vector of properties, where we can dynammicaly add some (a la OpenMesh)
like 
add a prop
int texCoordProp = addProperty(new Property<Verctor3>, "texCoord")
access 
property(texCoordProp)[index]
or property("texCoord")[index]
and with an automatic vertexAttrib "texCoord" to shader in "texCoord"

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
```diff
@@ -0,0 +1,14 @@
+#include <Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp>
+
+namespace Ra{
+namespace Core{
+
+Scalar dot(TopoVector3 a, TopoVector3 b){
+    return a.dot(b);
```
nmellado: Minor comment: use `const TopoVector3 &` as input.

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp
```diff
@@ -1,35 +1,130 @@
...
 
-namespace Ra {
-namespace Core {
+namespace Ra
+{
+namespace Core
+{
+
+///\class TopoVector3 : small extension to Vector3 for OpenMesh compatibility
+class TopoVector3 : public Ra::Core::Vector3
```
nmellado: If we need this, we also might need implicit cast operators (in both direction) with `Ra::Core::Vector3`

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -11,31 +11,42 @@ namespace Ra
...
             VectorArray<Triangle> m_triangles;
 
         public:
             EIGEN_MAKE_ALIGNED_OPERATOR_NEW
         };
-    }
-}
 
+
+struct TriangleMesh: public TriangleMeshT<VertexData>{
```
nmellado: Would better be `using TriangleMesh = TriangleMeshT<VertexData>;`

nmellado:
> Another solution would be to have a vector of properties, where we can dynammicaly add some (a la OpenMesh)

I like this idea, looks very convenient.
It would be interesting to have this working the same way for point-clouds, i.e. to use mechanisms compatible with but independent from OpenMesh.

dlyr:
Replaced by #305

PR #296, opened 2018-02-23
Request providing changes we brought to the Radium for the Chef-d'Oeuvre*

Zarzwick:
*The PBRT plugin

Most important changes involve `Light`s that became `Component`s (to link them with Entities (which is useful only in the plugin so this can be discussed)). There is also a `LightManager` now, with a Dummy example that serve as an example, and `Material`s and `Light`s are provided with a method `getShaderInclude()` which is useful for our in-plugin shader "Template-like" system.


nmellado:
This code is part of PR#299 that is going to be merged soon.
I'm closing this one.

Issue #295, opened 2018-02-22
Picking not available on MacOs and Windows


MathiasPaulin:
Due to problem when rendering in picking mode, picking is not functionnal on MacOs and is quite hazardous on windows.
Picking texture, as displayed in the main app for instance, is always -1 (the initial value) whatever the picking mode or object.

hoshiryu:
can you test PR #298 on mac?

MathiasPaulin:
Tested .. Picking now works.
But I don't understand why culling causes no fragment written in the output. Need to investigate what are the values of in_position, in_eye and in_normal and why the dot product is always negative (note that you do not need to normalize as only the sign matters).

hoshiryu:
Fixed by #298 , closing

PR #298, opened 2018-02-28
Fix picking on Windows ;

hoshiryu:
Apparently, issue #295 is due to the "discard" command, at least on Windows.

### DIFF
--- a/Shaders/Picking.frag.glsl
+++ b/Shaders/Picking.frag.glsl
@@ -12,6 +12,7 @@ uniform int objectId;
 
 void main()
 {
+/*
     // dot product between view and normal
     float angle = dot(normalize(in_eye-in_position),normalize(in_normal));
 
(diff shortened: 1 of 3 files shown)


hoshiryu:
@MathiasPaulin can you check it's working on mac please?
By the way, should we add a PICKING pass to the RenderTechnique so that we don't have to split the renderqueue w.r.t. the render primitive? 
This would put extra, related stuff in the render method though, like is the RO "pickable", set the RO id as uniform and specific transform computations for UI ROs

hoshiryu:
checked on windows

MathiasPaulin:
checked on macos.
concerning the remark by @hoshiryu, I agree that picking must be refactored, but this must be done deeply and wrt the rendering process. I think that it is something that go over the current PR.
So, first, merge this bug-fix, then, open a new feature request for picking refactoring.

nmellado:
I do agree with @MathiasPaulin (see Contributing Guidelines).

Btw, it’s now time to work on Radium V1 RC, starting by modifying the filesystem and namespaces, which will break the API and thus need to be done before any other change.

PR #303, opened 2018-04-16
Change the selection manager clear when clearing the scene ;

hoshiryu:
When clearing the scene with a selected component, while having one of our private plugins enabled, makes the app crash.
Actually, the selection manager is told to clear its selection, then entities are destroyed. While destroying the entities', and thus their components', treeItems in the treeItemView, it somehow happens that the "currentChanged" signal of the selection manager is triggered.
This causes one of our plugins, which is connected on this signal, to crash since the "currentItem" of the selection manager isn't invalid.

Clearing the selection manager with "clear" prevents this crash (currentItem is an invalid one).
Also, it's coherent with the way we clear the selection manager when handling picking while not RO has been picked.

### DIFF
--- a/Applications/MainApplication/Gui/MainWindow.cpp
+++ b/Applications/MainApplication/Gui/MainWindow.cpp
@@ -563,7 +563,7 @@ namespace Gui
     void MainWindow::resetScene()
     {
         // To see why this call is important, please see deleteCurrentItem().
-        m_selectionManager->clearSelection();
+        m_selectionManager->clear();
         Engine::RadiumEngine::getInstance()->getEntityManager()->deleteEntities();
         m_viewer->resetCamera();
     }


nmellado:
Ok, if we accept this, I'll cherry-pick that commit to PR#299.
Good for you ?

hoshiryu:
Ok for me

nmellado:
Side note, in `MainWindow::deleteCurrentItem`, the selectionManager is also cleared:
```c++
void MainWindow::deleteCurrentItem() {
    ItemEntry e = m_selectionManager->currentItem();

    // This call is very important to avoid a potential race condition
    // which happens if an object is selected while a gizmo is present.
    // If we do not do this, the removal of the object will call ItemModel::removeItem() which
    // will cause it to be unselected by the selection model. This in turn will cause
    // the gizmos ROs to disappear, but the RO mutex is already acquired by the call for
    // the object we want to delete, which causes a deadlock.
    // Clearing the selection before deleting the object will avoid this problem.
    m_selectionManager->clear();
    if ( e.isRoNode() )
    {
        e.m_component->removeRenderObject( e.m_roIndex );
    } else if ( e.isComponentNode() )
    {
        e.m_entity->removeComponent( e.m_component->getName() );
    } else if ( e.isEntityNode() )
    { Engine::RadiumEngine::getInstance()->getEntityManager()->removeEntity( e.m_entity->idx ); }
}
```

How does this interact with your change ?

hoshiryu:
Actually, it goes the same way, first clear selection, then delete (which clears the ItemTree), hence the "see deleteCurrentItem" above the code I changed (which was not the same for some reason).

PR #305, opened 2018-04-24
Topological on master v1

dlyr:

Replace PR #294 
Feature, add attribs to mesh, and topological mesh.
The topological Mesh is intended to replace DCEL and other halfedge data structure with openmesh.
It uses Eigen as basic type, so no conversion is needed for vector and points.
It also merge duplicate 3D point but preserve by vertex normal definition.

Concerning the use of Mesh and TriangleMesh, everything should be seamless, but we will enforce the uses of a new behavior to create, acces, modify any properties ( i.e. texcoord, or user define).
See the upcomming doc

* **Other information**:
Need testing and discussion before accept


src/Core/Mesh/TopologicalTriMesh/Operations/EdgeSplit.cpp
```diff
@@ -31,20 +31,20 @@ void splitEdge( TopologicalMesh& topologicalMesh, TopologicalMesh::EdgeHandle ed
...
-                                     .normalized() );
+
+    //        TopologicalMesh::Normal n =  TopologicalMesh::Normal((fraction *
+    //        topologicalMesh.normal(v1) + (1. - fraction) *
+    //        topologicalMesh.normal(v2)).normalized());
 
     TopologicalMesh::VertexHandle vh = topologicalMesh.add_vertex( p );
-    topologicalMesh.set_normal( vh, n );
+    //        topologicalMesh.set_normal(vh,n);
     topologicalMesh.split( edgeHandle, vh );
```
nmellado: What's the expected behavior of the split operation regarding vertex attribute ?
Is the normal vector still handled ?

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp
```diff
@@ -1,33 +1,122 @@
...
 
 // Define an OpenMesh TriMesh structure according to
 // http://openmesh.org/Documentation/OpenMesh-2.1-Documentation/mesh_type.html
 // Attributes define data store on structure.
+
 struct TopologicalMeshTraits : public OpenMesh::DefaultTraits {
-    using Point = OpenMesh::VectorT<Scalar, 3>;
-    using Normal = OpenMesh::VectorT<Scalar, 3>;
+    typedef TopoVector3 Point;
+    typedef TopoVector3 Normal;
```
nmellado: Why switching back to `typedef` ?

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
```diff
@@ -27,7 +27,7 @@ void MeshConverter::convert( TopologicalMesh& in, TriangleMesh& out ) {
     };
 
     out.clear();
-    using vMap = std::map<vertexData, int, comp_vec>;
+    typedef std::map<vertexData, int, comp_vec> vMap;
```
nmellado: Why switching back from `using` to `typedef` ?

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
```diff
@@ -88,35 +89,50 @@ void MeshConverter::convert( const TriangleMesh& in, TopologicalMesh& out ) {
     // Delete old data in out mesh
     out = TopologicalMesh();
     out.garbage_collection();
+    out.request_halfedge_normals();
     out.request_vertex_normals();
-    using vMap = std::unordered_map<Vector3, TopologicalMesh::VertexHandle, hash_vec>;
+    typedef std::unordered_map<Vector3, TopologicalMesh::VertexHandle, hash_vec> vMap;
```
nmellado: Why switching back from `using` to `typedef` ?

src/Engine/Renderer/Mesh/Mesh.cpp
```diff
@@ -110,14 +107,32 @@ void Mesh::loadGeometry( const Core::Vector3Array& vertices, const std::vector<u
...
     m_isDirty = true;
 }
 
 void Mesh::addData( const Vec4Data& type, const Core::Vector4Array& data ) {
-    m_v4Data[static_cast<uint>( type )] = data;
-    m_dataDirty[MAX_MESH + MAX_VEC3 + static_cast<uint>( type )] = true;
+    const int index = static_cast<uint>( type );
+
+    /// to handle dummy set, if not valid, and dummy data, do nothing
+    if ( data.size() == 0 && !m_v4DataHandle[index].isValid() )
```
hoshiryu: Why is this only for Vec4Data ?

src/Engine/Renderer/Mesh/Mesh.inl
```diff
@@ -18,19 +19,29 @@ Core::TriangleMesh& Mesh::getGeometry() {
...
+    const int index = static_cast<uint>( type );
+    CORE_ASSERT( m_v3DataHandle[index].isValid(), "Attrib must be initialized" );
+    return m_mesh.attribManager().getAttrib( m_v3DataHandle[index] ).data();
 }
 
 const Core::Vector4Array& Mesh::getData( const Mesh::Vec4Data& type ) const {
-    return m_v4Data[static_cast<uint>( type )];
+    const int index = static_cast<uint>( type );
+    if ( !m_v4DataHandle[index].isValid() )
+        return m_dummy;
```
hoshiryu: Why are Vec3Data and Vec4Data managed ina different way?

dlyr:
Well it’s a hack (should have commented it) to make new changes work with old code (from paint plugin). The next changes will allow to query if color an attrib is defined and should take that into account when accessing or modifying an attrib. 

++ david

nmellado:
@dlyr , can you review my last changes ?
In addition to this, what else do we need to discuss/add before merging ?

src/Core/Mesh/Wrapper/TopologicalMeshConvert.cpp
dlyr: well, I forgot this one during merge/patch

src/Core/Mesh/TopologicalTriMesh/Operations/EdgeSplit.cpp
dlyr: not handle for now ... need further dev (but time is missing ...)

dlyr:
seems fine, could not test in deep.
I think we should get rid of the dummy thing quite quickly after other things are stabilized.

dlyr:
Because no one should ask for an attribute that is not set (and get the dummy), but we may provide a isValid or something to check attrib name validity.

nmellado:
It was just to get a consistent return type between the two getters (by name and handle). To initialize the reference you need a dummy, but we can static_assert When this occurres. So this make the compiler happy (reference initialization), but in practice no-one use the dummy because of the assert. Of course this needs to be documented, and the getDummy method set private.
I don’t like this because it makes the application crash on request, I prefer the current version.
Another solution could be to use pointers, but I don’t like the fact people can delete them.

dlyr:
Maybe another design would be : get an handle by name, then get the prop with this handle.
So if the named prop is not present, it returns an invalid handle.
Not sure what is a good design, I check (a little bit) how [OpenMesh](http://www.openmesh.org/media/Documentations/OpenMesh-Doc-Latest/a01168.html) implement this concept, but I think it is a bit more "complicated" and we could keep it "simple" according to our needs.

nmellado:
I've updated my proposal according to our discussion.
Now, I'm wondering if the `VertexAttribManager` class should not been moved to the `Core/Utils` package, and renamed as `AttribManager`, as there is nothing that prevents to use it in other contexts.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -1,38 +1,282 @@
...
+
+    /// RW acces to container data
+    inline Container& data() { return m_data; }
+
+    /// R only acccess to container data
+    inline const Container& data() const { return m_data; }
+
+    virtual ~VertexAttrib() { m_data.clear(); }
+    uint getSize() override { return Container::Vector::RowsAtCompileTime; }
+    int getStride() override { return sizeof( typename Container::Vector ); }
```
nmellado: @dlyr , what is the purpose of these two functions `getSize` and `getStride` ?

They assume a type `Container::Vector` is defined, but it is not always true (see `VectorArray<Scalar>` specialization [here](https://github.com/STORM-IRIT/Radium-Engine/blob/topological_on_master-v1/src/Core/Containers/VectorArray.hpp#L47).
This currently makes the CI on windows failing (only on windows, as GCC ignore these functions as they are not explicitly used.

dlyr: It's in preparation of the automatic GPU allocation for attribs (but since it's not functionnal yet, we can comment or remove them for now)

Docs/mesh.md
```diff
@@ -0,0 +1,50 @@
...
+        m.vertices().push_back( {0, 0, 0} );
+        m.vertices().push_back( {1, 0, 0} );
+        m.vertices().push_back( {0, 2, 0} );
+        m.normals().push_back( {0, 0, 1} );
+        m.normals().push_back( {0, 0, 1} );
+        m.normals().push_back( {0, 0, 1} );
+        m.m_triangles.push_back( {0, 1, 2} );
+
+        Vector3 dummy;
+        auto handle1 = m.attribManager().addAttrib( dummy, "vector3_attrib" );
```
dlyr: Need to be updated (this way of creation has been deprecated ;))

src/Core/Utils/Attribs.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+namespace Ra {
+namespace Core {
+
+template <typename T>
+class Attrib;
+
+/// AttribBase is the base class for Attrib.
+/// An Attrib is data linked to Vertices of a mesh
+/// It is also used for rendering, and
```
dlyr: unfinished comment ?

src/Core/Utils/Attribs.hpp
```diff
@@ -0,0 +1,213 @@
...
+class AttribBase {
+  public:
+    /// attrib name is used to automatic location binding when using shaders.
```
dlyr: is used -> will be used, but not implemented yet

src/Engine/Renderer/Mesh/Mesh.cpp
dlyr: now we have clarified handle and invalid data, here we should clean this hack

nmellado:
@dlyr, I've pushed fixes related to your comments.
I'll fix the compilation errors later (I cannot compile now).

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -57,7 +57,8 @@ int MeshFeatureTrackingComponent::getMaxV() const {
             ->getRenderObject( m_pickedRoIdx )
             ->getMesh()
             ->getGeometry()
-            .m_vertices.size();
+            .vertices()
+            .size();
     }
     return 0;
 }
```
MathiasPaulin: I don't like so much this kind of notation that call getters through pointers in cascade. What if on pointer is null somewhere ? why is the guard "if ( m_data.m_mode != PickingMode::RO && getRoMgr()->exists( m_pickedRoIdx ) )" sufficient ?

src/Core/Mesh/TopologicalTriMesh/Operations/EdgeSplit.cpp
MathiasPaulin: What is the interest of adding user defined attributes if they are not handled thoroughly by all operators ?
Apart time, what are the difficulties of developing such attribute management ?

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
nmellado: Well, I do agree.
The problem is that this plugin has been coded this way, and we didn’t try to fix such style issues.
I can add guard if really needed, but in any case this need to be fixed at many places in Radium source code, which is out of the scope of this PR.

src/Core/Mesh/TopologicalTriMesh/Operations/EdgeSplit.cpp
nmellado: Well... time. 
We believe this could be added in an upcoming PR.

nmellado:
Here are some changes we could add to this PR, to get a better API:

- [ ] Move `static MeshConverter::convert( const TriangleMesh& in, TopologicalMesh& out )` function to `TopologicalMesh::TopologicalMesh( const TriangleMesh& in )`,

- [ ] Move `static MeshConverter::convert( TopologicalMesh& in, TriangleMesh& out )` function to `TriangleMesh TopologicalMesh::convertToMesh() const`. This would require to add a move constructor to `TriangleMesh`.

- [ ] After discussion with @MathiasPaulin , we agreed that the `TriangleMesh` attributes need to be converted to the `TopologicalMesh`, and vice-versa. This would also require to modify the convert functions, which currently take only into account the normal vector and vertices positions.

- [ ] Also, the `SplitEdge` operation need to be updated to provide a simple interpolation mechanism for attributes. In a preliminary version, we can start by linear interpolation, and provide in the future another mechanism to let users define how they want to interpolate attributes.

louen:
Do we remove the old DCEL in the process ?

nmellado:
Not yet, as several users are still using it, and we do not offer equivalent tools with the current version.
So, we planned to let the two side by side until people have all they need with `TopologicalMesh`.
In the meantime, we con continue to work on `Core` refactoring, by improving attribute management and interfaces with `Engine`.

nmellado:
@louen, it looks like you missed the dllexport.

dlyr:
The PR still miss property management during SplitEdge (and of course during any openmesh provided  algorithm such as subdivision).

PR #313, opened 2018-05-14
Move the MeshFeatureTrackingSphere RO to the SystemEntity::uiCmp ;

hoshiryu:

Enables to fit the camera on an empty scene (though the point of view is different from the one at base application start)

When fitting the camera on an empty scene, the RenderObjectManager cannot compute the scene aabb because the MeshFeatureTracking sphere RO, which is a UI RO, is not counted as one.

Now the sphere RO belongs to the SytemEntity::uiCmp and is counted as an UI RO, hence the correct aabb computation on empty scenes.


nmellado:
I don't understand why the camer fitting problem on empty scene can be solved in a plugin. Can you explain the rational behind your changes ?
What if the plugin is not loaded ?

hoshiryu:
Well, as it is set now, a scene is considered empty when there are only ROs of the SystemEntity::uiCmp: `auto ui = Engine::SystemEntity::uiCmp();
 bool` skipUi = m_renderObjects.size() != ui->m_renderObjects.size();`

I agree that better way would be to consider all ROs of the SystemEntity, something like:
`const auto& comps = Engine::SystemEntity::getComponents();
 uint nbUiRO = 0;
 std::for_each(comps.begin(), comps.end(), [nbUiRO](const auto& c){ nbUiRO += c->m_renderObjects.size(); });
 bool` skipUi = m_renderObjects.size() != nbUiRO;`
If any RO is not considered as part of the empty scene while it should be, then the scene is not considered empty and skipUi is false, then leading to problems when evaluating this code:
`if (ro->isVisible() && (!skipUi || ro->getComponent() != ui))`

Furthermore, if plugins want to add UI ROs, they have to add them to the SystemEntity.

PR #335, opened 2018-06-05
Change the Eigen ref to a working one

Zarzwick:
When trying to clone the submodules it appeared that the current repo for Eigen (https://github.com/RLovelett/eigen.git) was dead, so this PR replaces the reference with one to https://github.com/eigenteam/eigen-git-mirror.


nmellado:
Merged as hotfix.
Don’t forget to work on master-v1 for regular use.

PR #345, opened 2018-06-27
Fix skinning

### LINKED ISSUE #343: Broken skinning
On branch master-v1, the skinning plugin is broken: neither LBS or DQS work correctly.

hoshiryu:

This PR fixes #343 , which was due to non deep copy of attributes in `TriangleMesh`.
Also fixes compilation error with AppleClang due to the c++17 standard modification regarding std::random_shuffle.

No deep copy of attributes when copying a `TriangleMesh`.

No copy of attributes when copying a TriangleMesh.
Copy of needed/all attributes from the corresponding methods, plus improved management of attributes addition/deletion in AttribManager.
Thus, there is no shallow copies anymore.

The `TriangleMesh` doesn't give access to its `AttribManager` anymore, it only exposes a reduced interface for it.

* **Other information**:

### DIFF
--- a/Docs/mesh.md
+++ b/Docs/mesh.md
@@ -45,10 +45,11 @@ For instance
         m.m_triangles.push_back( {0, 1, 2} );
 
         auto handle1 = m.addAttrib<Vector3>( "vector3_attrib" );
-        m.getAttrib( handle1 ).data().reserve( 3 );
-        m.getAttrib( handle1 ).data().push_back( {1, 1, 1} );
-        m.getAttrib( handle1 ).data().push_back( {2, 2, 2} );
-        m.getAttrib( handle1 ).data().push_back( {3, 3, 3} );
+        auto& buf = m.getAttrib( handle1 ).data();
+        buf.reserve( 3 );
+        buf.push_back( {1, 1, 1} );
+        buf.push_back( {2, 2, 2} );
+        buf.push_back( {3, 3, 3} );
 
         VertexAttribHandle<float> handle2 =
         m.addAttrib<float>( "float_attrib" );
--- a/Docs/mesh.md
+++ b/Docs/mesh.md
@@ -17,17 +17,20 @@ This class handles data only, in a way suitable for rendering.
 ## Mesh creation
 `FancyMesh` is in charge of loading a `GeometryData` and create the corresponding `Mesh`.
 
-The user can add on the fly new attributes to vertex via `VertexAttribManager` (accessible with `Mesh::attribManager()`).
-An attrib is defined by an unique name (`std::string`) and then represented as a `VertexAttribHandle<T>`.
+The user can add on the fly new vertex attributes via the `VertexAttribManager` (accessible with `Mesh::attribManager()`).
+An attrib is defined by a unique name (`std::string`) and then represented as a `VertexAttribHandle<T>`.
 The type of the attrib is defined by the caller using the template parameter of the method `addAttrib`.
 To ensure consistency, it is strongly recommended to store the handle returned by the `addAttrib` method:
 ```
         auto handle1 = m.attribManager().addAttrib<Vector3>( "vector3_attrib" );
 ```
 Note that handles of existing attributes can be retrieved by name (see `getAttrib(const std::string& name)`), however the caller have to provide the type of the associate attribute. 
-There is no type check on this access, and we recommend not to use this. We keep this merthod for convenience in some specific cases, however note that it might be marked as deprecated at any time.
+There is no type check on this access, and we recommend not to use this. We keep this method for convenience in some specific cases such as copy of attributes, however note that it might be marked as deprecated at any time.
 
 There is no (in the current version) data checking, and attribs are simple `std::vector` added to the attrib manager. It's meant to be used as vector of data as vertices and normals, with the same size and accessed with the faces indices. But please be aware no check nor allocation are done, so the allocation has to be performed on client side.
+
+When copying a `TriangleMesh`, only a shallow copy of its attributes is performed to fasten functions return.
+In order to perform a deep copy of some/all of the attributes, the dedicated methods must be used (directly using the `AttribManager` is highly discouraged).
 
 For instance
 ```
@@ -50,5 +53,8 @@ For instance
         m.attribManager().addAttrib<float>( "float_attrib" );
         auto attrib2 = m.attribManager().getAttrib( handle2 );
         attrib2.data().assign( { 1.f, 2.f, 3.f } );
+        
+        TriangleMesh m2;
+        m2.partialCopy( m, handle1 );
 ```
 
--- a/Docs/mesh.md
+++ b/Docs/mesh.md
@@ -17,20 +17,21 @@ This class handles data only, in a way suitable for rendering.
 ## Mesh creation
 `FancyMesh` is in charge of loading a `GeometryData` and create the corresponding `Mesh`.
 
-The user can add on the fly new vertex attributes via the `VertexAttribManager` (accessible with `Mesh::attribManager()`).
+The user can add on the fly new vertex attributes to a `TriangleMesh`.
 An attrib is defined by a unique name (`std::string`) and then represented as a `VertexAttribHandle<T>`.
 The type of the attrib is defined by the caller using the template parameter of the method `addAttrib`.
 To ensure consistency, it is strongly recommended to store the handle returned by the `addAttrib` method:
 ```
-        auto handle1 = m.attribManager().addAttrib<Vector3>( "vector3_attrib" );
+        auto handle1 = m.addAttrib<Vector3>( "vector3_attrib" );
 ```
+Attribute names `in_position` and `in_normals` are reserved.
 Note that handles of existing attributes can be retrieved by name (see `getAttrib(const std::string& name)`), however the caller have to provide the type of the associate attribute. 
 There is no type check on this access, and we recommend not to use this. We keep this method for convenience in some specific cases such as copy of attributes, however note that it might be marked as deprecated at any time.
 
-There is no (in the current version) data checking, and attribs are simple `std::vector` added to the attrib manager. It's meant to be used as vector of data as vertices and normals, with the same size and accessed with the faces indices. But please be aware no check nor allocation are done, so the allocation has to be performed on client side.
+There is no (in the current version) data checking, and attribs are simple `std::vector`s added to the attrib manager. It's meant to be used as vector of data as vertices and normals, with the same size and accessed with the vertices indices. But please be aware no check nor allocation are done, so the allocation has to be performed on client side.
 
-When copying a `TriangleMesh`, only a shallow copy of its attributes is performed to fasten functions return.
-In order to perform a deep copy of some/all of the attributes, the dedicated methods must be used (directly using the `AttribManager` is highly discouraged).
+When copying a `TriangleMesh`, no copy of its attributes is performed.
+In order to copy some/all of the attributes, the dedicated methods must be used.
 
 For instance
 ```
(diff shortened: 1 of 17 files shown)


src/Core/Utils/Attribs.hpp
```diff
@@ -138,19 +139,36 @@ class AttribManager
 {
   public:
     using value_type = AttribBase*;
-    using Container = std::vector<value_type>;
+    using Container = Ra::Core::IndexMap<value_type>;
```
dlyr: Are we sure to switch from vector to deque as container structure ?

hoshiryu: When adding the tests for deep copy, i run into issues regarding deleted attributes.
The thing with the vector, we have to manage attributes deletion in a correct way, which was not the case: the test application crashed (segfault) because some `Attrib *` were null, and thus I had to check for it in a loop like
```
for ( auto attr : manager.attribs() )
```
which I think shouldn't be mandatory.

Using an `IndexMap` provides a nice and simple way to deal with attribute suppresion.
However, regarding `Handle` invalidation, the map will segfault when trying to access a deleted attribute from an old `Handle` to it.
This might be avoided by returning a `nullptr` if a `Handle` points to an invalid map entry.
Also to be noted that if the entry pointed by the old `Handle` has been replaced, this might be a problem too.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -19,67 +17,83 @@ namespace Core
...
         return m_vertexAttribs.getAttrib( m_verticesHandle ).data();
     }
-    NormalAttribHandle::Container& normals()
-    {
-        return m_vertexAttribs.getAttrib( m_normalsHandle ).data();
+    const PointAttribHandle::Container& vertices() const {
+        return m_vertexAttribs.getAttrib( m_verticesHandle ).data();
     }
 
-    const PointAttribHandle::Container& vertices() const
```
nmellado: Why removing the `vertices () const` method ?

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -19,67 +17,83 @@ namespace Core
...
+    using Vec4AttribHandle = AttribHandle<Vector4>;
+    using Face = VectorNui;
 
     /// Create an empty mesh.
     inline TriangleMesh() { initDefaultAttribs(); }
-    /// Copy constructor and assignment operator
-    /// \warning No deep copy of attributes !
+
+    /// Copy constructor and assignment operator.
+    /// Note: shallow copy of attributes.
```
nmellado: From my point of view, keeping the shallow copy by default is misleading. Do you have a specific need for that ?

I see that could be useful to perform partial copies, in that case we need to have a mechanism for that.

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: actually it hasn't been removed, the style added the `{` at the end of the line

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: Well, I had a discussion with @dlyr about that point.
When doing a deep copy of a mesh, you may need only some attributes you need / are aware of, not all. I don't really remember all the discussion though, but I ended up restoring shallow copy as the default behavior.

nmellado: So we wait @dlyr is back before taking a decision.

src/Core/Mesh/TriangleMesh.hpp
nmellado: ok, my bad, the diff visualization was misleading

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: Actually, the shallow copy is highly appreciated as the default behavior for functions/methods returning a `TriangleMesh`.
Thus I left it as is, renamed `copyAttribs` into `partialCopy` and added a `fullCopy` method.

src/Core/Utils/Attribs.hpp
hoshiryu: Random access to a deque element is constant time (https://en.cppreference.com/w/cpp/container/deque)

src/Core/Utils/Attribs.hpp
```diff
@@ -206,21 +224,21 @@ class AttribManager
         Attrib<T>* attrib = new Attrib<T>;
         attrib->setName( name );
 
-        auto itr = std::find_if( m_attribs.begin(), m_attribs.end(),
-                                 []( AttribBase* b ) { return b == nullptr; } );
-        if ( itr != m_attribs.end() )
-        {
+        auto it = std::find_if( m_attribs.begin(), m_attribs.end(),
+                                [&name](const auto& a){ return a->getName() == name; } );
```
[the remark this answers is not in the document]
hoshiryu: No since the Attributes are now stored in an IndexMap

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -19,67 +17,89 @@ namespace Core
...
+    /// For deep copy of attributes, use the partialCopy or fullCopy methods.
+    /// Note: Directly using the AttribManager is highly discouraged.
     TriangleMesh( const TriangleMesh& ) = default;
     TriangleMesh& operator=( const TriangleMesh& ) = default;
 
+    /// Copy only the required attributes (deep copy).
+    /// The position and normal attributes are always copied, no need to provide their handles.
+    /// Note: The original handles are not valid for the mesh copy.
+    template <typename... Handles>
+    void partialCopy( const TriangleMesh& input, Handles... attribs );
```
nmellado: The name of the method is to me very confusing.
I would prefer something like `copyAttributes`, which express a bit more what's happening when the method is called.
Also, I don't see why the positions and normals are necessary copied. I could understand this for the positions, but in some cases you might have meshes without normals.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -19,67 +17,89 @@ namespace Core
...
+
+    /// Copy all the attributes (deep copy).
+    /// Note: The original handles are also valid for the mesh copy.
+    inline void fullCopy( const TriangleMesh& input );
```
nmellado: Same as above, I've found the name confusing. `copy` is enough I believe.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -19,67 +17,89 @@ namespace Core
...
+    /// The handle for normals, making request faster.
     NormalAttribHandle m_normalsHandle;
 
-    inline void initDefaultAttribs()
-    {
+    /// Sets the default attribs.
+    inline void initDefaultAttribs() {
         m_verticesHandle =
             m_vertexAttribs.addAttrib<PointAttribHandle::value_type>( "in_position" );
         m_normalsHandle = m_vertexAttribs.addAttrib<NormalAttribHandle::value_type>( "in_normal" );
```
nmellado: Do we always have normals ?

src/Core/Utils/Attribs.hpp
nmellado: >This might be avoided by returning a `nullptr` if a `Handle` points to an invalid map entry.

I'm ok with this. If @dlyr agrees too, it can be added to the code.

> Also to be noted that if the entry pointed by the old Handle has been replaced, this might be a problem too.

This needs to be mentioned in the user documentation

src/Core/Utils/Attribs.hpp
```diff
@@ -134,23 +127,70 @@ class AttribHandle
...
+    AttribManager& operator=( const AttribManager& m ) = default;
+
+    // Deep copy of attributes
+    /// Base copy, does nothing.
+    void partialCopy( const AttribManager& m ) {}
+
+    /// Deep copy of some attributes.
+    /// Note: if some attrib already exist, it will be replaced but not de-allocated.
+    template <class T, class... Handle>
+    void partialCopy( const AttribManager& m, const AttribHandle<T>& attr, Handle... attribs ) {
```
nmellado: Change names as requested for `TriangleMesh`

src/Core/Utils/Attribs.hpp
```diff
@@ -134,23 +127,70 @@ class AttribHandle
...
+        auto h = addAttrib<T>( a.getName() );
+        getAttrib( h ).data() = a.data();
+        // deal with other attribs
+        partialCopy( m, attribs... );
+    }
 
-    /// const acces to attrib vector
-    const Container& attribs() const { return m_attribs; }
+    /// Deep copy of all attributes.
+    void fullCopy( const AttribManager& m ) {
```
nmellado: Change names as requested for `TriangleMesh`

src/Core/Utils/Attribs.hpp
```diff
@@ -186,78 +225,73 @@ class AttribManager
...
-    {
-        return *static_cast<Attrib<T>*>( m_attribs[h.m_idx] );
+    inline const Attrib<T>& getAttrib( AttribHandle<T> h ) const {
+        return *static_cast<Attrib<T>*>( m_attribs.at( h.m_idx ) );
     }
 
-    /// Add attribute by name
+    /// Add attribute by name.
+    /// Note: If an attribute with the same name already exists,
+    ///  it will be replaced but not de-allocated.
```
nmellado: Add doxygen `\warning` command instead of `Note:`

src/Core/Mesh/TriangleMesh.hpp
dlyr: yes
if not, there are set to 0 ?
but normals are initialized everywhere as a always present attrib

src/Core/Utils/Attribs.hpp
dlyr: well, the index map access, with at(index), perform a find on the deque ... I'm not convienced here we can afford the performance penality

Docs/mesh.md
```diff
@@ -40,15 +44,18 @@ For instance
...
-        auto handle1 = m.attribManager().addAttrib<Vector3>( "vector3_attrib" );
-        m.attribManager().getAttrib( handle1 ).data().reserve( 3 );
-        m.attribManager().getAttrib( handle1 ).data().push_back( {1, 1, 1} );
-        m.attribManager().getAttrib( handle1 ).data().push_back( {2, 2, 2} );
-        m.attribManager().getAttrib( handle1 ).data().push_back( {3, 3, 3} );
+        auto handle1 = m.addAttrib<Vector3>( "vector3_attrib" );
+        m.getAttrib( handle1 ).data().reserve( 3 );
+        m.getAttrib( handle1 ).data().push_back( {1, 1, 1} );
+        m.getAttrib( handle1 ).data().push_back( {2, 2, 2} );
+        m.getAttrib( handle1 ).data().push_back( {3, 3, 3} );
```
nmellado: Can I do this: 
```
auto handle1 = m.addAttrib<Vector3>( "vector3_attrib" );
auto &buf = m.getAttrib( handle1 ).data();
buf.reserve( 3 );
buf.push_back( {1, 1, 1} );
buf.push_back( {2, 2, 2} );
buf.push_back( {3, 3, 3} );
```
The underlying question is: what is the cost of calling `m.getAttrib( handle1 )` many times ?

hoshiryu: Yes you can ;)

src/Core/Utils/Attribs.hpp
```diff
@@ -17,8 +16,7 @@ class Attrib;
 /// An Attrib is data linked to Vertices of a mesh.
```
nmellado: I'm not sure this line is true anymore.

hoshiryu: sure

Docs/mesh.md
nmellado: Can you add the complexity of these functions in the AttribManager methods doc ?
Also, it could be nice to refer from TriangleMesh to AttribManager for the methods that are delegated to AttribManager, something like:
```c++
    /// Get attribute by handle (const).
    /// \see AttribManager::getAttrib
    template <typename T>
    inline const Attrib<T>& getAttrib( AttribHandle<T> h ) const {
        return m_vertexAttribs.getAttrib( h );
    }
```

hoshiryu: yes, forgot the getters and copyAttributes

Issue #324, opened 2018-05-18
Design to allow animation/simulation plugins


hoshiryu:
I am currently working on a physics simulation plugin that I would like to couple with the animation plugin, so that both can play/pause/step/reset at the same time. I also would like to save/load the current state at some frame so that I can tweak some keyframed parameters without having to rerun the whole animation.
I already have some more-or-less dirty modifications of the Radium API with a keyframed value editor, but I would like to integrate a better, nicer design for this kind of situation.

What I have for now:
 - play/pause/step/reset methods in `Ra::Engine::System`
 - load/save methods in `Ra::Engine::Component`
 - my plugin's system is responsible for calling the load/save methods on all the mandatory components
 - Plus I had to find a way to get the app framerate

I would like to discuss with you about a nice design for this. Any idea?
Thanks

MathiasPaulin:
Well, before giving some design intuition, somme comments on the following points : 
- play/pause/step/reset methods in ``Ra::Engine::System``
I'm not sure that having play/pause/step/reset method in ``Ra::Engine::System`` is a good idea. This interface is specific to animation systems, or, if we want to be more general, specific to time-dependent systems. 
- load/save methods in Ra::Engine::Component
I don't understand what you want. A component is created by a system and has no self load/save power. This allow to set and verify **invariant** on the components and ensure that a system will always known the state of its component without requesting them.
- The app framerate is, by definition, application specific. So, the app must communicate this to the system, and not the opposite (the system poll it from the app). This will ensure that systems and component will be application independent. Useful to develop small application demo outside of the main-app and to re-use several piece of software in plugins.

What I'll prefer is the following : 
- a subclass of System that define the time-dependent interface. Let's call it ``TimedSystem```
- All the animation systems implement this interface so that an application or plugin that manage animation system could use this interface
- A specific ``TimedSystem`` that is just in charge of coupling several ``TimedSystem`` i.e. that manage the synchro between several ``TimedSystem`` by propagating the plat/pause/step/reset calls and that could do even more ...

[...]

PR #347, opened 2018-06-27
Coupling system

hoshiryu:

This PR adds `CouplingSystems`, which synchronize sub-systems and dispatch specific calls to them.
This Feature was requested from #324.

For example, the AnimationSystem is now a `CouplingSystem` synchronizing `TimedSystems` through animation-related calls (play/pause/step/cacheFrame/restoreFrame).
Also, for the cacheFrame/restoreFrame calls, a specific directory is created at compile time in the Bundle.

* **Other information**:
IT is preferable to merge #345 before this.


nmellado:
Is code ready for review ?

Plugins/Animation/CMakeLists.txt
```diff
@@ -70,3 +70,8 @@ target_link_libraries( Animation
                      )
 
 
+add_custom_target( Anim_data_dirs
+    COMMAND ${CMAKE_COMMAND} -E make_directory "${RADIUM_BUNDLE_DIRECTORY}/${CMAKE_BUILD_TYPE}/bin/animFrames"
```
nmellado: The value
```cmake
${RADIUM_BUNDLE_DIRECTORY}/${CMAKE_BUILD_TYPE}/bin
```
is already available as
```cmake
CMAKE_RUNTIME_OUTPUT_DIRECTORY
```
Please replace.

Plugins/Animation/src/AnimationComponent.cpp
```diff
@@ -393,4 +378,55 @@ Scalar AnimationComponent::getTime() const {
     return m_animationTime;
 }
 
+Scalar AnimationComponent::getDuration() const {
+    if ( m_animations.empty() )
+    {
+        return 0;
```
nmellado:
```c++
return Scalar(0);
```
to avoid implicit cast warnings.

Plugins/Animation/src/AnimationComponent.cpp
```diff
@@ -393,4 +378,55 @@ Scalar AnimationComponent::getTime() const {
...
+    }
+    return m_animations[m_animationID].getDuration();
+}
+
+void AnimationComponent::cacheFrame( int frame ) const {
+    std::ofstream file( "animFrames/"+m_contentName+"_frame"+std::to_string(frame)+".anim",
```
nmellado: Doesn't Radium has a standard repository for runtime file outputs ?

Plugins/Animation/src/AnimationComponent.cpp
```diff
@@ -393,4 +378,55 @@ Scalar AnimationComponent::getTime() const {
...
+        return;
+    }
+    file.write( (const char*) &m_animationID, sizeof(uint) );
+    file.write( (const char*) &m_animationTimeStep, sizeof(bool) );
+    file.write( (const char*) &m_animationTime, sizeof(Scalar) );
+    file.write( (const char*) &m_speed, sizeof(Scalar) );
+    file.write( (const char*) &m_slowMo, sizeof(bool) );
+    const auto& pose = m_skel.getPose( Ra::Core::Animation::Handle::SpaceType::LOCAL );
+    file.write( (const char*) pose.data(), sizeof(Ra::Core::Transform) * pose.size() );
+    std::cout << "Saving anim data at time: " << m_animationTime << std::endl;
```
nmellado: Use Radium Log.

Plugins/Animation/src/AnimationComponent.cpp
```diff
@@ -393,4 +378,55 @@ Scalar AnimationComponent::getTime() const {
...
+}
+
+bool AnimationComponent::restoreFrame( int frame )
+{
+    std::ifstream file( "animFrames/"+m_contentName+"_frame"+std::to_string(frame)+".anim",
+                        std::ios::in | std::ios::binary );
```
nmellado: Same as line 389

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -18,39 +18,87 @@ namespace AnimationPlugin {
 
 class SkeletonBoneRenderObject;
 
+/// The AnimationComponent is responsible for the management of the animations of an Entity.
+/// It stores the animation Skeleton and the animation data as well as the animation playing speed.
```
nmellado: So here, `Animation` means Skeletal Animation (or equivalent).
This cannot be used to animate an object (e.g. camera, position) ?

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -18,39 +18,87 @@ namespace AnimationPlugin {
...
+
+    /// Create the skeleton from the given data.
+    /// @param data: the skeleton's joint transform hierarchy.
+    /// @param duplicateTable: the map from duplicate to single vertices.
+    /// @param the number of single vertices.
+    // FIXME: the 2 last parameters are needed only for the genereation of
+    //        the skinning weights matrix.
+    void handleSkeletonLoading( const Ra::Asset::HandleData* data,
+                                const std::vector<Ra::Core::Index>& duplicateTable,
+                                uint nbMeshVertices );
```
nmellado: It is not clear to me why we still have the duplicate table.
Isn't the attribute system supposed to fix this ?

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -18,39 +18,87 @@ namespace AnimationPlugin {
...
-    void handleSkeletonLoading( const Ra::Asset::HandleData* data,
-                                const std::vector<Ra::Core::Index>& duplicateTable,
-                                uint nbMeshVertices );
-    void handleAnimationLoading( const std::vector<Ra::Asset::AnimationData*> data );
+    /// @returns the duration of the current animation.
+    Scalar getDuration() const;
+
+    // Skeleton display
+    /// Turns xray display on/off for the skeleton bones.
+    void setXray( bool on ) const;
```
nmellado: Is this really related to the animation ? Shouldn't this be a property of the skeleton ?

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -18,39 +18,87 @@ namespace AnimationPlugin {
...
+
+    /// Toggle skeleton bones display.
+    void toggleSkeleton( const bool status );
```
nmellado: Same as above

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -18,39 +18,87 @@ namespace AnimationPlugin {
...
+
+    /// @returns the index of the skeleton bone associated to the RenderObject with index \p index.
+    uint getBoneIdx( Ra::Core::Index index ) const;
```
nmellado: Same as above

Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -63,54 +111,86 @@ class ANIM_PLUGIN_API AnimationComponent : public Ra::Engine::Component {
     virtual void setTransform( Ra::Core::Index roIdx,
                                const Ra::Core::Transform& transform ) override;
 
-  public:
+private:
+    // Internal function to create the skinning weights.
+    void createWeightMatrix( const Ra::Asset::HandleData* data,
+                             const std::map<uint, uint>& indexTable,
+                             const std::vector<Ra::Core::Index>& duplicateTable,
+                             uint nbMeshVertices );
```
nmellado: same as above.

Plugins/Animation/src/AnimationComponent.hpp
nmellado: Update: after reading the rest of the code.
The animation component seems to be in charge of many other things, like creating the skeleton bones for display.
Please clearly state the role of this object.

Plugins/Animation/src/AnimationSystem.cpp
```diff
@@ -118,7 +152,7 @@ void AnimationSystem::handleAssetLoading( Ra::Engine::Entity* entity,
         }
 
         // FIXME(Charly): Certainly not the best way to do this
```
nmellado: What do we do with this FIXME(Charly) ?

src/Engine/System/CouplingSystem.hpp
```diff
@@ -0,0 +1,55 @@
...
+    /// Saves all the state data related to the \p frameID -th frame into a cache file.
+    virtual void cacheFrame( int frameID ) const = 0;
+
+    /// Restores the state data related to the \p frameID -th frame from the cache file.
+    /// \returns true if the frame has been successfully restored, false otherwise.
+    virtual bool restoreFrame( int frameID ) = 0;
+};
+
+/// Coupling Systems are responsible for transmitting calls to a bunch of other Systems.
+template <typename CSystem>
```
nmellado: Is `CSystem` always inheriting `CoupledSystem` ?
If yes, you don't need the template parameter.
If no, why do you need a template parameter, and why not using `System *` directly ?

Plugins/Animation/src/AnimationComponent.hpp
hoshiryu: Well, for now, the skinnnig weights are not stored as attributes but as a matrix.
And when creating this matrix, which has a row per duplicated position, we must know how to map the loaded indices (in GeometryData) to the stored indices (in TriangleMesh).

Plugins/Animation/src/AnimationComponent.hpp
hoshiryu: I would say, without taking a deep look at it, it should be renamed AnimationSkeletonComponent, though I have to investigate about its use for other animation purposes (e.g. camera, position, ...)

Plugins/Animation/src/AnimationSystem.cpp
hoshiryu: I don't know either, will take a look at when it was posted

src/Engine/System/CouplingSystem.hpp
hoshiryu: `CSystem` stands for `CoupledSystem` and which might provide the members required by the `CouplingSystem` managing it.
Having only `System*`, how can the `CouplingSystem` call the required members?

nmellado: So, just storing pointers of CoupledSystem should do the job, no ?

Plugins/Animation/CMakeLists.txt
hoshiryu: well, this is set to `RADIUM_PLUGIN_OUTPUT_PATH` in each plugin's cmake... is this something to change?

Plugins/Animation/src/AnimationComponent.cpp
hoshiryu: Actually, these runtime outputs are not relevant only for the current session, it's more like added data to the scene.
I will change that and have the user specify where he wants to save these data (better be along the scene file)

Plugins/Animation/src/AnimationComponent.hpp
hoshiryu: For now, the `AnimationPlugin` only deals with skeleton-based character animation.
Two choices: 
 - just specify this in the doc for now, and improve the animation management in another PR.
 - include it in this PR, but then postpone its review/acceptance to let other PRs pass.

Regarding the `AnimationComponent`, I don't really know how to improve it. A possible solution would be the following.
Let the `AnimationComponent` only store the `Core::Animation`s and current animation id and time. It will also expose the current animation and pose.
Then there could be an `AnimationSkeletonComponent` which would only store the `Core::Animation::Skeleton` and a map from bone id to transform storage index in the animation pose. It would also be responsible for updating the skeleton pose, through Component IO, and display it.

But I'm not sure about it, and if we want to handle other animation purposes, we need a deep reflection on how to deal with such cases, i.e. how to update the animated data.
Let's say we have a cube, with keyframed transforms. Then the `AnimationComponent` might store all the transforms, but how to / who should change the cube's transform when playing the animation? And what about animated cameras?
We need a mechanism to bind the animation transform to its use.

Plugins/Animation/CMakeLists.txt
nmellado: Interesting indeed.
Yes this needs to be fixed.
The variables (`CMAKE_RUNTIME_OUTPUT_DIRECTORY`,...) shouldn't be modified in each Plugin CMakeList.
If needed, directly use `RADIUM_PLUGIN_OUTPUT_PATH` and directly set the required targets properties.

hoshiryu: shall we do this in this PR? I remember reordering such things within the test_install* branches about the installation procedure of Radium.

nmellado: Ok, let's keep it for another PR, when we will rework the build system (install)

Plugins/Animation/src/AnimationComponent.hpp
nmellado: Please fix this in a dedicated PR.

hoshiryu:
Ready for review/acceptance

PR #359, opened 2018-09-10
Fix pr#345

MathiasPaulin:

In the assignment operators of TriangleMesh, add the missing return 

Fix wrong material binding when doing picking (segfault when doing picking in a complex scene with lot of different materials)

* **Other information**:

### DIFF
--- a/src/Core/Mesh/TriangleMesh.hpp
+++ b/src/Core/Mesh/TriangleMesh.hpp
@@ -90,7 +90,7 @@ class TriangleMesh {
         return m_vertexAttribs.findAttrib<T>( name );
     }
 
-    /// Return true if \p h referss to an existing attribute in *this.
+    /// Return true if \p h refers to an existing attribute in *this.
     template <typename T>
     bool isValid( const AttribHandle<T>& h ) const {
         return m_vertexAttribs.isValid( h );
--- a/src/Core/Mesh/TriangleMesh.hpp
+++ b/src/Core/Mesh/TriangleMesh.hpp
@@ -10,14 +10,13 @@
 namespace Ra {
 namespace Core {
 
-/// Simple Mesh structure that handles indexed triangle mesh with vertex
+/// Simple Mesh structure that handles indexed polygonal mesh with vertex
 /// attributes.
 /// Attributes are unique per vertex, so that same position with different
 /// normals are two vertices.
-/// Points and Normals are always present, accessible with
-/// points() and normals().
-/// Other attribs could be added with addAttrib() and
-/// accesssed with getAttrib().
+/// Points and Normals, defining the mesh geometry, are always present.
+/// They can be accessed through points() and normals().
+/// Other attribs could be added with addAttrib() and accesssed with getAttrib().
 /// \note Attribute names "in_position" and "in_normal" are reserved.
 /// \see MeshUtils for geometric functions operating on a mesh.
 class TriangleMesh {
@@ -35,19 +34,19 @@ class TriangleMesh {
     /// Create an empty mesh.
     inline TriangleMesh() { initDefaultAttribs(); }
 
-    /// Copy constructor, copy all the mesh data (topology, geometry, attributes).
+    /// Copy constructor, copy all the mesh data (faces, geometry, attributes).
     /// \note Handles on \p other are not valid for *this.
     inline TriangleMesh( const TriangleMesh& other );
 
-    /// Move constructor, copy all the mesh data (topology, geometry, attributes).
+    /// Move constructor, copy all the mesh data (faces, geometry, attributes).
     /// \note Handles on \p other are also valid for *this.
     inline TriangleMesh( TriangleMesh&& other );
 
-    /// Assignment operator, copy all the mesh data (topology, geometry, attributes).
+    /// Assignment operator, copy all the mesh data (faces, geometry, attributes).
     /// \warning Handles on \p other are not valid for *this.
     inline TriangleMesh& operator=( const TriangleMesh& other );
 
-    /// Move assignment, copy all the mesh data (topology, geometry, attributes).
+    /// Move assignment, copy all the mesh data (faces, geometry, attributes).
     /// \note Handles on \p other are also valid for *this.
     inline TriangleMesh& operator=( TriangleMesh&& other );
 
@@ -129,17 +128,17 @@ class TriangleMesh {
         m_vertexAttribs.removeAttrib( h );
     }
 
-    /// Erases all attributes, leaving the mesh with the topology and geometry only.
+    /// Erases all attributes, leaving the mesh with the faces and geometry only.
     inline void clearAttributes();
 
-    /// Copy only the mesh topology and geometry.
+    /// Copy only the mesh faces and geometry.
     /// The needed attributes can be copied through copyAttributes().
     /// \warning Deletes all attributes of *this.
     inline void copyBaseGeometry( const TriangleMesh& other );
 
     /// Copy only the required attributes from \p input. Existing attributes are
     /// kept untouched, except if overwritten by attributes copied from \p other.
-    /// \return True if the attributes have been sucessfully copied.
+    /// \return True if the attributes have been sucessfully copied, false otherwise.
     /// \note *this and \p input must have the same number of vertices.
     /// \warning The original handles are not valid for the mesh copy.
     template <typename... Handles>
@@ -147,7 +146,7 @@ class TriangleMesh {
 
     /// Copy all the attributes from \p input. Existing attributes are
     /// kept untouched, except if overwritten by attributes copied from \p other.
-    /// \return True if the attributes have been sucessfully copied.
+    /// \return True if the attributes have been sucessfully copied, false otherwise.
     /// \note *this and \p input must have the same number of vertices.
     /// \warning The original handles are not valid for the mesh copy.
     inline bool copyAllAttributes( const TriangleMesh& input );
--- a/src/Core/Mesh/TriangleMesh.hpp
+++ b/src/Core/Mesh/TriangleMesh.hpp
@@ -15,7 +15,7 @@ namespace Core {
 /// Attributes are unique per vertex, so that same position with different
 /// normals are two vertices.
 /// Points and Normals, defining the mesh geometry, are always present.
-/// They can be accessed through points() and normals().
+/// They can be accessed through vertices() and normals().
 /// Other attribs could be added with addAttrib() and accesssed with getAttrib().
 /// \note Attribute names "in_position" and "in_normal" are reserved.
 /// \see MeshUtils for geometric functions operating on a mesh.
@@ -90,7 +90,7 @@ class TriangleMesh {
         return m_vertexAttribs.findAttrib<T>( name );
     }
 
-    /// Return true if \p h correspond to an existing attribute in *this.
+    /// Return true if \p h referss to an existing attribute in *this.
     template <typename T>
     bool isValid( const AttribHandle<T>& h ) const {
         return m_vertexAttribs.isValid( h );
@@ -128,7 +128,7 @@ class TriangleMesh {
         m_vertexAttribs.removeAttrib( h );
     }
 
-    /// Erases all attributes, leaving the mesh with the faces and geometry only.
+    /// Erases all attributes, leaving the mesh with faces and geometry only.
     inline void clearAttributes();
 
     /// Copy only the mesh faces and geometry.
--- a/src/Core/Mesh/TriangleMesh.hpp
+++ b/src/Core/Mesh/TriangleMesh.hpp
@@ -36,17 +36,19 @@ class TriangleMesh {
     inline TriangleMesh() { initDefaultAttribs(); }
 
     /// Copy constructor, copy all the mesh data (topology, geometry, attributes).
-    /// \note The original handles are also valid for the mesh copy.
+    /// \note Handles on \p other are not valid for *this.
     inline TriangleMesh( const TriangleMesh& other );
 
     /// Move constructor, copy all the mesh data (topology, geometry, attributes).
+    /// \note Handles on \p other are also valid for *this.
     inline TriangleMesh( TriangleMesh&& other );
 
     /// Assignment operator, copy all the mesh data (topology, geometry, attributes).
-    /// \note The original handles are also valid for the mesh copy.
+    /// \warning Handles on \p other are not valid for *this.
     inline TriangleMesh& operator=( const TriangleMesh& other );
 
     /// Move assignment, copy all the mesh data (topology, geometry, attributes).
+    /// \note Handles on \p other are also valid for *this.
     inline TriangleMesh& operator=( TriangleMesh&& other );
 
     /// Appends another mesh to this one, but only if they have the same attributes.
@@ -89,6 +91,12 @@ class TriangleMesh {
         return m_vertexAttribs.findAttrib<T>( name );
     }
 
+    /// Return true if \p h correspond to an existing attribute in *this.
+    template <typename T>
+    bool isValid( const AttribHandle<T>& h ) const {
+        return m_vertexAttribs.isValid( h );
+    }
+
     /// Get attribute by handle.
     /// \see AttribManager::getAttrib() for more info.
     template <typename T>
@@ -129,19 +137,19 @@ class TriangleMesh {
     /// \warning Deletes all attributes of *this.
     inline void copyBaseGeometry( const TriangleMesh& other );
 
-    /// Copy only the required attributes from \p input.
+    /// Copy only the required attributes from \p input. Existing attributes are
+    /// kept untouched, except if overwritten by attributes copied from \p other.
     /// \return True if the attributes have been sucessfully copied.
     /// \note *this and \p input must have the same number of vertices.
     /// \warning The original handles are not valid for the mesh copy.
-    /// \see AttribManager::copyAttributes() for more info.
     template <typename... Handles>
     bool copyAttributes( const TriangleMesh& input, Handles... attribs );
 
-    /// Copy all the attributes from \p input.
+    /// Copy all the attributes from \p input. Existing attributes are
+    /// kept untouched, except if overwritten by attributes copied from \p other.
     /// \return True if the attributes have been sucessfully copied.
     /// \note *this and \p input must have the same number of vertices.
     /// \warning The original handles are not valid for the mesh copy.
-    /// \see AttribManager::copyAllAttributes() for more info.
     inline bool copyAllAttributes( const TriangleMesh& input );
 
   public:
--- a/src/Core/Mesh/TriangleMesh.hpp
+++ b/src/Core/Mesh/TriangleMesh.hpp
@@ -58,22 +58,22 @@ class TriangleMesh {
     inline void clear();
 
     /// Access the vertices positions.
-    PointAttribHandle::Container& vertices() {
+    inline PointAttribHandle::Container& vertices() {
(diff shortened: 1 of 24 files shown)


hoshiryu:
TODO:
 * TriangleMesh: 
  - copy (constructor and assignment): copy all attributes
  - copyAttributes: check nb vertices
  - add "copyBaseGeometry" to copy only topology and geometry

 * AttribManager:
  - rename "getAttrib" to "findAttrib"
  - addAttrib: use findAttrib, remove log warning, and use \note instead of \warning in doc
  - m_attribIndex: explain why in comment

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -106,24 +114,34 @@ class TriangleMesh {
...
-    /// Copy only the required attributes (deep copy).
-    /// The position and normal attributes are always copied, no need to provide their handles.
+    /// Erases all attributes, leaving the mesh with the topology and geometry only.
+    inline void clearAttributes();
+
+    /// Copy only the mesh topology and geometry.
+    /// The needed attributes can be copied through copyAttributes().
+    inline void copyBaseGeometry( const TriangleMesh& other );
+
+    /// Copy only the required attributes from \p input, keeping the old ones.
```
nmellado: I don't understand what means *keeping the old ones*.

nmellado: This comment must also describe what happens if the input handles have the same name than already existing attributes: is the attribute replaced ?

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -106,24 +114,34 @@ class TriangleMesh {
...
+    /// \return True if the attributes have been sucessfully copied.
```
nmellado: What can make this method return `false`?

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -106,24 +114,34 @@ class TriangleMesh {
...
-    void copyAttributes( const TriangleMesh& input, Handles... attribs );
+    bool copyAttributes( const TriangleMesh& input, Handles... attribs );
 
-    /// Copy all the attributes (deep copy).
-    /// \warning The original handles are also valid for the mesh copy.
+    /// Copy all the attributes from \p input, keeping the old ones.
+    /// \return True if the attributes have been sucessfully copied.
+    /// \note *this and \p input must have the same number of vertices.
+    /// \warning The original handles are not valid for the mesh copy.
     /// \see AttribManager::copyAllAttributes() for more info.
```
nmellado: Same comments as for copyAttributes

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -150,8 +168,10 @@ class TriangleMesh {
         m_normalsHandle = m_vertexAttribs.addAttrib<NormalAttribHandle::value_type>( "in_normal" );
     }
 
-    /// Access the attribute manager
+    /// Access the attribute manager.
     const AttribManager& attribManager() const { return m_vertexAttribs; }
+
+    /// Access the attribute manager.
     AttribManager& attribManager() { return m_vertexAttribs; }
```
nmellado: The `AttributeManager` cannot be visible as we restrict its interface in `TriangleMesh` (in_position, ...).
This method needs to be removed.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -150,8 +168,10 @@ class TriangleMesh {
         m_normalsHandle = m_vertexAttribs.addAttrib<NormalAttribHandle::value_type>( "in_normal" );
     }
 
-    /// Access the attribute manager
+    /// Access the attribute manager.
     const AttribManager& attribManager() const { return m_vertexAttribs; }
```
nmellado: The attribute cannot be visible as we restrict its interface in `TriangleMesh` (in_position, ...).
This method needs to be removed.

src/Core/Mesh/TriangleMesh.inl
```diff
@@ -36,55 +32,132 @@ inline TriangleMesh& TriangleMesh::operator=( TriangleMesh&& other ) {
...
+    for ( const auto& attr : o_attr )
+    {
+        if ( m_attr.find( attr.first ) == m_attr.end() )
+        {
+            sameAttrib = false;
+            break;
+        }
+    }
+    if ( !sameAttrib )
+        return false;
```
nmellado: The computation of `sameAttrib` is generic and must be implemented in `AttribManager`, not in `TriangleMesh`.

src/Core/Mesh/TriangleMesh.inl
```diff
@@ -36,55 +32,132 @@ inline TriangleMesh& TriangleMesh::operator=( TriangleMesh&& other ) {
...
+    for ( uint f = facesBefore; f < m_faces.size(); ++f )
+    {
+        for ( uint i = 0; i < m_faces[f].size(); ++i )
+        {
+            m_faces[f][i] += verticesBefore;
+        }
+    }
+
+    // Deal with all attributes the same way (vertices and normals too)
+    for ( auto& attr : other.m_vertexAttribs.attribs() )
```
nmellado: I would recommend to use a local template helper here, to reduce code duplication. So the loop would look like:
```c++
if ( attr->isFloat() ) { append_attrib<float>(attr); }
if ( attr->isVec2() )  { append_attrib<Vector2>(attr); }
if ( attr->isVec3() )  { append_attrib<Vector3>(attr); }
if ( attr->isVec4() )  { append_attrib<Vector4>(attr); }
```

The helper would be private, and look like
```c++
template <typename T>
void append_attrib (Attrib<T> attr) {
  auto h = m_vertexAttribs.findAttrib<T>( attr->getName() );
  auto& v0 = static_cast<Attrib<T>*>( &( m_vertexAttribs.getAttrib( h ) ) )->data();
  const auto& v1 = static_cast<Attrib<T>*>( attr )->data();
  v0.insert( v0.end(), v1.cbegin(), v1.cend() );
}
```

src/Core/Mesh/TriangleMesh.inl
```diff
@@ -21,13 +19,11 @@ inline TriangleMesh::TriangleMesh( TriangleMesh&& other ) :
     m_normalsHandle( std::move( other.m_normalsHandle ) ) {}
 
 inline TriangleMesh& TriangleMesh::operator=( const TriangleMesh& other ) {
+    clear();
```
nmellado: Why calling `clear` here ?
Isn't `m_vertexAttribs.clear();` sufficient ?

src/Core/Mesh/TriangleMesh.inl
```diff
@@ -36,55 +32,132 @@ inline TriangleMesh& TriangleMesh::operator=( TriangleMesh&& other ) {
...
+    initDefaultAttribs();
+}
+
+inline void TriangleMesh::clearAttributes() {
+    const auto v = vertices();
+    const auto n = normals();
+    m_vertexAttribs.clear();
+    initDefaultAttribs();
+    vertices() = v;
+    normals() = n;
```
nmellado: Wait, you mean that you **copy** both the normal and vertices arrays twice in a row ?
This looks terrible, what would be the performances on a 10 millions vertices mesh ?
May I suggest using `std::move` ?

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -106,24 +114,34 @@ class TriangleMesh {
...
     }
 
-    /// Copy only the required attributes (deep copy).
-    /// The position and normal attributes are always copied, no need to provide their handles.
+    /// Erases all attributes, leaving the mesh with the topology and geometry only.
+    inline void clearAttributes();
+
+    /// Copy only the mesh topology and geometry.
+    /// The needed attributes can be copied through copyAttributes().
+    inline void copyBaseGeometry( const TriangleMesh& other );
```
nmellado: There is a side effect in this function: all the attributes previously existing in the instance are destroyed.
This must be specified in the documentation.

src/Core/Utils/Attribs.hpp
```diff
@@ -290,8 +278,8 @@ class AttribManager {
     }
 
     /// Remove attribute by handle, invalidate the handles to this attribute.
-    /// \note: If a new attribute is added, old invalidated handles may lead to
-    ///        the new attribute.
+    /// \warning If a new attribute is added, old invalidated handles may lead to
```
nmellado: This is not clear. 
What is a old invalidated handle ?
How can two different handles lead to the same attribute ?

src/Core/Utils/Attribs.hpp
```diff
@@ -305,18 +293,18 @@ class AttribManager {
...
-        const auto& att = getAttrib( h ); // check the attribute exists
+        const auto& att = getAttrib( h );
         removeAttrib( att.getName() );
     }
 
   private:
     /// Access to the list of attributes.
-    /// \note The complexity for acessing the list of attributes is O(n).
+    /// \note The complexity for accessing the list of attributes is O(n).
     Container attribs() const {
```
nmellado: What is the purpose of this method ?

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: 1- By *keeping the old ones* I meant that the attributes present in the mesh bafore calling this method are still there after, but it might be misleading in some way, I will remove that
2- For existing attributes, they will be replaced, as mentioned in the doc for AttribManager which is referred to, but I can put it there and remove the link to AttribManager since it won't provide more info

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: Same answer ;)

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: I couldn't find a nice way to put it in the doc, so I used the *\note* command to explain that the meshes must have the same # of vertices.

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: This method is used by the TopologicalMesh, but since it is a friend class it is useless.
So yes I agree to remove it, @dlyr are you ok with that?

nmellado: It has no sense to keep this function, has it breaks the encapsulation needed to store the positions and normals.
If you keep it, anyone can delete or modify the positions and normals without any control.

src/Core/Mesh/TriangleMesh.inl
hoshiryu: yes

src/Core/Mesh/TriangleMesh.inl
nmellado: As this is purely internal and meant to avoid code redundancy, I would put the cast in the helper, as the call is a bit shorter.
But both options looks ok to me.

src/Core/Mesh/TriangleMesh.inl
hoshiryu: using std::swap?

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: true

src/Core/Mesh/TriangleMesh.inl
nmellado: `std::exchange` seems more adapted, as it moves data in one single direction, while swap... well, swaps in both directions (x to y and y to x).

src/Core/Utils/Attribs.hpp
hoshiryu: *old invalidated handles* refers to handles to removed attributes. And since Handles only store an index, nothing prevents:
```c++
AttribManager m;
auto h = m.addAttrib<float>( "f_attr" );
m.removeAttrib( h ); // h is invalid in the sense that no attrib exists at its index, but h.isValid() == true
auto h2 = m.addAttrib<float>( "f_attr" ); // here h and h2 have the same index
{
    auto& attr2 = m.getAttrib( h2 ); // ok since its the right handle
    auto& attr = m.getAttrib( h ); // still ok while should not, nothing to forbid it
}
m.removeAttrib( h ); // still ok...
auto h3 = m.addAttrib<Vector2>( "v2_attr" ); // h, h2, and h3 have the same index
{
    auto& attr3 = m.getAttrib( h3 ); // ok since its the right handle
    auto& attr = m.getAttrib( h ); // compilation error for type mismatch
}
```

src/Core/Utils/Attribs.hpp
hoshiryu: This methods filters out the empty slots (nullptr from removal) from the attribs list, so that one can perform the following for loop without caring about nullptr:
```c++
AttribManager m;
...
for ( auto& attr : m.attribs() )
{
    // no need to check whether attr == nullptr
}
```

src/Core/Utils/Attribs.hpp
nmellado: Hum, this is not nice, or at least confusing...

So, to rephrase: the handlers are not persistent, and their validity is affected by any change in the manager. More, a handler cannot know if it is valid or not, because it doesn't know the state of the manager.

To fix that, I suggest we add a method 
```c++
template<typename T>
bool AttribManager::isValid(const AttribHandle<T>& in) const;
```
which check that `in` **looks** valid, in the sense that its name and type are the same than what is stored in the manager (this check needs to be carefully described in the documentation). This means that a handle cannot tell by itself if it his valid or not, but the manager can. If we do that, we also need to remove
```c++
template <typename T> class AttribHandle {
constexpr bool isValid() const { return m_idx.isValid(); }
}
```

Using such mechanism, you should be able to do:
```c++
AttribManager m;
auto h = m.addAttrib<float>( "f_attr" );
static_assert(m.isValid(h), "h is supposed to be valid");
m.removeAttrib( h ); // h is invalid
static_assert(! m.isValid(h), "h is supposed to be invalid");
auto h2 = m.addAttrib<float>( "f_attr" ); // here h and h2 have the same index
{
    auto& attr2 = m.getAttrib( h2 ); // ok since its the right handle
    auto& attr = m.getAttrib( h ); // still ok, because names and types are the same
    static_assert(m.isValid(h), "h is supposed to be valid");
}
m.removeAttrib( h ); // both h and h2 are invalid
static_assert(! m.isValid(h), "h is supposed to be invalid");
auto h3 = m.addAttrib<Vector2>( "v2_attr" );
static_assert(! m.isValid(h), "h is supposed to be invalid");
{
    auto& attr3 = m.getAttrib( h3 ); // ok since its the right handle
}
```

hoshiryu: To prevent it, we could do:
```c++
template <typename T>
void removeAttrib( AttribHandle<T>& h ) {
    const auto& att = getAttrib( h );
    removeAttrib( att.getName() );
    h.idx = Index::invalid();
}
```

hoshiryu: yes, handles validity should be defined by the manager, which can also prevent:
```c++
AttribManager m1, m2;
auto h1 = m1.addAttrib<float>( "f_attr1" );
auto h2 = m2.addAttrib<float>( "f_attr2" );
m1.removeAttrib( h1 );
if ( m1.isValid( h2 ) )
{
    m1.getAttrib( h2 );
...
}
```

src/Core/Utils/Attribs.hpp
hoshiryu: If we use Handles as iterators, we could have private begin, end methods in the AttribManager and private increment methods in the Handles

src/Core/Utils/Attribs.hpp
nmellado: Ok for your second comment.

src/Core/Utils/Attribs.hpp
nmellado: I don't see the link with what I wrote.

nmellado: I bet it will take more than the 4 lines needed for the functor-based loop version.

src/Core/Utils/Attribs.hpp
hoshiryu: make removeAttrib take a ref then? it is a bit safer, though it nothing prevents copying the handle beforehand and then use the copy

nmellado: I'm ok to remove the method `removeAttrib` by name, and keep only the one accepting a handle.

hoshiryu: What about the isValid method?

src/Core/Mesh/TriangleMesh.hpp
nmellado: Ok, you can replace `keeping the old ones` by `existing attributes are kept untouched, except if overwritten by attributes copied from \p other.`

src/Core/Utils/Attribs.hpp
```diff
@@ -115,9 +115,14 @@ class AttribHandle {
  * ...
  *
  * // somewhere else: access
- * auto iattribhandler = mng.getAttribHandler<float>("MyAttrib"); //  iattribhandler == inputfattrib
- * if (iattribhandler.isValid()) {
- *    auto &attrib = mng.getAttrib( iattribhandler );
+ * auto iattribhandler = mng.findAttrib<float>("MyAttrib"); //  iattribhandler == inputfattrib
+ * if (iattribhandler.isValid()) { // true
```
nmellado: This method is supposed to be removed, and replaced by `mng.isValid(iattribhandler)`.

src/Core/Utils/Attribs.hpp
nmellado: We use the one I suggested, implemented by `AttribManager` as described above.

src/Core/Mesh/TriangleMesh.hpp
```diff
@@ -35,33 +35,34 @@ class TriangleMesh {
     /// Create an empty mesh.
     inline TriangleMesh() { initDefaultAttribs(); }
 
-    /// Copy constructor, copy only the mesh topology,
-    /// the list of vertices and vertices normals.
-    /// For attributes copy, use the copyAttributes() or copyAllAttributes() methods.
+    /// Copy constructor, copy all the mesh data (topology, geometry, attributes).
+    /// \note The original handles are also valid for the mesh copy.
```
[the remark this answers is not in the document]
hoshiryu: actually they may not be valid if there has been removals (nullptr are skipped by copyAllAttributes)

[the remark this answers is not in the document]
nmellado: What is `they` ?
I'm lost here.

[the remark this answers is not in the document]
hoshiryu:
```c++
TriangleMesh m1; // attribs: 0- position, 1- normal
m1.vertices() = vertices;
m1.normals() = normals;
auto h1 = m1.addAttrib<float>( "atttr1" ); // attribs: 0- position, 1- normal, 2- attr1
auto h2 = m1.addAttrib<float>( "atttr2" ); // attribs: 0- position, 1- normal, 2- attr1, 3- attr2
m1.removeAttrib( h1 ); // attribs: 0- position, 1- normal, 2- nullptr, 3-attr2
TriangleMesh m2 = m1; // attribs: 0- position, 1- normal, 2- attr2
```

src/Core/Mesh/TriangleMesh.hpp
hoshiryu: it has been!

PR #361, opened 2018-09-18
Keep background alpha when taking snapshots ;

hoshiryu:

* **Other information**:


dlyr:
Fine for me, 
but remember when we render we usually multiply color by alpha (even with alpha in the buffer), so the resulting png will not be displayed as what is on screen (it do not affect buffer where alpha is 1. everywhere)

Issue #360, opened 2018-09-14
TriangleMesh vs PolygonMesh


hoshiryu:
Currently, the mesh classes in Core `TriangleMesh` and `TopologicalMesh` exhibit interoperability issues.
On the one hand, a `TriangleMesh` is meant to store only triangular faces, though other polygonal faces can be stored in it without disrupting the attributes management.
On the other hand, a `TopologicalMesh` is meant to store arbitrary polygonal faces, due to its inheritance from `OpenMesh::PolyMeshT`. However, if we want to deal with triangle-only meshes, we need to inherit from `OpenMesh::TriMeshT` since the topological operations (e.g. edge split) are not processed the same way, and may lead to complete rewrite of those for the `PolyMeshT` version to behave the expected way.

The shortest way would be to only consider triangles, and never deal with other polygons. This would only imply to change the inheritance of `TopologicalMesh` to `OpenMesh::TriMeshT`. However, this also implies to always triangulate input polygonal meshes and work only on the triangles. This can be a problem when applying several Catmull-Clark subdivision passes since the faces regularity will be lost.
This is also a problem when skinning an animated character, where a nice triangulation for rendering might depend on the polygon shape after deformation.

A more viable solution would be to actually implement the difference between triangle-only meshes and polygonal meshes; and also add a triangulation conversion.
Thus, the `TriangleMesh` class would only store triangles and a `PolyMesh` class would store any polygons (with a rather subtle implementation difference).
Similarly, a `TopologicalTriMesh` class would manage the topology for triangles only, and a `TopologicalPolyMesh` class would deal with other polygons.

dlyr:
Well I think TriangleMesh could be renamed Mesh
Then the conversion back and forth TopologicalMesh should keep polygons
And the Render (or Engine) Mesh should triangulate 
One implementation issue is to keep a list of triangle, and a list of other poly, to ease the work fo RenderMesh

nmellado:
After a discussion offline, we decided to keep the `Core::TriangleMesh` class as is, just to rename it as `Core::Mesh`.
The motivation behind this choice is that keeping `Mesh` to store triangles and polys as few downsides.

On the other hand, keeping a polytopologicalmesh prevents to use some functionalities of OpenMesh when processing triangle-only meshes. To mitigate this issue, we decided to have a `TopologicalTriMesh` and a `TopologicalPolyMesh` as you mentioned. These structures will be both converted from/to a `Core::Mesh`. 
[...]

PR #364, opened 2018-09-21
Remove DCEL and HalfEdgeData

### LINKED ISSUE #120: Duplicate code for HalfEdge
Core/Mesh/DCEL/HalfEdge
Core/Mesh/

hoshiryu:

- The DCEL and HalfEdgeData topological structures are not available anymore (fix #120).
For such a structure, use the TopologicalMesh.
- This PR is related to #360 in the sense it provides effective 2-4 edge split for edges shared by 2 triangles in a PolyMesh.

* **Other information**:

We asked for an update in OpenMesh code to replace faulty code about constructing a matrix from a single value.
We must wait for this (very tiny) change to be done and then update the OpenMesh submodule.


hoshiryu:
Ok so windows compilation passes, but I'm not sure the fix is that good, especially since `RaCore.hpp` is not always included before any Eigen files (couls this lead to strange state of Eigen matrices?).

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -1,3 +1,5 @@
+#include <Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp>
```
nmellado: I don't like this kind of fix, the order in which we include files should not cause compilation error.
Can you explain the rational behind this choice ?

If this due to the fact that we use a plugin to extend Eigen, that means that this plugin is not used everywhere, but only in files that include `TopologicalMesh.hpp`.
Why not defining the plugin using a preprocessor variable defined in cmake ? So we are sure **all** files in Radium use it ?

hoshiryu: I had problems trying to fix this.
Actually, the issue comes from the addition of the constructor `Matrix<T>(T value)` which fills all the matrix entries with the given value.
When defining classes with Eigen matrices as attributes, or functions with both Eigen matrix parameter and such class parameter signatures, then the compiler cannot decide which one to call.
For example it was the case there:
```c++
Grid<Scalar, 2> grid( Vector2(3,3) );
grid.at( {0,0} ) = Scalar( 1 );
```
there are 2 possible calls: one with `Grid<Scalar, 2>::IdxVector`, e.g. `Vector2ui`, one with `Grid<Scalar, 2>::Iterator` which stores a `Vector2ui` and a `uint`, and can now be initialized through `{0,0}`.
Same with this call:
```c++
Grid<uint,2> grid( {nb_rows, nb_cols} );
```
For which both `Grid( const IdxVector& size = IdxVector::Zero(), const T& val = T() )` and `Grid( const Grid<T, D>& other )` are now valid interpretations.


If you have an idea on how to properly address this, please go on.

hoshiryu: We asked for an update on OpenMesh side to replace the falty code.

nmellado:
Waiting for a response from OpenMesh developers before going further.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
MathiasPaulin: No need to add a constructor to initialize a Matrix<T> to a constant value.
there is a method, Constant(...), already existing in Eigen that do that : 
From the Eigen doc : 
```
Similarly, the static method Constant(value) sets all coefficients to value. 
If the size of the object needs to be specified, the additional arguments 
go before the value argument, as in MatrixXd::Constant(rows, cols, value).
```
See also reference : 
https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#aed89b5cc6e3b7d9d5bd63aed245ccd6d

MathiasPaulin:
I'am not sure OpenMesh developers will try to fix an error that is not their fault but that results from a misuse of Eigen ... Instead of adding a new constructor to the Eigen::Matrix type (it is never a good idea to add new method, function, ... in an a namespace defined in an external library, question of "bon sens"). So, use the ``Matrix::Constant(T value)`` instead (or one of its overlaoded variant if the size must be specified).

nmellado:
You didn't follow @MathiasPaulin.
In some part of the code in OpenMesh, there are calls to `Matrix(Scalar fillvalue)`, while in most cases they use `Matrix(Scalar x, Scalar y, Scalar z)`.
We just asked them to place call of `Matrix(Scalar fillvalue)` by the `xyz` version.
We cannot ask them to use `Constant` as this static function is not defined for OpenMesh matrices.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
nmellado: deprecated. Closing comment.

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp
```diff
@@ -98,6 +100,130 @@ class RA_CORE_API TopologicalMesh : public OpenMesh::PolyMesh_ArrayKernelT<Topol
     inline const std::vector<OpenMesh::HPropHandleT<Vector3>>& getVector3PropsHandles() const;
     inline const std::vector<OpenMesh::HPropHandleT<Vector4>>& getVector4PropsHandles() const;
     ///@}
+    
+    /// \name Dealing with normals
```
MathiasPaulin: prefer /** ... */ for multiline documentation. Easyer to maintain and to isolate ...

MathiasPaulin:
Well, @nmellado, your right. But why there is a conflict between OpenMesh matrices and Radium (Eigen) ones ?

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +52,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
     {
         maxWeightDistance = 0;
 
         // Stores the edges to split
-        std::vector<Index> edgesToSplit;
+        std::vector<TopologicalMesh::EdgeHandle> edgesToSplit;
 
         // Compute all weights distances for all edges.
-        for ( const auto& edge : dcel.m_fulledge )
+        for ( auto e_it = topoMesh.edges_begin(); e_it != topoMesh.edges_end(); ++e_it )
```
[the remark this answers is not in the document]
hoshiryu: no, because of `edgesToSplit.push_back(...)`

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +52,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
+    dataInOut.m_CoR.resize( nVerts, Vector3::Zero() );
 
-    const uint nVerts = dataInOut.m_referenceMesh.vertices().size();
     // naive implementation : iterate over all the triangles (of the subdivided mesh)
     // for all vertices (of the original mesh).
+
+    // first precompute triangle data
+    std::map<TopologicalMesh::FaceHandle, std::tuple<Vector3, Scalar, Eigen::SparseVector<Scalar>>>
+        triangleData;
+    for ( auto f_it = topoMesh.faces_begin(); f_it != topoMesh.faces_end(); ++f_it )
```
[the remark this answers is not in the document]
hoshiryu: no, because of `triangleData[*f_it] = ...`

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +52,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
-        CORE_ASSERT( subdividedMesh.vertices()[i] == p, "Inconsistency in the meshes" );
-
         Vector3 cor( 0, 0, 0 );
         Scalar sumweight = 0;
-        const Eigen::SparseVector<Scalar> Wi = subdividedWeights.row( i );
+        const Eigen::SparseVector<Scalar> Wi = subdivW.row( mapV2I[V[i]] );
 
         // Sum the cor and weights over all triangles of the subdivided mesh.
-        for ( uint t = 0; t < subdividedMesh.m_triangles.size(); ++t )
+        for ( auto f_it = topoMesh.faces_begin(); f_it != topoMesh.faces_end(); ++f_it )
```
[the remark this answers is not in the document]
hoshiryu: what do you mean?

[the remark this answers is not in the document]
MathiasPaulin: You are doing a reduction of your collection (computing the sum of the element). Parallel reduction is a standard pattern in parallel programing and OpenMP handle this through the data sharing clause``reduction`` of a ``#pragma parallel for loop``
e.g.
```
#pragma omp parallel for private(val) reduction(+:sum)
for (row = 0; row < Rows; row++)
{
   bla bla code
   sum += val;
}
```
see parallel reduction on OpenMp tutorials for more explanation.
spec at https://www.openmp.org/wp-content/uploads/openmp-4.5.pdf

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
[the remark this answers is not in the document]
MathiasPaulin: See "thread safety" section at https://en.cppreference.com/w/cpp/container.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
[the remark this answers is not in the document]
MathiasPaulin: See "thread safety" section at https://en.cppreference.com/w/cpp/container.

nmellado:
It’s because we use Eigen matrices in openmesh.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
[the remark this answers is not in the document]
hoshiryu: same perfs with parallel for and critical section.
b.t.w., thread safety is not ensured with stl insertion methods/functions.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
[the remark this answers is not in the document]
hoshiryu: slower with parallel for and critical section.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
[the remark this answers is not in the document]
hoshiryu: I tried making the outer loop sequential and the inner one parallel with reduction, but it ended up slower (x1.5)

MathiasPaulin:
Well, openMesh people have answered. I think that we can accept this PR as is and resolve later the OpenMesh-Eigen interoperability.

nmellado:
Openmesh bug entry: https://graphics.rwth-aachen.de:9000/OpenMesh/OpenMesh/issues/59
If the CI windows build fails, we can temporary switch to this branch on *windows only* https://graphics.rwth-aachen.de:9000/OpenMesh/OpenMesh/commits/eigen

nmellado:
Regarding the OpenMesh problem, I suggest we update the OpenMesh submodule to the tip of their `eigen` branch https://graphics.rwth-aachen.de:9000/OpenMesh/OpenMesh/tree/eigen
By doing this, we solve our compilation problem without modifying OpenMesh ourselves, and we don't need to wait they find a patch for their tests.
If one day they fix it, we can again switch back to their `master` branch.

Good for all ?

MathiasPaulin:
Ok to switch to the eigen branch.

hoshiryu:
Ready for review

nmellado:
In your latest commit, what means the `file still there` ?

hoshiryu:
I first pushed with the file not removed just to check all works fine.
Then I removed the file, amend the commit and force pushed, but I forgot to rewrite the commit message

src/Core/Containers/Grid.hpp
```diff
@@ -206,8 +206,8 @@ class Grid {
     inline T& at( uint idx );
 
     /// Access an element with an iterator.
-    inline const T& at( const Iterator& it ) const;
-    inline T& at( const Iterator& it );
+//    inline const T& at( const Iterator& it ) const;
+//    inline T& at( const Iterator& it );
```
MathiasPaulin: If these methods are no longer used, remove them instead of just commenting them.

src/Core/Containers/Grid.inl
```diff
@@ -127,7 +127,7 @@ T& Grid<T, D>::at( const typename Grid<T, D>::Iterator& it ) {
     CORE_ASSERT( it.getGridSize() == m_size, "Incompatible iterator" );
     return at( it.getLinear() );
 }
-
+*/
```
MathiasPaulin: If this method is no longer used, remove it instead of just commenting.

src/Core/Mesh/TopologicalTriMesh/OpenMesh.h
```diff
@@ -1,55 +1,45 @@
...
-#define RELEASE_EIGEN_PLUGIN_PROTECTION_AGAINST_DIRECT_INCLUSION
-#define EIGEN_MATRIX_PLUGIN "Core/Mesh/TopologicalTriMesh/EigenOpenMeshPlugin.h"
-
 #include <Core/RaCore.hpp>
 
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 
 // Need to define openMesh namespace function before including openmesh, for
 // MSVC
```
MathiasPaulin: This comment could be made more explicit. Why do we need this ?

src/Core/Mesh/TopologicalTriMesh/Operations/EdgeSplit.hpp
```diff
@@ -8,10 +8,20 @@
...
+class TMOperations {
+  public:
+    /// \brief Apply a 2-4 edge split.
+    /// \param mesh The mesh on which to apply the edge split.
+    /// \param eh The handle to the edge to split.
+    /// \param f The interpolation factor to place the new point on the edge.
+    ///          Must be in [0,1].
+    /// \return True if the edge has been split, false otherwise.
+    /// \note Only applies on edges between 3 triangles, and if \p f is in [0,1].
+    /// \note Mesh attributes are linearly interpolated on the newly created halfedge.
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
     LOG( logDEBUG ) << "Precomputing CoRs";
 
     // First step : subdivide the original mesh until weights are sufficiently close enough.
-    Scalar maxWeightDistance = 0.f;
-    TriangleMesh subdividedMesh;
-    subdividedMesh.copyBaseGeometry( dataInOut.m_referenceMesh );
+    // Convert the mesh to TopologicalMesh for easy processing.
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
+            size_t hy = std::hash<Scalar>()( lvalue[1] );
+            size_t hz = std::hash<Scalar>()( lvalue[2] );
+            return ( hx ^ ( hy << 1 ) ) ^ hz;
+        }
+    };
 
-    // Store the weights as row major here because we are going to query the per-vertex weights.
-    Eigen::SparseMatrix<Scalar, Eigen::RowMajor> subdividedWeights = dataInOut.m_weights;
+    // fill map from vertex position to handle index, used to access the weight
+    // matrix from initial mesh vertices
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
+    for ( auto vit = topoMesh.vertices_begin(); vit != topoMesh.vertices_end(); ++vit, ++i )
+    {
+        mapV2I[topoMesh.point( *vit )] = vit->idx();
+    }
 
-    // Convert the mesh to DCEL for easy processing.
-    Dcel dcel;
-    convert( subdividedMesh, dcel );
+    // Squash weight matrix to fit TopologicalMesh (access through handle indices)
+    // Store the weights as row major here because we are going to query the per-vertex weights.
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
+    const auto& V = triMesh.vertices();
+    for ( int i = 0; i < V.size(); ++i )
+    {
+        subdivW.row( mapV2I[V[i]] ) = dataInOut.m_weights.row( i );
+    }
 
     // The mesh will be subdivided by repeated edge-split, so that adjacent vertices
     // weights are distant of at most `weightEpsilon`.
     // New vertices created by the edge splitting and their new weights are computed
     // and appended to the existing vertices.
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
-                edgesToSplit.push_back( edge->idx );
+                edgesToSplit.push_back( *e_it );
             }
         }
-
-        LOG( logDEBUG ) << "Max weight distance is " << maxWeightDistance;
+        LOG( logDEBUG ) << "Max weight distance is " << sqrt( maxWeightDistance );
+
+        // sort edges to split according to growing weightDistance to avoid
+        // creating edges larger than weightDistance
```
MathiasPaulin: Change this to multiline comment.

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
```diff
@@ -55,116 +51,161 @@ void computeCoR( Skinning::RefData& dataInOut, Scalar sigma, Scalar weightEpsilo
...
-                 "Weights and vertices don't match" );
-
+    const uint nVerts = V.size();
     dataInOut.m_CoR.clear();
-    dataInOut.m_CoR.resize( dataInOut.m_referenceMesh.vertices().size(), Vector3::Zero() );
+    dataInOut.m_CoR.resize( nVerts, Vector3::Zero() );
 
-    const uint nVerts = dataInOut.m_referenceMesh.vertices().size();
     // naive implementation : iterate over all the triangles (of the subdivided mesh)
     // for all vertices (of the original mesh).
```
MathiasPaulin: Change this to multiline comment.

src/Core/Mesh/TopologicalTriMesh/OpenMesh.h
hoshiryu: @nmellado could you be more specific about it?

src/Core/Animation/Skinning/RotationCenterSkinning.cpp
hoshiryu: these are 2 different one-line comments, so what to do about it?

src/Core/Mesh/TopologicalTriMesh/OpenMesh.h
nmellado: ok, what about:
```c++
// We use Eigen::Matrix as base type to represent matrices in OpenMesh.
// This is not standard, and several namespace functions are not available in OpenMesh
// for Eigen::Matrix. 
// We need these functions to be defined before including OpenMesh, otherwise some 
// compilers (e.g. MSVC) will not be able to compile the code.
```

hoshiryu: What about the inclusion of `OpenMesh.h`?

hoshiryu:
Regarding code (implementation) comments, I don't get the point of enforcing C-style multi-line comments, especially since the code-formatting script may split C++-style one-line comments in two.

src/Core/Containers/Grid.hpp
hoshiryu: restored

src/Core/Containers/Grid.inl
hoshiryu: restored

src/Core/Mesh/TopologicalTriMesh/OpenMesh.h
nmellado: I don't understand your question.

hoshiryu: File `src/Core/Mesh/TopologicalTriMesh/OpenMesh.h` should be included before any OpenMesh include right?

nmellado: I'm not sure this is required, as the failure is caused by the compilation of the TopologicalMesh.

hoshiryu:
Can someone relaunch the CI on travis please?
`apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install g++-7 cmake3 doxygen doxygen-doc doxygen-latex doxygen-gui graphviz" failed and exited with 100 during .
`

src/Core/Mesh/TopologicalTriMesh/OpenMesh.h
MathiasPaulin: OK for this comment.

nmellado:
I would like to squash-merge this PR, do you see any problem with this @hoshiryu @MathiasPaulin ?

hoshiryu:
I would not squash it all in one, but rather squash all DCEL related commits as one, then all Subdivision/TopologicalMesh related as another, and have the ones regarding OpenMesh on their own too.
It might be better for history

nmellado:
Yes it is better, but as I had no time to do it, squash was a decent solution.
Thanks for your suggestion, you have free hands to rebase

PR #366, opened 2018-09-26
Remove the duplicate table 

hoshiryu:
Also remove the RADIUM_WITH_TEXTURE which is now useless ;
Update some documentation ;
Cleanup code ;

Due to the new mesh attribute management available in Radium, the TopologicalMesh is supposed to manage unique vertices, while the `TriangleMesh` stores one vertice per different attribute (e.g. same position but different normals).
Thus, there is no need to worry about duplicates at load time, hence the suppression of the duplicate table from both `Ra::Core::Asset::GeometryData` and `FancyMeshComponent`.
This also simplifies the loading of skinning weights, since there is no need for mapping duplicate vertices indices to unique vertices indices.
However, in order to compute smooth mesh normals in the SkinningPlugin, we still need to compute a duplicate table and store it.
Also, the RADIUM_WITH_TEXTURE option has been removed since it became useless.

* **Other information**:


nmellado:
As the topology does not change, I would expect to compute the topology once, and then update only the vertices positions and normals.
If we don’t have that now, what kind of mechanism would you need in Radium ?

hoshiryu:
Yeah, but we need a TriangleMesh for rendering, thus we need to map the normals from the `TopologicalMesh` to the ones of the `TriangleMesh`, hence a map for duplicate vertices!

nmellado:
So, if I follow your reasoning, we need an object or method which synchronize attributes of ˋTriangleMesh` from a ˋTopologicalMesh`, while taking care of dispatching attributes from half-edges to the ˋTirangleMesh` vertices.
I agree such structure might need to store a map like that, but I don’t see why a duplicate_map would be stored anywhere else.

dlyr:
Actually the index going from halfedge in `TopologicalMesh` to vertex in `TriangleMesh` si computed here https://github.com/STORM-IRIT/Radium-Engine/blob/439eeb3408f46deaf877200e91997c1de2f73ad6/src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp#L251 
One easy solution is to store this index as a prop of the halfedge during conversion (and each time topology changes), then update will be straightforward.

A better solution should be to identify when attrib are really defined by halfedge and not by vertex, but it is another story.

nmellado:
You need the TopologicalMesh because you need to move on the mesh, to collect topological neighbors and aggregate attributes defined on faces (normal vectors) surrounding a vertex.
What you do with the duplicate_table is storing a part of the information stored in the TopologicalMesh.

hoshiryu:
After discussing with @dlyr , I better understand your point.
I agree with the solution he proposed, i.e. storing the indices on the halfedges (we would need a dedicated access to this property).
However, I don't get how to keep them up-to-date when applying topological changes, nor how to use them when converting to `TriangleMesh`.
On the same basis, should all mesh treatment we have on `TriangleMesh` thus be translated to work on `TopologicalMesh` only ?

Nevertheless, regarding the current implementation of skinning methods, which lack normal transformation, we still need to recompute the normals at each frame, and I'm quite sure that using the proposed method will degrade performances, since we would have to upload the `TriangleMesh` vertices to the `TopologicalMesh`, compute the normals (no overhead), and then download the normals back.
I thus propose to move the duplicate_table computation/use for normals within the Skinning Plugin for now (to come in this PR) and add an issue about normal transformation for skinning.

nmellado:
@MathiasPaulin can you check these change do no break anything on your side ? (it shouldn't)

MathiasPaulin:
Everything is fine for me. My external plugins just required some minor modifications (RADIUM_WITH_TEXTURE option being removed) but all works as expected.

PR #368, opened 2018-10-01
Remove dispatch of component registration methods

hoshiryu:

Fix the default behavior of `CouplingSystem` regarding the `registerComponent`, `unregisterComponent` and `unregisterAllComponents` methods of `Ra::Engine::System`.
Also give the file directory to the AbstractTimedSystem for caching/restoring frame data, so that the  CouplingSystem and its subsytems use the same by default.

By default, the `CouplingSystem` dispatches the `registerComponent`, `unregisterComponent` and `unregisterAllComponents` to the coupled systems, which makes them register its own managed `Components`.
This is undesirable, since the `Systems` must only register `Components` they manage (create, run tasks over, link to some plugin interface).

By default, there is no dispatch for these methods.
In the case one wants the coupled systems to also register `Components` related to the `CouplingSystem`, then one must declare/implement it explicitely.

* **Other information**:

### DIFF
--- a/Plugins/Animation/src/AnimationPlugin.cpp
+++ b/Plugins/Animation/src/AnimationPlugin.cpp
@@ -28,7 +28,6 @@ void AnimationPluginC::registerPlugin( const Ra::PluginContext& context ) {
         path = QString( context.m_exportDir.c_str() );
     }
     m_system = new AnimationSystem;
-    m_system->setDataDir( path.toStdString() );
     context.m_engine->registerSystem( "AnimationSystem", m_system );
     context.m_engine->getSignalManager()->m_frameEndCallbacks.push_back(
         std::bind( &AnimationPluginC::updateAnimTime, this ) );
@@ -142,11 +141,11 @@ void AnimationPluginC::updateAnimTime() {
 }
 
 void AnimationPluginC::cacheFrame() {
-    m_system->cacheFrame();
+    m_system->cacheFrame( m_dataDir );
 }
 
 void AnimationPluginC::restoreFrame( int frame ) {
-    if ( m_system->restoreFrame( frame ) )
+    if ( m_system->restoreFrame( m_dataDir, frame ) )
     {
         m_widget->frameLoaded( frame );
     }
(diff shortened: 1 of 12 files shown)


nmellado:
Ok, but why ?
Can you explain why dispatching is a problem ?

hoshiryu:
Because then each coupled system will register the component as one of its own, which is not correct.
The `System`s must only registers `Component`s they manage (create, run tasks over, link to some plugin interface).
In the case one wants the coupled systems also register Components related to the `CouplingSystem`, then one must declare it explicitely, but it's not the right default behavior.

MathiasPaulin:
Ok, but why ? I do not find your answer to this question very clear. Why it is not correct, why the Coupled systems can't register their components ?
Please, when you submit a PR, explain clearly, in the required fields recalled below, all these things instead of waiting we ask you more precision. Take the habits to be as explicit as possible. 

* What kind of change does this PR introduce? (Bug fix, feature, docs update, ...)
* What is the current behavior? (You can also link to an open issue here)
* What is the new behavior (if this is a feature change)?
* Does this PR introduce a breaking change? (What changes might users need to make in their application due to this PR?)
* Other information:

hoshiryu:
As discussed, I tried to make the component registration methods non virtual in `Ra::Engine::System`, however I couldn't due to possible additional treatment done by the inheriting System.
For example, LightManager and CameraManager not only register the component, but also store it into a dedicated storage.

nmellado:
Methods `registerComponent`, `unregisterComponent` and `unregisterAllComponents` must be `virtual` and `protected`.

nmellado:
This PR introduces several changes that are ***not*** related to the bug fix mentioned in the title + summary:
 - AnimationPlugin directory management,
 - Use of `this->foo()` instead of direct method call `foo()` in `TriangleMesh`,
 - Changes of logic in `CameraManager` and `LightManager`,
 - Renaming of `dummy` to `dummy4` in `Engine::Mesh` + change of logic in `Engine::Mesh::getData()`

Also, none of these change is motivated.
I'm closing this PR, as its goal is not clear.

PR #370, opened 2018-10-02
Fix bugs, compilation errors and CouplingSystem design

hoshiryu:

- Fix the default behavior of `CouplingSystem` regarding the `registerComponent`, `unregisterComponent` and `unregisterAllComponents` methods of `Ra::Engine::System`.
- Fix method  `unregisterAllComponents` for Light and Camera managers to ensure only the data related to the given entity are removed.
- methods `*register*Component` in `System` are made protected, and `Component` is friended so that it can call `unregisterComponent` on its `System`.
- Fix compilation issues in Debug mode and crash due to strange management of Vec3 attributes in `Ra::Engine::Mesh`.
- Fix compilation issue due to wrong call in lambda function.
- Give the file directory to the AbstractTimedSystem for caching/restoring frame data, so that the  CouplingSystem and its subsytems use the same by default.
- Fix memory leak in AttribManager.
- Also fix `RA_DISPLAY_LINE` macro.

- By default, the `CouplingSystem` dispatches the `registerComponent`, `unregisterComponent` and `unregisterAllComponents` to the coupled systems, which makes them register its own managed `Components`.
This is undesirable, since the `Systems` must only register `Components` they manage (create, run tasks over, link to some plugin interface).
- The implementation of the `unregisterAllComponents` method in Light and Camera managers removes all the components from the storage, not only the ones from the given Entity.
- Vec3 attributes management crashes on access when in Debug mode if there is no such data, while for Vec4 it doesn't.
- Vec3 and Vec4 attributes handles are not updated when changing the mesh, leading to a crash in some cases when adding attributes.
- Direct call to `TriangleMesh::append_attrib` inside the lambda function generates a compilation error.
- animation data file directory is managed by the Animation component, preventing subsystems to know where to save/load the files.
- No specific destructor is provided for the `AttribManager`, hence no deletion of the new-ed attributes data on default destructor.
- `RA_DISPLAY_LINE_ONCE` macro calls a removed method.

- By default, there is no dispatch for these methods.
In the case one wants the coupled systems to also register `Components` related to the `CouplingSystem`, then one must declare/implement it explicitely.
- The implementation of the `unregisterAllComponents` method in Light and Camera managers removes only the ones from the given Entity.
- Vec3 attributes management is similar to the one for Vec4 attributes and does not crash anymore.
- Vec3 and Vec4 attributes handles are fetched when changing the mesh.
- Calling `TriangleMesh::append_attrib` inside the lambda function, prepended by the lambda context variable `this`.
- AbstractTimedSystem must be given the directory for data file save/load. Management of this directory for the Animation data is performed by the AnimationPlugin.
- The `AttribManager` destructor clears all attribs data.
- `RA_DISPLAY_LINE` macro calls the right method.

* **Other information**:
- The `registerAllComponent` method is never used, while there should be at least one binding with Entity deletion.


hoshiryu:
Could someone relaunch the CI please? it stopped because it was unable to fetch packages!

nmellado:
Is this PR still WIP, or can we review it now ?

hoshiryu:
Ready for review

hoshiryu:
Ready for review

PR #371, opened 2018-10-07
Topological cleanup and test

dlyr:

Cleanup and fixes TopologicalMesh conversion, and add a bit of verbosity to coretests.

n/a
n/a

n/a

* **Other information**:


nmellado:
@hoshiryu can you double check please ?

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
```diff
@@ -208,10 +206,6 @@ TriangleMesh TopologicalMesh::toTriangleMesh() {
     for ( auto oh : m_vec4Pph )
         addAttribPairToCore( out, this, oh, vprop_vec4 );
 
-    request_face_normals();
-    request_vertex_normals();
-    update_vertex_normals();
-
     // iterator over all faces
```
hoshiryu: I would suggest to also remove these from the Subdividers.
However, there is some issue with the normals there.

hoshiryu:
Can you add some documentation about the fact that we cannot directly call the inherited method `normal(VertexHandle)` (calling `PolyMeshT::normal(VertexHandle)` is possible), though we can directly call `set_normal(VertexHandle)`.

dlyr:
Actually I think you can use `normal(VertexHandle)` but the conversion to TriangleMesh uses `normal(VertexHandle, FaceHandle)`.
One solution might be to redefine those function to affect all halfedges surrouding a vertex.

hoshiryu:
The compiler complains when calling `normal(VertexHandle)`

dlyr:
Yes i find out why (overloading masking). Will be corrected soon. 

++ david

hoshiryu:
Could someone relaunch the CI please? it stopped due to failed apt-get install, thanks!

dlyr:
Seems ok for review.

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
```diff
@@ -27,7 +27,7 @@ void addAttribPairToTopo( const TriangleMesh& triMesh, TopologicalMesh* topoMesh
 }
 
 template <typename T>
-void addAttribPairToCore( TriangleMesh& triMesh, TopologicalMesh* topoMesh,
+void addAttribPairToCore( TriangleMesh& triMesh, const TopologicalMesh* topoMesh,
                           OpenMesh::HPropHandleT<T> oh, std::vector<PropPair<T>>& vprop ) {
     AttribHandle<T> h{triMesh.addAttrib<T>( topoMesh->property( oh ).name() )};
```
hoshiryu: why not keeping code consistent with `Attr h = ... ` ?

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
```diff
@@ -48,7 +48,7 @@ using HandleAndValueVector = std::vector<std::pair<AttribHandle<T>, T>,
                                          Eigen::aligned_allocator<std::pair<AttribHandle<T>, T>>>;
 
 template <typename T>
-void copyAttribToCoreVertex( HandleAndValueVector<T>& data, TopologicalMesh* topoMesh,
+void copyAttribToCoreVertex( HandleAndValueVector<T>& data, const TopologicalMesh* topoMesh,
                              std::vector<PropPair<T>>& vprop,
```
hoshiryu: why not `const ...& vprop` ?

hoshiryu: same at line 61, why not `const ...& data` ?

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp
```diff
@@ -58,19 +58,36 @@ class RA_CORE_API TopologicalMesh : public OpenMesh::PolyMesh_ArrayKernelT<Topol
...
+    using base::set_normal;
+    ///@}
 
-    /// return the normal of the vertex vh, when considering its membership to
-    /// the face fh
-    inline Normal& normal( VertexHandle vh, FaceHandle fh );
+    /// Set the normal n to all the halfedges that points to vh.
+    /// If you work with vertex normals, please call this function on all vertex
+    /// handle before convertion with toTriangleMesh
+    void propagate_normal_to_surronding_he( TopologicalMesh::VertexHandle vh );
```
hoshiryu: typo: `surrounding`, could also be `incoming`

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
```diff
@@ -226,7 +220,7 @@ TriangleMesh TopologicalMesh::toTriangleMesh() {
         int i = 0;
 
         // iterator over vertex (thru halfedge to get access to halfedge normals)
-        for ( TopologicalMesh::FaceHalfedgeIter fh_it = fh_iter( *f_it ); fh_it.is_valid();
+        for ( TopologicalMesh::ConstFaceHalfedgeIter fh_it = cfh_iter( *f_it ); fh_it.is_valid();
```
hoshiryu: Nice to improve semantic (though it's the same)! Should also be applied within the Subdividers and the TopologicalMeshTest!

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
hoshiryu: also, why not using `const Handle& h` in argument? (same for functions below)

dlyr: Well, I'm not sure to catch the first comment.
For the second, as a PropHandle is something like an int, and all OpenMesh method take arg by value, I don't think we need to use const ref.

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
dlyr: Actually, I'm not aware of what copyAttrib as to do, I just propagate constness from the toTriangleMesh operator.
So feel free to add const if appropriate.

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.hpp
dlyr: right

src/Core/Mesh/TopologicalTriMesh/TopologicalMesh.cpp
dlyr: Actually I had to change to Const because of the constness of toTriangleMesh member function. 
Feel free to improve semantic elsewhere if appropriate

dlyr: Resolved here, but could you please do the same for subdividers (I prefer not to mess up there ;))

src/Core/Algorithm/Subdivision/SubdividerUtils.hpp
```diff
@@ -4,6 +4,105 @@
 namespace Ra {
 namespace Core {
 
+// Deal with normals
+
+/// Copy normal \p input property of mesh.
```
nmellado: What is a `normal input property` ? The normal vector ?

src/Core/Algorithm/Subdivision/SubdividerUtils.hpp
```diff
@@ -4,6 +4,105 @@
...
+#pragma omp parallel for
+    for ( int i = 0; i < mesh.n_halfedges(); ++i )
+    {
+        auto heh = mesh.halfedge_handle( i );
+        mesh.property( copy, heh ) = mesh.property( input, heh );
+    }
+}
+
+/// Create a new property for \p input property of \p mesh on faces.
+inline void addPropCopyF( const OpenMesh::HPropHandleT<TopologicalMesh::Normal>& input,
```
nmellado: The method name is not really reflecting what is done in it. 
Something like `createProp` would be more meaningful.
Is the current name motivated by any naming convention ?

src/Core/Algorithm/Subdivision/SubdividerUtils.hpp
```diff
@@ -25,7 +124,7 @@ void addPropsCopy( const std::vector<OpenMesh::HPropHandleT<T>>& input, Topologi
     }
 }
 
-/// Add a copy of \p input properties of \p mesh on faces.
+/// Create a new property for each \p input properties of \p mesh on faces.
 template <typename T>
 void addPropsCopyF( const std::vector<OpenMesh::HPropHandleT<T>>& input, TopologicalMesh& mesh,
```
nmellado: Same as `addPropCopyF`.

hoshiryu:
Ready!

PR #377, opened 2018-10-12
Update Light asset : add management of area light sources

MathiasPaulin:
- [x ] The commit message follows our guidelines
- [ x] Docs have been added / updated (for bug fixes / features)

Allow the definition of area light sources in the asset that are compatible with punctual ones so that renderer that support only point light sources might use a crude approximation of area lights.

No support for area light.

Support for a statistical representation of arealight.

* **Other information**:
Update the doc on ``File::IO::LightData``
Tests are not in the Radium distribution but in a separate private plugin.

### DIFF
--- a/src/Core/File/LightData.hpp
+++ b/src/Core/File/LightData.hpp
@@ -47,7 +47,7 @@ class RA_CORE_API LightData : public AssetData {
     /// CONSTRUCTOR
     LightData( const std::string& name = "", const LightType& type = UNKNOWN );
 
-    LightData( const LightData& data ) = default;
+    LightData( const LightData& data );
 
     /// DESTRUCTOR
     ~LightData();
@@ -82,7 +82,7 @@ class RA_CORE_API LightData : public AssetData {
     inline void setLight( Core::Color color, Core::Vector3 position, Core::Vector3 direction,
                           Scalar inangle, Scalar outAngle, LightAttenuation attenuation );
     /// construct an area light
-    inline void setLight( Core::Color color, LightAttenuation attenuation );
+    inline void setLight( Core::Color color, Core::Vector3 cog, Core::Matrix3 spatialCov, Core::Matrix3 normalCov, LightAttenuation attenuation );
 
     /// QUERY
     inline bool isPointLight() const;
@@ -120,7 +120,11 @@ class RA_CORE_API LightData : public AssetData {
             LightAttenuation attenuation;
         } m_spotlight;
         struct {
-            // TODO (Mathias) : find how to represent arealight in a generic way
+            // TODO (Mathias) : this representation is usefull but might be improved
+            Core::Vector3 position;
+            // TODO : gives the eigenvalues/eigenvectors for spatial and normal covariance matrices.
+            Core::Matrix3 spatialCovariance;
+            Core::Matrix3 normalCovariance;
             LightAttenuation attenuation;
         } m_arealight;
     };
--- a/src/Core/File/LightData.hpp
+++ b/src/Core/File/LightData.hpp
@@ -152,7 +152,7 @@ class RA_CORE_API LightData : public AssetData {
     * class could not ensure such an invariant when this method is called.
     * @param type the type to set.
     */
-    inline void setType( const LightType& type );
+    [[deprecated( "Do not change the type of a LightData, use setLight that ensures data coherency." )]] inline void setType( const LightType& type );
 
     /**
      * Returns true if the light is a PointLight
--- a/src/Core/File/LightData.hpp
+++ b/src/Core/File/LightData.hpp
@@ -20,12 +20,21 @@ class Light;
 
 namespace Asset {
 
+  /**
+   * Asset defining how lights must be passed to the Radium systems from a loader.
+   *
+   * Supported light type are :
+   *    Point light : a local, punctual, isotropic spherical light.
+   *    Spot light : a local, punctual, isotropic conical light.
+   *    Directional light : an infinite, isotropic light.
+   *    Area light : a local isotropic area light.
+   */
 class RA_CORE_API LightData : public AssetData {
 
   public:
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW
 
-    /// ENUM
+    /// Supported light type
     enum LightType {
         UNKNOWN = 1 << 0,
         POINT_LIGHT = 1 << 1,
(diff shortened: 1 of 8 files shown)


src/Core/File/LightData.hpp
```diff
@@ -43,54 +56,121 @@ class RA_CORE_API LightData : public AssetData {
...
-    /// TYPE
-    inline LightType getType() const;
-    inline void setType( const LightType& type );
 
-    /// FRAME
+    /**
+     * Acces to the local frame of the light.
+     * @return the local frame
+     */
     inline Core::Matrix4 getFrame() const;
```
nmellado: return reference here ?

src/Core/File/LightData.hpp
```diff
@@ -43,54 +56,121 @@ class RA_CORE_API LightData : public AssetData {
...
-    LightData( const std::string& name = "", const LightType& type = UNKNOWN );
-
-    LightData( const LightData& data ) = default;
-
-    /// DESTRUCTOR
+    // TODO : allow to define other attenuation function such \max\left(0,1-\frac{\left|x\right|^2}{r^2}\right)^2
+
+    /**
+     * Light constructor.
+     * Must be calle explicitely with a name for the light and its type.
```
nmellado: typo: must be calle*d*

src/Core/File/LightData.hpp
```diff
@@ -43,54 +56,121 @@ class RA_CORE_API LightData : public AssetData {
...
-    inline void setLight( std::shared_ptr<Ra::Engine::Light> light );
+/**
+ * \defgroup HelperSetters Helper functions to set the various light parameters.
+ * @{
+ */
+    /**
+     * Construct a directional light.
+     * A directional light is only defined by its color and its lighting direction.
+     * No attenuation on directional.
+     */
```
nmellado: It might be useful to say here that this function changes the type of light.
By the way, why these methods are not static initialization functions ? They don't take into account the state of the instance, but just set a new light from a set of parameters.

src/Core/File/LightData.hpp
```diff
@@ -43,54 +56,121 @@ class RA_CORE_API LightData : public AssetData {
...
 
-    /// QUERY
+    /**
+    * Modify the type of a light asset after its creation.
+    * \note Beware of the risk of data inconsistancy when using the operator. As a LightData object is a union like
+    * object, caller of this method must explicitely ensure a consistancy invariant over its object as the LightData
+    * class could not ensure such an invariant when this method is called.
+    * @param type the type to set.
+    */
+    inline void setType( const LightType& type );
```
nmellado: I understand this method can cause inconsistency, as it changes the accessing type of the union storing the light parameters.
So my question is: do we need it ? If there is no express need for this method I would prefer we remove it.

src/Core/File/LightData.hpp
MathiasPaulin: no need for reference nor const reference. Copy elision and rvalue optimisation is now well supported by all the C++ compiler. So, there will be always at most one copy with both the ref or object return version of such a function. And, in this case, it is clear that the returned Matrix is a rvalue.

src/Core/File/LightData.hpp
MathiasPaulin: OK, I though the "Construct a 'type' light, in this method and following, was sufficiently clear. I will precise that these methods change the object so that it become an object of the type inferred from the overloaded method called.
And these method are not static due to inheritance (in coding style) of quite discutable conception choices and programming approach in the Old Radium. Yes, we need to refactor the LightData construction process but this will imply modifying much more files than this PR need to modify to just add the management of area light sources.
We can do this refactoring now (this will imply to modify also geometry, material and animation asset management that suffer from quite similar problems and all the systems that handle asset loading. Quite a huge work i think) or integrate it on the Radium-v2 roadmap as this future roadmap will contain a lot of such refactoring to simplify the writing of loaders/writers and, more generally, the input/ouput general process.

src/Core/File/LightData.hpp
MathiasPaulin: We don't need these function, we can mark them as deprecated and removed progressively their uses from Asset importers and Light managers. For now, in the Radium distribution, they are used by ``AssimpLoader`` ecosystem and ``LightManager``. 
We can either uses the following (i.e ``isTypeLight``) but this will be less convenient for future extension. 
The better (that will require much more refactoring and code rewriting) is to rely on polymorphic factory (the same way the materials are managed through converter functions that ease the extension of supported material type) that will allow to extend the set of supported light sources in the engine from an external plugin, extension that is not feasible for now.

src/Core/File/LightData.hpp
nmellado: Ok, it's just that I don't like to rely on copy elision, but it is a matter of style I guess.
In debug mode the matrix will be copied anyway.

src/Core/File/LightData.hpp
nmellado: Please mark `setType` with the `deprecated` attribute, and will properly fix this in v2.

src/Core/File/LightData.hpp
MathiasPaulin: Why ??? rvalues are handled independently of the compilation mode no ? (well, not sure with visualC++ compiler but ...)
And we can find lot of pertinent reasons to return an object and not a ref. More, in this case, the ref must be constant as the method is marked as const. and so the copy will be done.
for recommandations and explanation (by Stroustrup and Sutter - and even the whole C++ comunity) see https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f15-prefer-simple-and-conventional-ways-of-passing-information.

nmellado: According to https://en.cppreference.com/w/cpp/language/copy_elision#Notes :
>  Because some compilers do not perform copy elision in every situation where it is allowed (e.g., in debug mode), programs that rely on the side-effects of copy/move constructors and destructors are not portable.

nmellado: So for me it should be 
```c++
inline const Core::Matrix4& getFrame() const;
```

MathiasPaulin: Well, It is not because some compilers do not generate portable programs (depending on the debug/or release mode) that we must write some over-engineered code. 
If we have to program according to the way a (unknown) compiler might understand our code instead of the way the langage is constructed to make the programmer concentrate on its concepts rather than on the code generation (that, if i recall well, is the job of the compiler), we are not "sorti de l'auberge :p  )
That being said, **i give up. You won** (even if i disagree). 
Definitively I personally prefer B. Stroustrup and H. Sutter over VisualC++ :p

PR #379, opened 2018-10-15
Fix dirty bits in Mesh ;

hoshiryu:

Fix `Engine::Mesh::m_dataDirty` when loading `TriangleMesh`

When loading a `Core::TriangleMesh` into an `Engine::Mesh`, the dirty bits for the attributes are not set, thus the attributes are not loaded on the GPU.

The dirty bits are now set for all data.

* **Other information**:


hoshiryu:
What about `Mesh::updateMeshGeometry()` and `Mesh::loadGeometry( vertices, indices)`?

nmellado:
It is true that `updateMeshGeometry` might lead to inconsistent state. On the other hand it doesn't seems to be used anywhere. Shall we remove it ?
`loadGeometry` sets the dirty flags, what's wrong with it ?
The non-const version of `getData` is maybe too intrusive, + I cannot find any use in the main project. Shall we also remove it ?

hoshiryu:
I'm ok to remove these. @dlyr @MathiasPaulin is it ok for you too?

hoshiryu:
Regarding `loadGeometry( vertices, indice )`, what shall we do about the attributes?

nmellado:
In `loadGeometry`, there is basically two options: either the vertices are updated (ie. there is not vertex addition or removal), or the vertex list is modified.
In the second case, it is clear that attributes must be invalidated. In the first case however, we don’t know, it can be an update of the geometry, or a completely new geometry with the same number of vertices.

Considering this, we first need to confirm we really need that method. If yes, in which case ? With all these information we will decide what to do.
 Personally I don’t like this kind of method as it make break the data integrity. But I might be wrong, so I’m open to discussions.

hoshiryu:
I removed `Mesh::updateGeometry` and the non-const versions of `Mesh::getData`.
Regarding`Mesh::loadGeometry( vertices, indices )` I kept it since it is used for non-`RM_TRIANGLES` meshes (e.g. lines, points, strips, ...).

Issue #202, opened 2017-07-07
Mesh Management


hoshiryu:
There is some common agreement on the need to erase the FancyMesh plugin and replace it with a proper and nicer mesh management.
Also, there is some considerations about where stuff like texture coordinates, tangent, bitangent and others should be stored.

## Current State

For now, when loading a mesh, the loader gives a FileData to the Engine that calls for the system to handle it.
In particular, the system of the FancyMesh plugin use the GeometryData to build one FancyMeshComponent per mesh for visualization purposes:
* the TriangleMesh is embedded inside an Engine::Mesh, managing OpenGL memory consistency;
* this Engine::Mesh is stored in a RenderObject through a shared_ptr, along with the material and shader config;
* the FancyMeshComponent give read/write access to the TriangleMesh via the ComponentMessenger.
For the RenderObject to take into account modifications made on the mesh by other components (such as the SkinningComponent), the "dirty" flag is set, for the corresponding data, by the FancyMeshComponent when the write-access getter is called via the ComponentMessenger.
Also to be noted, there is a never-used method in RenderObject to create one from a GeometricData.

Regarding the geometry data, the only ones stored in the TriangleMesh are the vertices, the per-vertice normals and the triangles (vertices indices). Texture coordinates, tangent and bitangent vectors, and other vertex-wise attributes (not talking about skinning weights here) are stored by the Engine::Mesh.
Also to be noted, for now there can be only one attribute of a kind.

## Issues with the current state

For now, the one owning the TriangleMesh is the FancyMeshComponent, but through the Engine::Mesh used by its RenderObject.
Therefore, the FancyMesh plugin is for now mandatory to have a mesh to work on, which breaks the concept of plugin and which may lead to undesired code structure and complexity in some cases.

Regarding the geometric data, having them split into the TriangleMesh and the Engine::Mesh causes an over-complex structure of the code for geometric treatment algorithms, which would depend on the Engine part of Radium, which is non-sense.

For example, in order to develop a mesh processing application that may be run without any visualization requirement (e.g. command-line tool), one may need to build its own geometric component owning the TriangleMesh.
Moreover, the TriangleMesh class is not self-sufficient for some treatment targeting or dependent on additional vertex-wise (or even triangle-wise) attributes, and a separate class have to be built.

## Possible Solutions

[...]

PR #380, opened 2018-10-15
Replace FancyMeshPlugin by GeometrySystem and GeometryComponent.

nmellado:

Remove FancyMesh plugin. 
Partially address #202.

GeometryData are converted to `Ra::Engine::Mesh` by a plugin FancyMesh.
As plugin loading order is unspecified, accessing the Mesh data from another plugin when loading assets is not robust.

New classes `GeometrySystem` and `GeometryComponent` have been added.
The `GeometrySystem` is now created in `BaseApplication` Before loading plugins.

This PR might break other plugins explicitly calling for FancyMesh. However such use is not recommended and thus might not prevent to merge this PR.

* **Other information**:
The names `GeometrySystem` and `GeometryComponent` can be discussed.


src/GuiBase/BaseApplication.hpp
```diff
@@ -140,6 +141,9 @@ class RA_GUIBASE_API BaseApplication : public QApplication {
     /// Timer to wake us up at every frame start.
     QTimer* m_frameTimer;
 
+    /// GeometrySystem taking care of converting assets to geometry
+    Ra::Engine::GeometrySystem* m_geometrySystem;
```
hoshiryu: Is there a need to keep it as an attribute?

hoshiryu:
Could you also update `Docs/model.md` ?

hoshiryu:
Regarding `explicitly calling for FancyMesh`, the only way that should have been considered is through the `ComponentMessenger`.

src/Engine/Component/GeometryComponent.cpp
```diff
@@ -105,170 +101,170 @@ void FancyMeshComponent::handleMeshLoading( const Ra::Asset::GeometryData* data
...
+        displayMesh->addData( Mesh::VERTEX_TEXCOORD, data->getTexCoords() );
     }
 
     if ( data->hasColors() )
     {
-        displayMesh->addData( Ra::Engine::Mesh::VERTEX_COLOR, data->getColors() );
+        displayMesh->addData( Mesh::VERTEX_COLOR, data->getColors() );
     }
 
     // FIXME(Charly): Should not weights be part of the geometry ?
```
MathiasPaulin: Why Charly is still there ??? This FIXME is very old ... (from the begining of the Radium dev).

src/Engine/Renderer/Camera/Camera.cpp
```diff
@@ -43,7 +43,7 @@ void Camera::initialize() {
     m->addData( Mesh::VERTEX_COLOR, c );
 
     // Create the RO
-    m_RO = RenderObject::createRenderObject( m_name + "_RO", this, RenderObjectType::Fancy, m );
+    m_RO = RenderObject::createRenderObject( m_name + "_RO", this, RenderObjectType::Geometry, m );
```
MathiasPaulin: Set the ype of the Ro as ``RenderObjectType::UI`` or ``RenderObjectType::Debug``
Camera geometrical representation (which, even if usefull,  could be improved) must not be part of the scene but considered as ui  or debug object. This will allow to easily toggle the display of the camera in the scene throug main-app but also to facilitate some scene wide computation such as bounding boxes computation or colision detection, made only versus Geometry objects.

nmellado:
Ready for review.

hoshiryu:
There is still some references to FancyMesh in `Docs/model.md`

Plugins/Skinning/src/Display/SkinningDisplayComponent.hpp
```diff
@@ -139,7 +139,7 @@ class SKIN_PLUGIN_API SkinningDisplayComponent : public Ra::Engine::Component {
             std::string name = m_name + "_Partition";
 
             Ra::Engine::RenderObject* renderObject =
-                new Ra::Engine::RenderObject( name, this, Ra::Engine::RenderObjectType::Fancy );
+                new Ra::Engine::RenderObject( name, this, Ra::Engine::RenderObjectType::Geometry );
```
hoshiryu: There is an issue with the rendering of cameras due to the rendering architecture.
Please Restore to  `Geometry`.

src/Engine/RadiumEngine.hpp
```diff
@@ -99,7 +106,18 @@ class RA_ENGINE_API RadiumEngine {
     const std::vector<std::shared_ptr<Asset::FileLoaderInterface>>& getFileLoaders() const;
 
   private:
-    std::map<std::string, std::shared_ptr<System>> m_systems;
+    using priority = int;
+    using SystemKey = std::pair<priority, std::string>;
+    using SystemContainer = std::map<SystemKey, std::shared_ptr<System>, std::greater<SystemKey>>;
```
hoshiryu: Then when going through this map, `System` with the same priority will be processed by reversed alphabetical order right?

src/GuiBase/BaseApplication.cpp
```diff
@@ -203,6 +204,9 @@ BaseApplication::BaseApplication( int argc, char** argv, const WindowFactory& fa
     createConnections();
     processEvents();
 
+    // Register the GeometrySystem converting loaded assets to meshes
+    m_engine->registerSystem( "GeometrySystem", new Ra::Engine::GeometrySystem, 1000 );
```
hoshiryu: use `std::limits<int>::max()` instead

src/Engine/RadiumEngine.hpp
nmellado: Yes, as it was before introducing the priority.
But for me this is implementation dependent and I don't think it is a good idea to explain it (we just say random), otherwise users may be tempted to alter the system name to change the order, instead of using the priority.

src/GuiBase/BaseApplication.cpp
nmellado: No, there might be cases where we want to be able to set a higher priority.

src/Engine/RadiumEngine.hpp
hoshiryu: Ok (though it was the other way around due to standard string comparison being lower, not greater ;) )

nmellado: You see, we don't want to explain on this ;)

src/GuiBase/BaseApplication.cpp
hoshiryu: really? like changing loading parameters?

nmellado: For instance. The thing is that we don't know, so why closing doors for no reason ?
Someone doing that is expected to know what he/she is doing.
Also it is in the application side, so no real impact on the library. We might need to update this if we add a management UI to load/unload plugins and adjust priorities in the main app (one day maybe).

Issue #378, opened 2018-10-14
Application crashes when clearing a scene with cameras.


MathiasPaulin:
When a loaded scene contains cameras, the main app - through CameraManip plugin - allow to use loaded camera instead of default one.
Even if such loaded camera are not very well handled by the manipulator ( these is another issue that will be raised soon), there is a bigger issue that requires to be fixed quickly.
After using a loaded camera, if the scene is cleared, the application crashes as the deleted camera is still the one in use at next draw request.
This could be fixed by switching first to the default (system) camera  before clearing the scene and then destroy all its entities.

PR #381, opened 2018-10-17
Fix issue #378 - main-app crashes when clearing a scene from which the active camera was selected.

MathiasPaulin:

Fix issue #378 

When selecting a camera from a loaded scene, then clearing the scene, the Camera interface manage a dangling pointer toward a deleted ``Engine::Camera``. This make the main-app crash at the next draw event.

Before clearing the scene, the CameraInterface switch back to the system default camera that has the lifetime of the application.

No

* **Other information**:
This fix a a quick workaround and must not be considered for other use except in the actual main-app.

### DIFF
--- a/src/GuiBase/Viewer/CameraInterface.cpp
+++ b/src/GuiBase/Viewer/CameraInterface.cpp
@@ -36,6 +36,23 @@ Gui::CameraInterface::CameraInterface( uint width, uint height ) :
     setCameraZFar( 1000.0 );
 }
 
+void Gui::CameraInterface::resetToDefaultCamera() {
+    // get parameters from the current camera
+    // Thisis awfull and requires that the current camera is still alive ...
+    Scalar w = m_camera->getWidth();
+    Scalar h = m_camera->getHeight();
+    auto it = std::find_if(
+        Engine::SystemEntity::getInstance()->getComponents().cbegin(),
+        Engine::SystemEntity::getInstance()->getComponents().cend(),
+        []( const auto& c ) { return c->getName().compare( "CAMERA_DEFAULT" ) == 0; } );
+    if ( it != Engine::SystemEntity::getInstance()->getComponents().cend() )
+    {
+        m_camera = static_cast<Engine::Camera*>( ( *it ).get() );
+        m_camera->resize(w,h);
+        m_camera->show( false );
+    }
+}
+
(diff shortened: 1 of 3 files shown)


src/GuiBase/Viewer/CameraInterface.cpp
```diff
@@ -36,6 +36,23 @@ Gui::CameraInterface::CameraInterface( uint width, uint height ) :
...
+    auto it = std::find_if(
+        Engine::SystemEntity::getInstance()->getComponents().cbegin(),
+        Engine::SystemEntity::getInstance()->getComponents().cend(),
+        []( const auto& c ) { return c->getName().compare( "CAMERA_DEFAULT" ) == 0; } );
+    if ( it != Engine::SystemEntity::getInstance()->getComponents().cend() )
+    {
+        m_camera = static_cast<Engine::Camera*>( ( *it ).get() );
+        m_camera->resize(w,h);
+        m_camera->show( false );
+    }
```
nmellado: Can we add a bit of logging here ? 
```suggestion
    }
    else 
    {
        LOG(logWARNING) << "A living camera is required. The application might now behave unexpectedly." ;
    }
```

MathiasPaulin: Why not, I applied your suggestion ...
But Remind that the else will never be reached except if the global invariant saying that a default camera is always created by the viewer when it is constructed (see the constructor above)

nmellado: Sure.
We should never reach that point, except, of course.... if we do ;)
I just wanted to avoid silent errors.

src/GuiBase/Viewer/CameraInterface.hpp
```diff
@@ -71,6 +71,15 @@ class CameraInterface : public QObject {
     /// \note CameraInterface doesn't have ownership.
     virtual void setCamera( Engine::Camera* camera ) = 0;
 
+    /**
+     * Set the Engine::Camera used to the default one.
+     * This method allow to have a quick fix of issue #378 before switching to Radium v2 development.
+     * TODO : have a cleaner camera management and control in the GuiBase Radium library.
```
nmellado:
```suggestion
     * \todo have a cleaner camera management and control in the GuiBase Radium library.
```

PR #407, opened 2019-01-18
3 bug fixes ;

hoshiryu:

Fixes 3 bugs:
- `SkinningData` clean reset
- `LoopSubdivider` default valence
- `Engine::Mesh` indices update 

- When reseting the animation, the relative poses are not reset.
- The default valence for computing weights are not enough for the Hand model.
- When adding triangles to a mesh, these are not rendered even after `setDirty(INDEX)` because the value of `m_numElements`, which is used to tell the number of elements to display, is not updated.

All fixed.
Though it would be better to get the max valence from the mesh, though I don't see how to...

no

* **Other information**:

### DIFF
--- a/src/Core/Geometry/LoopSubdivider.hpp
+++ b/src/Core/Geometry/LoopSubdivider.hpp
@@ -29,10 +29,9 @@ class RA_CORE_API LoopSubdivider
     using SP_OPS = std::vector<P_OPS>;
 
   public:
-    LoopSubdivider() : base() { init_weights(); }
+    LoopSubdivider() : base() {}
 
     explicit LoopSubdivider( TopologicalMesh& mesh ) : base() {
-        init_weights();
         attach( mesh );
     }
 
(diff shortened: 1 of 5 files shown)


src/Core/Geometry/LoopSubdivider.hpp
```diff
@@ -67,7 +67,7 @@ class RA_CORE_API LoopSubdivider
 
   protected:
     /// Pre-compute weights.
-    void init_weights( size_t max_valence = 50 ) {
+    void init_weights( size_t max_valence = 100 ) {
```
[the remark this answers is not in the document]
hoshiryu: Yes, I will change this to match the input mesh

hoshiryu:
If everyone is ok, can someone merge this PR please?

src/Engine/Renderer/Mesh/Mesh.cpp
```diff
@@ -243,12 +243,14 @@ void Mesh::updateGL() {
         {
             if ( m_renderMode == RM_POINTS )
             {
+                m_numElements = m_mesh.vertices().size();
```
nmellado: @dlyr @MathiasPaulin do we want to do this update each time we render the mesh ?

An alternative would be to have a mechanism to update the GL object from it's `Core` version. I remember it is something we discussed, but I can figure out the current status.

Any remark/comment would help here.

hoshiryu: this update is not done each time we render the mesh, it is done only if we did set the `INDEX` as dirty

nmellado: Ok, my bad.

PR #408, opened 2019-01-29
Master v1 fixes

dlyr:

Fixes to reintroduce Ray alias and tranformRay function.
It also gives accessor in Animation Plugin (need further refactor, but I need it for my application)
If not adapted, I could keep this commit in my own branch.

* **What is the current behavior?** (You can also link to an open issue here)**

`Eigen::ParametrizedLine<Scalar, 3> transformedRay = {t * r.origin(), t.linear() * r.direction()};`

`Ra::Core::Ray transformedRay = Ra::Core::transformRay( ray, t );`

* **Other information**:


Plugins/Animation/src/AnimationComponent.hpp
```diff
@@ -112,6 +112,15 @@ class ANIM_PLUGIN_API AnimationComponent : public Ra::Engine::Component {
     virtual void setTransform( const Ra::Core::Utils::Index& roIdx,
                                const Ra::Core::Transform& transform ) override;
 
+    /// Setup CC.
+    void setupIO( const std::string& id );
```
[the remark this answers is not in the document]
dlyr: Well, I build my own animation stuff and need access to these methods. 
I will keep it in my branch only for the moment.

src/Core/Math/LinearAlgebra.hpp
```diff
@@ -46,20 +46,32 @@ template <typename Vector>
 inline Vector trunc( const Vector& v );
 
 /// Component-wise clamp() function on a floating-point vector.
-template <typename Vector>
-inline Vector clamp( const Vector& v, const Vector& min, const Vector& max );
 
+template <typename Derived>
+inline Eigen::MatrixBase<Derived> clamp( const Eigen::MatrixBase<Derived>& v,
+                                         const Eigen::MatrixBase<Derived>& min,
+                                         const Eigen::MatrixBase<Derived>& max );
```
hoshiryu: why changing from `Vector` to `Eigen::MatrixBase<Derived` ?

src/Core/Math/LinearAlgebra.hpp
```diff
@@ -99,26 +112,27 @@ inline Scalar getNormAndNormalize( Vector_& v );
...
-} // namespace Vector
+/// Transform a ray, direction is only transformed by linear part of the
+/// transformation, while origin is fully transformed
+/// corresponds to t*r
+/// \param t : transform matrix
+/// \param r : ray to transform
+/// \return transoformed ray, origine is translated while direction is only
+/// linarly transformed.
+template <typename Scalar>
+inline Eigen::ParametrizedLine<Scalar, 3>
```
hoshiryu: Why not using `Ray` ?

src/Core/Math/LinearAlgebra.hpp
```diff
@@ -46,20 +46,32 @@ template <typename Vector>
...
 /// Vector range check, works for any numeric vector.
 template <typename Vector_>
 inline bool checkRange( const Vector_& v, const Scalar& min, const Scalar& max );
 
-/// Call std::isnormal on vector entries.
-template <typename Vector_>
-inline bool checkInvalidNumbers( Eigen::Ref<const Vector_> v, const bool FAIL_ON_ASSERT = false );
+/// Call std::isnormal on quaternion entries.
+template <typename S>
+inline bool checkInvalidNumbers( Eigen::Ref<Eigen::Quaternion<S>> q,
```
hoshiryu: We should start using `Eigen::Ref` everywhere

src/Engine/Component/Component.cpp
```diff
@@ -1,5 +1,6 @@
 #include <Engine/Component/Component.hpp>
 
+#include <Core/Math/LinearAlgebra.hpp>
```
hoshiryu: As already stated, this will highly increase compilation time for `Engine` and client Plugins using `Component`

src/Core/Math/LinearAlgebra.hpp
dlyr: Because you can't template on Scalar type when using Ray.
this function is templated so it is meaningfull for Ray, Rayf, Rayd.

src/Core/Math/LinearAlgebra.hpp
dlyr: Here again, it's a mater of template parameter, when using Ra::Core::Vector, you can't differentiate `clamp(float, float, float)`, `clamp(vector, float, float)` and clamp `(vector, vector, vector)`
(the change comes from the fact that this clamp function have been move to the same namespace as `clamp(T,T,T)` ...)

dlyr:
@hoshiryu, I think the bug in helloRadium comes from 0 sized FBO which are valid on my config but not on yours.
Could you try to track FBO initialization, and add the corresponding guards ? 
(since it's working on my setup, it is harder for me to do so ;)
Then you could PR to this branch.

src/Engine/Component/Component.cpp
hoshiryu: oh yes, my bad

nmellado:
I propose we merge this PR, as it wasn't targeting to solve the minimal app issue.
All the other proposed changes have been validated.
@hoshiryu are you fine with it ?

hoshiryu:
I am ok with this.
Just a little question: why moving `Types.hpp` out of `Math`?

nmellado:
Keeping Types in Math would require to move all the types in namespace `Ra::Core::Math`, which might be a bit annoying to use in practice.

PR #411, opened 2019-02-22
Disambiguate different clamps.

dlyr:

Bug fix, `clamp(T, T, T)` and Vector version clash (since there is no partial template specialisation on function).

generic clamp has been move to `Ra::Core` (instead of math).
That's dirty, but temporary, please use std::clamp instead.
"Eigen" clamp has been redesign to work with all kind of expression (I hope ;)

`Ra::Core::Math::clamp` works on scalar type, not on vector types

use std::clamp or Ra::Core on scalar types
`Ra::Core::Math::clamp(Vector, Scalar, Scalar)` or `Ra::Core::Math::clamp(Vector, Vector, Vector)` works as expected

* **Other information**:

### DIFF
--- a/src/Core/Animation/Skeleton.cpp
+++ b/src/Core/Animation/Skeleton.cpp
@@ -225,7 +225,7 @@ Vector3 Skeleton::projectOnBone( int boneIdx, const Ra::Core::Vector3& pos ) con
     CORE_ASSERT( length_sq != 0.f, "bone has lenght 0, cannot project." );
 
     // Project on the line segment
-    const Scalar t = std::clamp( op.dot( dir ) / length_sq, (Scalar)0.0, (Scalar)1.0 );
+    const Scalar t = std::clamp( op.dot( dir ) / length_sq, Scalar(0), Scalar(1) );
     return start + ( t * dir );
 }
 
--- a/src/Core/Animation/Skeleton.cpp
+++ b/src/Core/Animation/Skeleton.cpp
@@ -1,5 +1,7 @@
 #include <Core/Animation/Skeleton.hpp>
 #include <Core/Math/LinearAlgebra.hpp> // Math::clamp
+
+#include <algorithm>
 #include <stack>
 
 namespace Ra {
@@ -223,7 +225,7 @@ Vector3 Skeleton::projectOnBone( int boneIdx, const Ra::Core::Vector3& pos ) con
     CORE_ASSERT( length_sq != 0.f, "bone has lenght 0, cannot project." );
 
     // Project on the line segment
-    const Scalar t = Math::clamp( op.dot( dir ) / length_sq, (Scalar)0.0, (Scalar)1.0 );
+    const Scalar t = std::clamp( op.dot( dir ) / length_sq, (Scalar)0.0, (Scalar)1.0 );
     return start + ( t * dir );
 }
 
--- a/src/Core/Geometry/PolyLine.inl
+++ b/src/Core/Geometry/PolyLine.inl
@@ -1,5 +1,5 @@
 #include <Core/Geometry/PolyLine.hpp>
-#include <Core/Math/LinearAlgebra.hpp>
+#include <Core/Math/LinearAlgebra.hpp> // cotan, saturate (from Math.hpp)
 
 namespace Ra {
 namespace Core {
--- a/src/Core/Geometry/PolyLine.inl
+++ b/src/Core/Geometry/PolyLine.inl
@@ -1,4 +1,5 @@
 #include <Core/Geometry/PolyLine.hpp>
+#include <Core/Math/LinearAlgebra.hpp>
 
 namespace Ra {
 namespace Core {
--- a/src/Core/Math/LinearAlgebra.inl
+++ b/src/Core/Math/LinearAlgebra.inl
@@ -38,17 +38,18 @@ inline Vector_ trunc( const Vector_& v ) {
         std::function<Scalar_( Scalar_ )>( static_cast<Scalar_ ( & )( Scalar_ )>( std::trunc ) ) );
 }
 
-template <typename Derived>
-inline Eigen::MatrixBase<Derived> clamp( const Eigen::MatrixBase<Derived>& v,
-                                         const Eigen::MatrixBase<Derived>& min,
-                                         const Eigen::MatrixBase<Derived>& max ) {
+template <typename Derived, typename DerivedA, typename DerivedB>
+inline typename Derived::PlainMatrix clamp( const Eigen::MatrixBase<Derived>& v,
+                                            const Eigen::MatrixBase<DerivedA>& min,
+                                            const Eigen::MatrixBase<DerivedB>& max ) {
     return v.cwiseMin( max ).cwiseMax( min );
 }
 
 template <typename Derived>
-inline Eigen::MatrixBase<Derived> clamp( const Eigen::MatrixBase<Derived>& v, const Scalar& min,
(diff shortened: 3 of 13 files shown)


src/Core/Math/Math.inl
```diff
@@ -2,6 +2,12 @@
 
 namespace Ra {
 namespace Core {
+
+template <typename T>
+inline constexpr T clamp( T v, T min, T max ) {
+    return std::max( min, std::min( v, max ) );
+}
+
```
louen: I guess this function can go away if we use `std::clamp` instead

dlyr: well, yes, I was just lazy ;) but it will be done soon

src/Core/Animation/Skeleton.cpp
```diff
@@ -223,7 +225,7 @@ Vector3 Skeleton::projectOnBone( int boneIdx, const Ra::Core::Vector3& pos ) con
     CORE_ASSERT( length_sq != 0.f, "bone has lenght 0, cannot project." );
 
     // Project on the line segment
-    const Scalar t = Math::clamp( op.dot( dir ) / length_sq, (Scalar)0.0, (Scalar)1.0 );
+    const Scalar t = std::clamp( op.dot( dir ) / length_sq, (Scalar)0.0, (Scalar)1.0 );
```
hoshiryu: std::clamp or Ra::Core::saturate? (same every time we clamp between 0 and 1)

tests/CoreTests/src/algebra.cpp
```diff
@@ -36,6 +38,70 @@ void run() {
...
+    RA_VERIFY( min == std::clamp( min, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == std::clamp( max, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == std::clamp( max + s, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( min == std::clamp( min - s, min, max ), "Clamp fails on Scalar" );
+
+    RA_VERIFY( s == Ra::Core::clamp( s, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( min == Ra::Core::clamp( min, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == Ra::Core::clamp( max, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == Ra::Core::clamp( max + s, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( min == Ra::Core::clamp( min - s, min, max ), "Clamp fails on Scalar" );
```
hoshiryu: these don't exist anymore, right,

tests/CoreTests/src/algebra.cpp
```diff
@@ -36,6 +38,70 @@ void run() {
...
+    //    using Ra::Core::Math::clamp;
+    Scalar min = -12.;
+    Scalar max = +27.;
+    Scalar s = 0.6;
+
+    RA_VERIFY( s == std::clamp( s, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( min == std::clamp( min, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == std::clamp( max, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( max == std::clamp( max + s, min, max ), "Clamp fails on Scalar" );
+    RA_VERIFY( min == std::clamp( min - s, min, max ), "Clamp fails on Scalar" );
```
hoshiryu: do we really need to test the std version?

tests/CoreTests/src/algebra.cpp
dlyr: right, just forgot to buildtests

tests/CoreTests/src/algebra.cpp
dlyr: do not cost so much, so yes ;)

src/Core/Animation/Skeleton.cpp
dlyr: well, i'm not a fan of saturate, so feel free to change it as you wish (but I would re recommend to remove saturate and replace by clamp ...)

dlyr:
> Could you also check Tex.hpp and Tex.inl?

Well, could you add a test for this ? as I understand, the only use of clamp is on vector type (maybe you can remove the specification with VectorIdx ...

src/Core/Animation/Skeleton.cpp
nmellado: Don't use old cast style.
```suggestion
    const Scalar t = std::clamp( op.dot( dir ) / length_sq, Scalar(0), Scalar(1) );
```

src/Core/Geometry/PolyLine.inl
```diff
@@ -1,4 +1,5 @@
 #include <Core/Geometry/PolyLine.hpp>
+#include <Core/Math/LinearAlgebra.hpp>
```
nmellado: Why adding this ?

src/Core/Math/LinearAlgebra.inl
```diff
@@ -38,17 +38,18 @@ inline Vector_ trunc( const Vector_& v ) {
...
 }
 
 template <typename Derived>
-inline Eigen::MatrixBase<Derived> clamp( const Eigen::MatrixBase<Derived>& v, const Scalar& min,
-                                         const Scalar& max ) {
-    return v.cwiseMin( max ).cwiseMax( min );
+inline typename Derived::PlainMatrix clamp( const Eigen::MatrixBase<Derived>& v, const Scalar& min,
+                                            const Scalar& max ) {
+    return v.unaryExpr( [min, max]( Scalar x ) { return std::clamp( x, min, max ); } );
+    //    return v.cwiseMin( max ).cwiseMax( min );
```
nmellado:
```suggestion
```

src/Core/Animation/Skeleton.cpp
dlyr: Well your right, but it was as is (btw I accept your suggestion), but I will not review all code to check for that.

src/Core/Geometry/PolyLine.inl
dlyr: because I mess up with Math.hpp

src/Core/Animation/Skeleton.cpp
nmellado: Sure, I just wanted to avoid this new addition.

src/Core/Geometry/PolyLine.inl
nmellado:
```suggestion
```

dlyr: I'm checking, but Math.hpp is needed here ...

nmellado: Ok, I let you find out ;)

src/Core/Animation/Skeleton.cpp
dlyr: (for the record, it was like that, so it's not an addition ...)

src/Core/Geometry/PolyLine.inl
dlyr: oh cotan is from LinearAlgebra, so it's needed here

nmellado:
```suggestion
#include <Core/Math/LinearAlgebra.hpp> // Core::Math::cotan
```

nmellado:
Are you ok if we squash this one ?

Issue #337, opened 2018-06-13
Feature request : proper internal resources management


MathiasPaulin:
## Objective of the Request feature.

The Radium Engine and the plugins or applications built upon Radium, require several file resources to be available at runtime.
Finding a way to structure and to abstract these resources, to allow easy development and efficient deploy is mandatory.

This request feature is a first proposal for robust and efficient resource management and structuring into Radium, its plugins and applications.

## What are resources ?

When we speak about resources, we here consider the following resources. The resource list might later be extended in the discussion.
- Shaders : this is the main resources the Radium depends on when using the engine. Shaders are text files split into several components allowing code re-use and dynamic specialization of their content.
- Data file : Even if these are not in the engine, some plugins or applications need external (precomputed) data to function. A data file is then an octet stream, which structure is usage dependent that will be accessed at runtime

## Actual state :

Shaders are for the moment accessed at runtime explicitly from the OS filesystem.

All Engine, gui and application related shaders are located into the `Shaders` directory into the Radium sources hierarchy that is linked or copy into the `bin` directory at compile time to be accessed at runtime through the standard input/output functionalities of C++.

Even if this is efficient for development, this cause several difficulties
- Code maintenance : as everything is melt in this directory, it is difficult to maintain all the shader code as we do not know which part of Radium need which shader (Engine, Gui, Applications, Plugins ... ). More, as these shaders were programmed independently by several developers, there is not so much code re-using and when the Engine require an evolution of the shader code one have to go through a constantly increasing number of files to fix the code.
- Application deployment : When we want to deploy an application (eg, a technological demo), we have to ensure that we could reproduce the same directory structure from the Radium source hierarchy into the target installation or executable directory. This is not a good idea as we then must expose a lot of files unrelated to the demo (Radium internal shaders)
- System independence : accessing directly the OS file system prevent Radium to be deployed on systems with specific file systems (e.g IOS) or without access to a filesystem. More, each OS put different constraints on the way an application must be deployed (e.g. MacOs .app structure), and how it might access to the filesystem.

## File resources management proposal

This proposal aims at defining the structure and access method to Radium related resources both for an easy development and for efficient deployment. It the propose in source structuring and how to transform this structure to a runtime collection integrated in the deployed binaries.

### In source structure :
Instead of the (contains everything) directory `Shaders`, we will define the following hierarchy in the root Radium source directory :

```
Radium-Engine
└───src
│   └───...
└───Resources
│   └───Engine
[...]

PR #417, opened 2019-03-25
Improve Skinning

hoshiryu:

- Addition of the "Stretchable and Twistable Bone Skinning" (STBS) geometric skinning method;
- Cleanup OpenMP directives for DQS and cotangent weights computation;
- Improvement of the Skinning plugin to use this method on LBS and DQS, and also to display skinning weights;
- Addition of an AnimationComponent output for bone index from RO, so that other plugins can test if a bone has been selected.


Plugins/Skinning/src/SkinningComponent.cpp
```diff
@@ -129,6 +144,28 @@ void SkinningComponent::initialize() {
...
+        auto matT = std::static_pointer_cast<Ra::Engine::BlinnPhongMaterial>(
+            ro->getRenderTechnique()->getMaterial() );
+        std::shared_ptr<Ra::Engine::BlinnPhongMaterial> mat(
+            new Ra::Engine::BlinnPhongMaterial( "IS2016_Mat" ) );
+        mat->m_kd = Ra::Core::Utils::Color::Skin();
+        mat->m_ks = Ra::Core::Utils::Color::White();
+        m_weightTechnique->setMaterial( mat );
+        // assign texture
+        Ra::Engine::TextureParameters texParam;
+        texParam.name = "../../../Assets/Textures/Influence0.png";
```
nmellado: This path is wrong.
If you want to use assets they need to be installed alongside the application so the path as runtime is known and guaranteed to be correct.

Plugins/Skinning/src/SkinningPlugin.cpp
```diff
@@ -49,91 +57,216 @@ QAction* SkinningPluginC::getAction( int id ) {
...
+    if ( !m_system )
+    {
+        return;
+    }
+    Ra::Engine::TextureParameters texData;
+    texData.wrapS = GL_CLAMP_TO_EDGE;
+    texData.wrapT = GL_CLAMP_TO_EDGE;
+    texData.minFilter = GL_NEAREST;
+    texData.magFilter = GL_NEAREST;
+    texData.name = "../../../Assets/Textures/Influence0.png";
```
nmellado: Again, this path is wrong.

Plugins/Skinning/src/SkinningComponent.cpp
nmellado: Also, the mentioned file is not part of this commit, and not in the trunk....

hoshiryu: The resources issue is already known , see #337 .
The mentionned file is in Assets/Textures since February 26, 2016.

nmellado: What I am asking is to copy/link the file in a subdirectory of `bin`, like we do for the shaders.
I cannot accept we add a path like that in the code.

PR #423, opened 2019-05-03
Improve engine genericity

nmellado:

This PR aim at simplifying the use of Radium to produce result images and videos: customizable background color, per-vertex color attribute in blinnphong, fixed material editor for blinnphong. It also brings new abstraction layers, preparing the API for upcoming changes (new `Displayable` objects).

Buildchain:
 * submodules are automatically fetched when compiling for the first time

Improved API:
 * [core] Builtin support for Scalar from literals (suffix `_ra`) + `Ra::almost_equals(Scalar, Scalar)` 
 * [core] Add support for `std::optionnal` 
 * [core/engine] Add new abstract classes `Core::AbstractGeometry` and `Engine::Displayable` 
 * [engine] Clean attribute management in `Engine::Mesh`
 * [engine] Fix ad-hock mechanism for shader named-strings 
 * [engine] Simplify mechanism to use material names 

New Features:
 * Add customizable background color in Renderer and main app 
 * Add splatting shader for point-clouds 
 * Add support for per-vertex color attributes in blinn-phong material 

Fixes:
 - Minor fix in tinyplyloader (better load colors) 
 - Fix rotation lock 
 - Fix material editor for BlinnPhong 

- Direct access to TriangleMesh from Engine::Mesh now requires casting, or to use component connectors
- Ray-cast queries method is not yet available in `Mesh` and any `Displayable` 
- The material of an object cannot be changed anymore from the GUI 

* **Other information**:


MathiasPaulin:
Well, I think we need to integrate PR #422 before this.

nmellado:
I do agree.
But after a first review of #422 these two PRs are othogonal and #422 should not change #423 too much (the opposite is not necessary true however).

Shaders/Materials/BlinnPhong/DepthAmbientBlinnPhong.frag.glsl
```diff
@@ -4,14 +4,16 @@ layout (location = 2) out vec4 out_diffuse;
...
 
 layout (location = 0) in vec3 in_position;
 layout (location = 1) in vec3 in_texcoord;
 layout (location = 2) in vec3 in_normal;
 layout (location = 3) in vec3 in_tangent;
 layout (location = 4) in vec3 in_viewVector;
 layout (location = 5) in vec3 in_lightVector;
+layout (location = 6) in vec3 in_vertexcolor;
+
+#include "BlinnPhongMaterial.glsl"
```
hoshiryu: why that change?

Shaders/Materials/BlinnPhong/LitOITBlinnPhong.frag.glsl
```diff
@@ -3,14 +3,16 @@ layout (location = 1) out vec4 f_Revealage;
...
 
 layout (location = 0) in vec3 in_position;
 layout (location = 1) in vec3 in_texcoord;
 layout (location = 2) in vec3 in_normal;
 layout (location = 3) in vec3 in_tangent;
 layout (location = 4) in vec3 in_viewVector;
 layout (location = 5) in vec3 in_lightVector;
+layout (location = 6) in vec3 in_vertexcolor;
+
+#include "BlinnPhongMaterial.glsl"
```
hoshiryu: why that change?

Shaders/Materials/BlinnPhong/BlinnPhongMaterial.glsl
```diff
@@ -30,11 +30,18 @@ struct Material
     float ns;
     float alpha;
 
+    int hasPerVertexKd;
+    int renderAsSplat;
```
hoshiryu: I'm pretty sure this (`m_renderAsSplat`) has nothing to do here and must be managed elsewhere.
About `m_hasPerVertexKd` I'm not sure

src/Engine/Renderer/Material/BlinnPhongMaterial.hpp
```diff
@@ -87,6 +80,8 @@ class RA_ENGINE_API BlinnPhongMaterial final : public Material {
     Core::Utils::Color m_ks{0.0, 0.0, 0.0, 1.0};
     Scalar m_ns{1.0};
     Scalar m_alpha{1.0};
+    bool m_hasPerVertexKd{false};
+    bool m_renderAsSplat{false};
```
hoshiryu: I'm pretty sure this (`m_renderAsSplat`) has nothing to do here and must be managed elsewhere.
About `m_hasPerVertexKd` I'm not sure

src/Core/Geometry/TriangleMesh.cpp
```diff
@@ -3,6 +3,8 @@
 #include <Core/Geometry/TriangleOperation.hpp> // triangleArea
 #include <Core/Types.hpp>
 
+#include <Engine/Renderer/Mesh/Mesh.hpp>
```
hoshiryu: This is a bad idea: `Core` must not depend on `Engine`.

src/Engine/Renderer/Mesh/Mesh.cpp
```diff
@@ -95,8 +97,23 @@ void Mesh::render() {
...
-void Mesh::loadGeometry( const Core::Geometry::TriangleMesh& mesh ) {
-    m_mesh = mesh;
+size_t Mesh::getNumFaces() const {
+    switch ( getRenderMode() )
+    {
+    case MeshRenderMode::RM_TRIANGLE_STRIP:
+        [[fallthrough]];
+    case MeshRenderMode::RM_TRIANGLE_FAN:
+        return ( getTriangleMesh().m_triangles.size() - 1 ) * 3 + 1;
+        return ( getTriangleMesh().m_triangles.size() - 1 ) * 3 + 1;
```
hoshiryu: only one is enough ;)

Shaders/Materials/BlinnPhong/BlinnPhongMaterial.glsl
nmellado: Why are you so sure ?
In the current state, `m_renderAsSplat` is used in `toDiscard`, and thus needs to be defined in BlinnPhongMaterial.

hoshiryu: I mean this is not related to the material, only to point clouds.
Thus the material should not have to decide whether or not the fragment should be discarded when splatting, it must be decided on top of it.

nmellado: Yes, but this is only way to do it today.
This should be fixed in v2

dlyr: could you simply use another material ? not sure how to specify it in the loader

nmellado: This minor comment allow to use bling phong as is.
If you prefer and can copy all the shaders for 2 lines of code.

MathiasPaulin: I agree that renderAsSplat is not really a part of a BlinnPhong shader. Nevertheless, it is the simplest way to achieve the objective of splat-based rendering. This clearly justify the need for shader composition and on-line glsl code generation to be integrated in v2. For the moment, we can keep it like this (it will be not be the only thing in the vast amount of shaders used by radium to be not really clean). I have an idea to make this clean that relies on shader configuration and will try to implement it as soon as possible on this branch  but this will be a step towards v2 and material-shader polymorphism.
hasPerVertexKd is welcome in BlinnPhong and could be generalized to other materials with a small naming effort. So, we can keep it for now.
To conclude, yes, this is not the good way to implement BlinnPhong for splatting but it works and do not necessitate huge code re-writing based on non validated interfaces. So we can keep it for now.

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -61,9 +57,9 @@ class RA_ENGINE_API GeometryComponent : public Component {
 
   private:
     Ra::Core::Utils::Index m_meshIndex{};
-    Ra::Core::Utils::Index m_aabbIndex{};
     std::string m_contentName{};
-    bool m_deformable{false};
+    // directly hold a reference to the displayMesh to simplify accesses in handlers
+    std::shared_ptr<Mesh> _displayMesh{nullptr};
```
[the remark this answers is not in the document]
nmellado: Is this written in the coding conventions ?

src/Engine/RadiumEngine.cpp
```diff
@@ -41,9 +41,31 @@ void RadiumEngine::initialize() {
...
+    shaderProgramManager->addNamedString( "/Helpers.glsl", "Shaders/Helpers.glsl" );
+    shaderProgramManager->addNamedString( "/Structs.glsl", "Shaders/Structs.glsl" );
+    shaderProgramManager->addNamedString( "/Tonemap.glsl", "Shaders/Tonemap.glsl" );
+    shaderProgramManager->addNamedString( "/LightingFunctions.glsl",
+                                          "Shaders/LightingFunctions.glsl" );
+    shaderProgramManager->addNamedString( "/TransformStructs.glsl",
+                                          "Shaders/Transform/TransformStructs.glsl" );
+    shaderProgramManager->addNamedString( "/DefaultLight.glsl",
+                                          "Shaders/Lights/DefaultLight.glsl" );
+
```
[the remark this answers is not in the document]
nmellado: This is hard-coded resource management, but without code duplication, and hand-crafted indices.
The complete resource management is planned for v2.

src/GuiBase/Viewer/Gizmo/Gizmo.hpp
```diff
@@ -72,12 +78,26 @@ class Gizmo {
...
+
   protected:
     Core::Transform m_worldTo;   ///< World to local space where the transform lives.
     Core::Transform m_transform; ///< Transform to be edited.
     Engine::Component* m_comp;   ///< Engine Ui component.
     Mode m_mode;                 ///< local or global.
-    std::vector<Core::Utils::Index> m_renderObjects; ///< ros for the gizmo.
+  private:
+    std::vector<Core::Utils::Index> m_renderObjects;     ///< ros for the gizmo.
+    std::vector<std::shared_ptr<Engine::Mesh>> m_meshes; ///< Display meshes of the gizmo
```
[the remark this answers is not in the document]
nmellado: They are private in purpose: these two fields need to be managed jointly, and we do not want to let the child classes messing them up.
The getters could be protected, but I see no reason why they couldn't be public. Do you see any ?

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -22,37 +22,33 @@ namespace Engine {
  *  - normals: rw (if deformable)
  *  - triangles: rw (if deformable)
  */
-class RA_ENGINE_API GeometryComponent : public Component {
+class RA_ENGINE_API TriangleMeshComponent : public Component {
   public:
-    GeometryComponent( const std::string& name, bool deformable, Entity* entity );
-    ~GeometryComponent() override;
+    TriangleMeshComponent( const std::string& name, Entity* entity,
+                           const Ra::Core::Asset::GeometryData* data );
```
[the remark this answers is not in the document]
hoshiryu: Well actually I was thinking about a plugin developper who may want to add `TriangleMeshComponents` to (only) display generated/debug meshes.
This developer would then have to create its own `Component` class which will do exactly the same as `TriangleMeshComponent` or create an empty `TriangleMeshComponent` and then find a way to write the `TirangleMesh` in.

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
hoshiryu: yes:
`Class members have the m_ prefix. Other prefixes (apart from g_) are discouraged.`

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -22,37 +22,33 @@ namespace Engine {
...
-    void addMeshRenderObject( const Ra::Core::Geometry::TriangleMesh& mesh,
-                              const std::string& name );
-    void handleMeshLoading( const Ra::Core::Asset::GeometryData* data );
-
-    /// Returns the index of the associated RO (the display mesh)
-    Ra::Core::Utils::Index getRenderObjectIndex() const;
-
     /// Returns the current display geometry.
     const Ra::Core::Geometry::TriangleMesh& getMesh() const;
+    Mesh* getDisplayMesh();
```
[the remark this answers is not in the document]
hoshiryu: What is an abnormal situation? Do you have an example of how `_displayMesh` can become `nullptr`?

src/GuiBase/Viewer/Gizmo/Gizmo.hpp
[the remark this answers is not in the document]
hoshiryu: Ok for the members.
Nevermind about the accessors, they are protected in the current implementation, didn't see it.
Actually the RO management is something private to the class and not related to the API offered by the `Gizmo` class, and so these accessors, which are made for the derived classes, should be kept protected and not public (same for `addRenderObject`).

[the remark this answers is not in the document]
hoshiryu: Nevermind about the accessors, they are protected in the current implementation, didn't see it.

src/Engine/RadiumEngine.cpp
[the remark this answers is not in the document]
hoshiryu: ok then can you update the commit message for `b5dde0f` ?

src/GuiBase/Viewer/TrackballCamera.cpp
```diff
@@ -230,9 +231,9 @@ void Gui::TrackballCamera::fitScene( const Core::Aabb& aabb ) {
     Scalar f = m_camera->getFOV();
     Scalar a = m_camera->getAspect();
 
-    const Scalar r = ( aabb.max() - aabb.min() ).norm() / 2.0;
-    const Scalar x = r / std::sin( f / 2.0 );
-    const Scalar y = r / std::sin( f * a / 2.0 );
+    const Scalar r = ( aabb.max() - aabb.min() ).norm() / Scalar( 2 );
+    const Scalar x = r / std::sin( f / Scalar( 2 ) );
+    const Scalar y = r / std::sin( f * a / Scalar( 2 ) );
```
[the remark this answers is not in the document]
MathiasPaulin: I prefer just to write "2". More simpler, more readable, less error prone as the compiler will convert automatically to the desired type, possibly emitting a warning if the conversion implies a loss of precision ... see [https://en.cppreference.com/w/cpp/language/operator_arithmetic](https://en.cppreference.com/w/cpp/language/operator_arithmetic)

src/Core/Geometry/TriangleMesh.cpp
MathiasPaulin: I agree this is not a good solution. But the way attrib binding is done is also not the better way to do that (attrib name - and then sementic if names are well chosen - is lost and replace by a number). 
So, instead of adding a dependency to the Engine, and as the need is only for the colorize method, why not just write : 
```suggestion
/// TODO : in Radium V2, put here proper attrib binding interface
void TriangleMesh::colorize( const Utils::Color& color ) {
    static const std::string colorAttribName{ "Vec4_attr_0"};
    auto colorAttribHandle = addAttrib<Core::Vector4>( colorAttribName );
    getAttrib( colorAttribHandle ).data() = Vector4Array( vertices().size(), color );
}
```
This is not really clean and depend on the way attribs are managed by the Engine mesh but for now, the color attrib is named `"Vec4_attr_0"` in the Engine mesh.

Applications/MainApplication/Gui/MainWindow.cpp
```diff
@@ -314,18 +318,18 @@ void MainWindow::onSelectionChanged( const QItemSelection& selected,
...
-        }
+                                                ->getMaterial()
+                                                ->getMaterialName();
+            CORE_ASSERT( m_currentShaderBox->findText( shaderName.c_str() ) != -1,
+                         "RO shaders must be already added to the list" );
+            m_currentShaderBox->setCurrentText( shaderName.c_str() );
+            // m_currentShaderBox->setEnabled( true ); // commented out, as there is no simple way
+            // to change the material type
+        } else
+            m_currentShaderBox->setCurrentText( "" );
```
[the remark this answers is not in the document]
MathiasPaulin: Well, this is a typical example of semantic errors brought by badly chosen names. The variable shaderName (and the use of a limited subset of the shader configuration capabilities elsewhere) have generated this confusion. As we can not change the shader type but only the shader configuration, we must ensure that it is always safe. As it is not the case, changing the shader configuration is disabled for now.

src/Engine/Renderer/Material/BlinnPhongMaterial.cpp
```diff
@@ -95,16 +93,16 @@ void BlinnPhongMaterial::bind( const ShaderProgram* shader ) {
 }
 
 bool BlinnPhongMaterial::isTransparent() const {
-    return ( m_alpha < 1.0 ) || ( m_kd[3] < 1.0 ) || Material::isTransparent();
+    return ( m_alpha < Scalar( 1 ) ) || ( m_kd[3] < Scalar( 1 ) ) || Material::isTransparent();
```
MathiasPaulin: Simply write `xxx < 1`.
C++ will always convert the operand to the good type (https://en.cppreference.com/w/cpp/language/operator_comparison : 

> If the operands has arithmetic or enumeration type (scoped or unscoped), usual arithmetic conversions are performed on both operands following the rules for arithmetic operators. The values are compared after conversions

src/Core/Utils/StdOptional.hpp
```diff
@@ -0,0 +1,50 @@
+#pragma once
```
MathiasPaulin: This is not the "standard" include guard for Radium. We have to change the coding directive if we want to switch to this (simpler) guard.

tests/Tests.hpp
```diff
@@ -37,125 +39,156 @@ struct NonCopy {
...
+inline bool isApprox( const Core::Utils::optional<S>& lhs, const S& rhs ) {
+    return std::fabs( ( *lhs ) - rhs ) < std::numeric_limits<S>::epsilon();
+}
 
-#if defined (_MSC_VER) && ! defined (__INTEL_COMPILER)
-#  define MYPRAGMA(X) __pragma(X)
+template <typename S>
+inline bool isApprox( const S& lhs, const Core::Utils::optional<S>& rhs ) {
+    return std::fabs( lhs - ( *rhs ) ) < std::numeric_limits<S>::epsilon();
+}
```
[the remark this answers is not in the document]
MathiasPaulin: as said in https://en.cppreference.com/w/cpp/utility/optional/operator*

> The behavior is undefined if *this does not contain a value.

So, either optional is not needed here, either there will be an undefined behavior of the operator ...

Shaders/Materials/BlinnPhong/DepthAmbientBlinnPhong.frag.glsl
hoshiryu: why moving the include after the layout declarations?

Plugins/Skinning/src/Display/SkinningDisplayComponent.hpp
```diff
@@ -143,8 +144,11 @@ class SKIN_PLUGIN_API SkinningDisplayComponent : public Ra::Engine::Component {
             renderObject->setVisible( true );
             renderObject->setRenderTechnique( technique );
             std::shared_ptr<Ra::Engine::Mesh> displayMesh( new Ra::Engine::Mesh( name ) );
-            displayMesh->loadGeometry( mesh );
-            displayMesh->addData( Ra::Engine::Mesh::Vec4Data::VERTEX_COLOR, color );
+            TriangleMesh meshCopy( mesh );
+            auto colorAttribHandle = meshCopy.addAttrib<Ra::Core::Vector4>(
+                Ra::Engine::Mesh::getAttribName( Ra::Engine::Mesh::VERTEX_COLOR ) );
+            auto colorAttrib = meshCopy.getAttrib( colorAttribHandle ).data() = color;
+            displayMesh->loadGeometry( std::move( meshCopy ) );
```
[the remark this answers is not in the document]
nmellado: What exactly would you like me to comment on ?

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: This use case indeed make sens.
What about adding a dedicated constructor:
```c++
TriangleMeshComponent( const std::string& name, 
                       Entity* entity, 
                       Mesh* mesh // will be owned using a std::shared_ptr
                       )
```
where `m_content_name` is also constructed from `name`.

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: Ok, let's explaining it differently.
`m_displayMesh` is not supposed to be `nullptr`, so there is no error handling.
However, as we use a pointer, there is in theory a possibility that this pointer equals `nullptr`, and thus we protect it using `CORE_ASSERT`. Remember that this macro is activated only for debugging, so there is not cost at runtime when compiling in Release mode.

So, to conclude, I don't believe we should add a comment explaining that this method can crash, as it is not supposed to. However, if for any reason this pointer is 0, the program will crash whatever we do (even without `CORE_ASSERT`), however we would have a better error message for debugging. 

Makes sens ?

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -22,37 +22,33 @@ namespace Engine {
...
 
   public:
     // Component communication management
     void setupIO( const std::string& id );
     void setContentName( const std::string& name );
     void setDeformable( bool b );
 
+    /// Returns the index of the associated RO (the display mesh)
```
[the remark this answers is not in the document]
nmellado: Same comment as above

Plugins/Skinning/src/Display/SkinningDisplayComponent.hpp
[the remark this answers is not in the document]
hoshiryu: I though it was about adding the color attribute to the GeometryMesh, but it's not.
My mistake

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
hoshiryu: Ok with this, except that the mesh should not be a pointer but an rvalue, so that we can ensure it exists and `m_displayMesh` cannot be `nullptr`

tests/Tests.hpp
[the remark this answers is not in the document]
nmellado: This function is needed when testing the result of a function returning an `optional`, something that happens in tests in practice.
I propose to modify it by adding a first test on the `optional` (notice the use of the new namespace function `Ra::almost_equals` introduced in today's revisions):
```c++
template <typename S>
inline bool isApprox( const Core::Utils::optional<S>& lhs, const S& rhs ) {
    return lhs.has_value() && almost_equals( *lhs, rhs );
}
```
Are you guys fine with it ?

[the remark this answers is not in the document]
MathiasPaulin: OK for me

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: Sounds good. Thanks.

Shaders/Materials/BlinnPhong/BlinnPhong.frag.glsl
```diff
@@ -12,12 +11,17 @@ layout (location = 2) in vec3 in_normal;
...
+layout (location = 6) in vec3 in_vertexcolor;
+
+#include "BlinnPhongMaterial.glsl"
 
 void main() {
     // Discard non fully opaque fragments
     if ( toDiscard(material, in_texcoord.xy) || (material.alpha < 1) )
         discard;
 
+    //in_vertexcolor = vec3(1.0,0.,0.);
```
hoshiryu: remain from tests?

Shaders/Materials/BlinnPhong/BlinnPhong.frag.glsl
```diff
@@ -12,12 +11,17 @@ layout (location = 2) in vec3 in_normal;
 layout (location = 3) in vec3 in_tangent;
 layout (location = 4) in vec3 in_viewVector;
 layout (location = 5) in vec3 in_lightVector;
+layout (location = 6) in vec3 in_vertexcolor;
+
+#include "BlinnPhongMaterial.glsl"
```
hoshiryu: same here, why moving the include after the layouts declaration?

src/Core/CoreMacros.hpp
```diff
@@ -257,6 +257,46 @@ using Scalar = float;
...
+ * \see https://en.cppreference.com/w/cpp/language/user_literal
+ * To create a Scalar from a litteral, do:
+ * \code
+ * auto s = 1_ra; // s is a Scalar
+ * \endcode
+ */
+constexpr Scalar operator"" _ra ( unsigned long long n )
+{
+    return Scalar( n );
+}
```
hoshiryu: I'm not fond of these but they do the job.

src/Core/CoreMacros.hpp
```diff
@@ -257,6 +257,46 @@ using Scalar = float;
...
+  template<class T>
+  typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type
+      almost_equal(T x, T y, int ulp = 2)
+  {
+      // the machine epsilon has to be scaled to the magnitude of the values used
+      // and multiplied by the desired precision in ULPs (units in the last place)
+      return std::abs(x-y) <= std::numeric_limits<T>::epsilon() * std::abs(x+y) * ulp
+          // unless the result is subnormal
+          || std::abs(x-y) < std::numeric_limits<T>::min();
+  }
```
hoshiryu: what about `Ra::Core::AreApprox`?

src/Engine/Renderer/Camera/Camera.cpp
```diff
@@ -28,11 +28,12 @@ void Camera::initialize() {
...
-    triMesh.vertices() = {{0, 0, 0},       {-0.5, -0.5, -1}, {-0.5, 0.5, -1}, {0.5, 0.5, -1},
-                          {0.5, -0.5, -1}, {-0.3, 0.5, -1},  {0, 0.7, -1},    {0.3, 0.5, -1}};
+    triMesh.vertices() =
+    {{0_ra, 0_ra, 0_ra}, {-.5_ra, -.5_ra, -1_ra}, {-.5_ra, .5_ra, -1_ra}, {.5_ra, .5_ra, -1_ra},
+     {.5_ra, -.5_ra, -1_ra}, {-.3_ra, .5_ra, -1_ra},  {0_ra, .7_ra, -1_ra}, {.3_ra, .5_ra, -1_ra}};
     triMesh.m_triangles = {{0, 1, 2}, {0, 2, 3}, {0, 3, 4}, {0, 4, 1}, {5, 6, 7}};
-    m->loadGeometry( triMesh );
-    Core::Vector4Array c( triMesh.vertices().size(), {0.2, 0.2, 0.2, 1.0} );
+    m->loadGeometry( std::move( triMesh ) );
+    Core::Vector4Array c( 8, {.2_ra, .2_ra, .2_ra, Scalar( 1. )} );
```
hoshiryu: please change `Scalar( 1. )` into `1_ra`

src/Engine/Renderer/Displayable/DisplayableObject.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    //// Utility methods, used to display statistics
+    virtual size_t getNumFaces() const { return 0; }
+    virtual size_t getNumVertices() const { return 0; }
+
+  protected:
+    PickingRenderMode _pickingRenderMode{NO_PICKING};
+
+  private:
+    std::string _name{}; /// Name of the mesh.
```
hoshiryu: please use `m_` prefix

src/Engine/Renderer/Displayable/DisplayableObject.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    /// Draw the mesh.
+    virtual void render() = 0;
+
+    //// Utility methods, used to display statistics
+    virtual size_t getNumFaces() const { return 0; }
+    virtual size_t getNumVertices() const { return 0; }
+
+  protected:
+    PickingRenderMode _pickingRenderMode{NO_PICKING};
```
hoshiryu: please use `m_` prefix

src/Engine/Renderer/Material/BlinnPhongMaterial.cpp
hoshiryu: would `1_ra` be ok?

tests/Tests.hpp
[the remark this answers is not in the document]
hoshiryu: ok for me too

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: This is now fixed, see [here](https://github.com/STORM-IRIT/Radium-Engine/pull/423/files#diff-fa3c219ebaa8a537433c207d6eea46e0R40).

tests/Tests.hpp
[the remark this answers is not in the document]
nmellado: Fixed

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -22,37 +22,44 @@ namespace Engine {
...
+                           const Ra::Core::Asset::GeometryData* data );
 
-    void initialize() override;
+    /*!
+     * Constructor from an existing mesh
+     * \warning Moves the mesh and takes its ownership
+     */
+    TriangleMeshComponent( const std::string& name,  Entity* entity,
+                           Core::Geometry::TriangleMesh&& mesh,
+                           Core::Asset::MaterialData* mat = nullptr );
```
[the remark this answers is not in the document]
nmellado: Yeah, I tried to keep the code generic.
But true, it Is confusing, to be honest I don’t know what is the most convenient.
We can have both by the way.

src/Core/Geometry/TriangleMesh.cpp
nmellado: Fixed.

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: Ping @hoshiryu  ?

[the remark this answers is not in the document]
hoshiryu: Yeah ok, but then we should do this everywhere we use pointers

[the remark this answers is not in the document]
nmellado: I agree.

src/Core/Utils/StdOptional.hpp
hoshiryu: This hasn't been resolved

src/Core/Geometry/MeshPrimitives.cpp
```diff
@@ -140,7 +141,7 @@ TriangleMesh makeGeodesicSphere( Scalar radius, uint numSubdiv ) {
 
             const Scalar x = 2.f * sq5_5 * std::cos( theta );
             const Scalar y = 2.f * sq5_5 * std::sin( theta );
-            const Scalar z = j == 0 ? sq5_5 : -sq5_5;
+            const Scalar z = Math::areApproxEqual( Scalar( j ), 0_ra ) ? sq5_5 : -sq5_5;
```
hoshiryu: I'm confused here, `j` is an `int`, so the code was right, why casting to `Scalar`?

src/Core/Utils/StdOptional.hpp
nmellado: @hoshiryu see: https://github.com/STORM-IRIT/Radium-Engine/pull/423/commits/6c7d2d63b6825e6a0969413ea2c85ec6c61e66f0#diff-de7544cb3809030281cea13d469fab66R9

hoshiryu: this must have been changed because when I open the file at the current commit, I still see `# pragma once`

nmellado: `#pragma once` offer simpler guards compatible on all plateforms.
We started to use them in Radium for some time, that's something that has already been discussed and accepted.
So I updated the documentation accordingly.

hoshiryu: ok, I wasn't aware of this discussion.

Plugins/CMakeLists.txt
```diff
@@ -4,6 +4,6 @@ set( CMAKE_DEBUG_POSTFIX "")
 
 add_subdirectory( Animation )
 add_subdirectory( Skinning )
-add_subdirectory( MeshFeatureTracking )
+# add_subdirectory( MeshFeatureTracking )
```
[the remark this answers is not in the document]
MathiasPaulin: What is the reason one might keep such a plugin ? What is its rôle ? Why to make it mandatory. @hoshiryu Either update the plugin documentation to tell why this is important or let's remove this unknown plugin.

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -215,22 +199,19 @@ void MeshFeatureTrackingComponent::setData( const Ra::Engine::Renderer::PickingR
         return;
     }
     auto ro = getRoMgr()->getRenderObject( m_pickedRoIdx );
-    auto rm = ro->getMesh()->getRenderMode();
-    if ( rm == MeshRenderMode::RM_POINTS && m_data.m_mode != PickingMode::VERTEX )
+    auto rm = ro->getMesh()->pickingRenderMode();
```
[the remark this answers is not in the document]
MathiasPaulin: resolved ?

Plugins/MeshPaint/src/MeshPaintComponent.cpp
```diff
@@ -48,10 +53,29 @@ void MeshPaintComponent::initialize() {
 
     auto ro = getRoMgr()->getRenderObject( *m_renderObjectReader() );
     m_baseConfig = ro->getRenderTechnique()->getConfiguration();
-    m_baseColors = ro->getMesh()->getData( Ra::Engine::Mesh::VERTEX_COLOR );
-    m_paintColors.resize( ro->getMesh()->getGeometry().vertices().size(),
-                          Ra::Core::Utils::Color::Skin() );
-    ro->getMesh()->addData( Ra::Engine::Mesh::VERTEX_COLOR, m_paintColors );
+
+    Mesh* mesh = dynamic_cast<Mesh*>( ro->getMesh().get() );
```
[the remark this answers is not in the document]
MathiasPaulin: Why ?

Plugins/MeshPaint/src/MeshPaintComponent.cpp
```diff
@@ -60,94 +84,115 @@ void MeshPaintComponent::startPaint( bool on ) {
     {
         ro->getRenderTechnique()->setConfiguration(
             Ra::Engine::ShaderConfigurationFactory::getConfiguration( "Plain" ) );
-        ro->getMesh()->addData( Ra::Engine::Mesh::VERTEX_COLOR, m_paintColors );
     } else
     {
+        Ra::Engine::Mesh* mesh = dynamic_cast<Ra::Engine::Mesh*>( ro->getMesh().get() );
```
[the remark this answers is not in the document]
MathiasPaulin: why ?

Plugins/MeshPaint/src/MeshPaintComponent.cpp
```diff
@@ -60,94 +84,115 @@ void MeshPaintComponent::startPaint( bool on ) {
...
+         ( picking.m_mode != Ra::Engine::Renderer::VERTEX &&
+           picking.m_mode != Ra::Engine::Renderer::C_VERTEX ) )
     {
-        // not supported yet -> might use same util functions as MeshFeatureTrackingComponent
         return;
     }
 
-    const auto& T = ro->getMesh()->getGeometry().m_triangles;
-    auto colors = ro->getMesh()->getData( Ra::Engine::Mesh::VERTEX_COLOR );
+    Mesh* mesh = dynamic_cast<Mesh*>( ro->getMesh().get() );
```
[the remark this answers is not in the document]
MathiasPaulin: why ?

Shaders/Materials/BlinnPhong/BlinnPhong.frag.glsl
MathiasPaulin: Because a bad BSDF interface design (blend between varying and uniform parameters in some BSDF related functions). To be corrected in V2.

Shaders/Materials/BlinnPhong/DepthAmbientBlinnPhong.frag.glsl
MathiasPaulin: Because a bad BSDF interface design (blend between varying and uniform parameters in some BSDF related functions). To be corrected in V2.

Shaders/Materials/BlinnPhong/LitOITBlinnPhong.frag.glsl
MathiasPaulin: Because a bad BSDF interface design (blend between varying and uniform parameters in some BSDF related functions). To be corrected in V2.

src/Core/Geometry/MeshPrimitives.cpp
MathiasPaulin: I'm really not an adept of over-typing. I propose to limit explicit typing (preferably with _ra) only where it is mandatory. Let the compiler do the job when it is able to, it does generally better than we. I agree, j is an int and integer comparison is exact. And from my point of view, this loop might be conveniently unrolled.

nmellado:
Well, I'm still trying to fix the first round of comments, so please do not add more and more.
Also, there are some proposals (like `_ra`) that have been discussed IRL, and now that are asked to be partially reverted. It takes time to do all these changes, so please do not change your mind every two days...

Plugins/CMakeLists.txt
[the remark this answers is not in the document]
hoshiryu: This plugin is an example of how one can use picking to retrieve mesh information.
Also, this plugin is very useful when debugging mesh processing / edition / skinning / other algorithms.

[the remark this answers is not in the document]
nmellado: I will revert that changes, as already said multiple time.
Just give me some time..

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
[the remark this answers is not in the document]
hoshiryu: not yet

src/Core/Geometry/MeshPrimitives.cpp
nmellado: There is no loop here.

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -215,22 +199,19 @@ void MeshFeatureTrackingComponent::setData( const Ra::Engine::Renderer::PickingR
...
+    if ( rm == Displayable::PKM_POINTS && m_data.m_mode != PickingMode::VERTEX )
     {
         m_data.m_mode = PickingMode::RO;
         return;
     }
-    if ( ( rm == MeshRenderMode::RM_LINES || rm == MeshRenderMode::RM_LINE_LOOP ||
-           rm == MeshRenderMode::RM_LINE_STRIP || rm == MeshRenderMode::RM_LINES_ADJACENCY ||
-           rm == MeshRenderMode::RM_LINE_STRIP_ADJACENCY ) &&
-         m_data.m_mode == PickingMode::TRIANGLE )
+    if ( rm == Displayable::PKM_TRI && m_data.m_mode == PickingMode::TRIANGLE )
```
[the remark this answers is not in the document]
nmellado: What about this ?
```c++
if ( rm != Displayable::NO_PICKING && m_data.m_mode == PickingMode::TRIANGLE )
```

Plugins/MeshPaint/src/MeshPaintComponent.cpp
```diff
@@ -60,94 +84,115 @@ void MeshPaintComponent::startPaint( bool on ) {
...
         return;
     }
 
-    if ( ( ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_LINES ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_LINE_LOOP ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_LINE_STRIP ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_LINES_ADJACENCY ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_LINE_STRIP_ADJACENCY ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_TRIANGLE_STRIP ||
-           ro->getMesh()->getRenderMode() == Ra::Engine::Mesh::RM_TRIANGLE_FAN ) )
```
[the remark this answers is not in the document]
nmellado: Because with this changes, this is the Mesh that takes care of computing the right render mode, see `Mesh::updatePickingRenderMode`. 
So the correct version should be:
```c++
    if ( pickingRenderMode == Ra::Engine::Displayable::NO_PICKING ||
         pickingRenderMode == Ra::Engine::Displayable::PKM_LINES ||   // not supported yet
         pickingRenderMode == Ra::Engine::Displayable::PKM_LINE_ADJ ) // not supported yet
    {
        return;
    }
    if ( pickingRenderMode == Ra::Engine::Displayable::PKM_POINTS &&
         ( picking.m_mode != Ra::Engine::Renderer::VERTEX &&
           picking.m_mode != Ra::Engine::Renderer::C_VERTEX ) )
    {
        return;
    }
```
Good for you ?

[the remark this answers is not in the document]
hoshiryu: This sounds good for lines and ponts, but what about `RM_TRIANGLE_STRIP` and `RM_TRIANGLE_FAN` modes? These are baked into `PKM_TRIANGLES`, we then have to also check these separately

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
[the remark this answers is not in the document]
hoshiryu: I would check `NO_PICKING` in the first test.
For `PKM_POINTS`, the configuration is checked right above.
At this point in the checking, we just don't want to pick triangles on line meshes, so I would stay on
```
if ( ( rm == Displayable::PKM_LINES || rm == Displayable::PKM_LINE_ADJ ) && m_data.m_mode == PickingMode::TRIANGLE )
```

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
nmellado: I suggest we stick to the current version, and if there is a need of the other version (using `Material` directly) we can add it.

src/Engine/RadiumEngine.hpp
```diff
@@ -89,8 +89,9 @@ class RA_ENGINE_API RadiumEngine {
      * @deprecated Will be removed from this class in the next release. A Mesh manager, that could
      * serve mesh by name will be implemented.
      */
-    [[deprecated]] Mesh* getMesh( const std::string& entityName, const std::string& componentName,
-                                  const std::string& roName = std::string() ) const;
+    [[deprecated]] Displayable* getMesh( const std::string& entityName,
+                                         const std::string& componentName,
+                                         const std::string& roName = std::string() ) const;
```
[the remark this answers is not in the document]
nmellado: I don't want to update it to avoid side-effect on older plugins, and anyway the method will be removed in v2.
Also, what's wrong with the comment ? It says it is deprecated and will be removed, right ?

src/Engine/Component/GeometryComponent.hpp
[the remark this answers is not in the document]
hoshiryu: ok for now, looking after radium-v2 for better material management

Plugins/MeshPaint/src/MeshPaintComponent.cpp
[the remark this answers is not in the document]
nmellado: Fixed.

nmellado:
Ok guys, I think I've fixed all your comments.
It's up to you now.

src/Engine/RadiumEngine.hpp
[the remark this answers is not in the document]
hoshiryu: ok

nmellado:
@hoshiryu painting is working for me.
But rosbif de the painted colors you need to enable per vertex color in the object material (using the material editor)

Applications/MainApplication/Gui/MaterialEditor.cpp
```diff
@@ -64,7 +67,7 @@ void MaterialEditor::onKsColorChanged( int ) {
     {
         Core::Utils::Color c( ksR->value() / 255.f, ksG->value() / 255.f, ksB->value() / 255.f,
```
MathiasPaulin: I saw several ".f" suffix in the code. Should they be "_ra" instead (or just nothing ?)

Applications/MainApplication/Gui/MaterialEditor.cpp
```diff
@@ -96,17 +99,14 @@ void MaterialEditor::newKsColor( const QColor& color ) {
     if ( m_renderObject && m_usable )
     {
         Core::Utils::Color c( color.redF(), color.greenF(), color.blueF(), 1.0 );
-        m_material->m_ks = c;
+        m_blinnphongmaterial->m_ks = c;
```
MathiasPaulin: in all these simple affectation, why not allowing the compiler to construct the object in place by doing `m_blinnphongmaterial->m_ks = { color.redF(), color.greenF(), color.blueF(), 1.0 };`

MathiasPaulin:
Just to be clear. For the _ra proposal, I do not change anything. I just prefer that explicit type definition is used only when mandatory. If type inference by the compiler, that has done great progress accross c++ version even for specific implementation such as CUDA (not used anyway in radium), is able to derive the good type, let's the compiler do its job. If there-is ambiguity or potential type checking errors (in {} initializer list e.g.) specify the type using `_ra` on litterals or `Scalar()` operator on others variable.

hoshiryu:
@nmellado checked the per-vertex coloring and painting on meshes works for me.
Can you add this to the documentation of both the material editor and the paint plugin?
Also, regarding point clouds, this doesn't change the fact that points are not picked, then not painted.

Applications/MainApplication/Gui/MaterialEditor.cpp
nmellado: Thanks. Updated.

Applications/MainApplication/Gui/MaterialEditor.cpp
nmellado: Thanks. Fixed.

nmellado:
@hoshiryu about picking on point-cloud, this is crashing for a long time, this PR does not fix the bug, but do not introduce it neither.

Applications/MainApplication/Gui/MaterialEditor.cpp
```diff
@@ -79,8 +81,8 @@ void MaterialEditor::newKdColor( const QColor& color ) {
 
     if ( m_renderObject && m_usable )
     {
-        Core::Utils::Color c( color.redF(), color.greenF(), color.blueF(), 1.0 );
-        m_material->m_kd = c;
+        m_blinnphongmaterial->m_kd = Core::Utils::Color (
+              color.redF(), color.greenF(), color.blueF(), 1. );
```
hoshiryu: `1_ra` ?

Applications/MainApplication/Gui/MaterialEditor.cpp
```diff
@@ -95,57 +97,59 @@ void MaterialEditor::newKsColor( const QColor& color ) {
 
     if ( m_renderObject && m_usable )
     {
-        Core::Utils::Color c( color.redF(), color.greenF(), color.blueF(), 1.0 );
-        m_material->m_ks = c;
+        m_blinnphongmaterial->m_ks = Core::Utils::Color (
+              color.redF(), color.greenF(), color.blueF(), 1. );
```
hoshiryu: `1_ra` ?

nmellado: No, `blueF` and others are double, so we need a double here to have all the parameters with the same type (see `ColorBase` constructor)

Applications/MainApplication/Gui/MaterialEditor.cpp
nmellado: No, `blueF` and others are double, so we need a double here to have all the parameters with the same type (see `ColorBase` constructor)

hoshiryu: Color works from `Scalar`s, so these should all be as such right?

nmellado: There is a constructor with a template parameter allowing to create a `ColorBase<Scalar>` from 4 values of another Scalar type, that are altogether converted to `Scalar` in the constructor.
As such, the 4 types given as input needs to be of the same type, otherwise there is a confusion for the compiler. This is done in purpose by design, to force the user to give consistent type and avoid unwanted implicit conversions.

Makes sense ?

hoshiryu: ok

nmellado:
I confirm the problem with picking on point-clouds is related to #425 .
There is currently a crash at runtime when point-clouds are rendered for picking using the shader designed for triangle meshes. Skipping this shader solves the crash, but do not fix the picking.

I'll investigate on fixing the picking shader for point clouds.

hoshiryu:
Actually, there is some direct setting of `Mesh::m_renderMode` in the code.
Calling `setRenderMode` instead also updates the picking mode , and thus solves the problem.

src/Core/Geometry/MeshPrimitives.cpp
```diff
@@ -113,13 +121,17 @@ TriangleMesh makeSharpBox( const Aabb& aabb ) {
...
+
     return result;
 }
 
-TriangleMesh makeGeodesicSphere( Scalar radius, uint numSubdiv ) {
+TriangleMesh makeGeodesicSphere( Scalar radius, uint numSubdiv,
+                                 const Utils::optional<Utils::Color>& color ) {
     TriangleMesh result;
-    uint faceCount = std::pow( 4, numSubdiv ) * 20;
+    uint faceCount = uint( std::pow( 4, numSubdiv ) ) * 20;
```
dlyr: This one do not generate warning on my  gcc version 8.3.0 (Debian 8.3.0-6) 
without a cast, but maybe needed by other compiler.
`uint faceCount = std::pow( 4, numSubdiv ) * 20;` compiles fine
(btw uint is not standard right ?)

nmellado: As `pow` returns a floating point value, we actually need a cast to avoid a warning for implicit conversion:
 - `warning: implicit conversion turns floating-point number into integer: 'double' to 'uint' (aka 'unsigned int')`
 - https://en.cppreference.com/w/cpp/numeric/math/pow

nmellado: And yes, uint is not standard, yet massively used in the code.
It is defined here: https://github.com/STORM-IRIT/Radium-Engine/blob/master-v1/src/Core/CoreMacros.hpp#L249

Shaders/Materials/BlinnPhong/BlinnPhong.frag.glsl
```diff
@@ -12,6 +11,9 @@ layout (location = 2) in vec3 in_normal;
 layout (location = 3) in vec3 in_tangent;
 layout (location = 4) in vec3 in_viewVector;
 layout (location = 5) in vec3 in_lightVector;
+layout (location = 6) in vec3 in_vertexcolor;
```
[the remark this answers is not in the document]
nmellado: done.

PR #427, opened 2019-06-03
Fix attribs

hoshiryu:

This PR fixes some issues with the management of mesh attributes.

- When removing an attribute from a mesh though its handle, we get the attribute through `getAttrib`, and then use it's name to look for it to check if it exists, which causes a crash when it does not.
- When building a `TopologicalMesh`, we copy all the mesh attributes onto halfedges, but without any check on the data size, thus causing a crash when the attribute is empty.
- Upon initialization, the `SkinningComponent` ask for a UV attribute within the mesh, but leave it empty if it does not exist yet, thus putting the mesh in an inconsistent state.

- When removing an attribute, we look for the attribute using its name within the handle.
- When copying the attributes, we first check the attribute has exactly one value per vertex in the input `TriangleMesh`, and output a warning message otherwise.
- The `SkinningComponent` now adds the UV attribute only when showing the skinning weights.
When disabling weights display, the attribute is reset to its initial state (if present from the beginning) or removed (otherwise).

* **Other information**:
- The method `getSize` from `Attrib<T>` previously returned the dimension of `T`. It now returns the number of elements in the data container, which seams more suited with its name.
- Fix Debug mode compilation.
- Prevent building of glm tests and static libraries.


Plugins/Skinning/src/SkinningComponent.cpp
```diff
@@ -429,17 +434,28 @@ void SkinningComponent::showWeights( bool on ) {
...
-        mesh->getAttrib( handle ).data() = m_baseUV;
+        handle = mesh->getAttribHandle<Ra::Core::Vector3>( attrUV );
+        // if the UV attrib existed before, reset it, otherwise remove it.
+        if ( m_baseUV.size() > 0 )
+        {
+            mesh->getAttrib( handle ).data() = m_baseUV;
+        } else
+        {
+            mesh->removeAttrib( handle );
+            m_meshWritter(); // FIXME: update Engine::Mesh's attrib handles
```
nmellado: check without

src/Core/Geometry/TopologicalMesh.cpp
```diff
@@ -23,10 +23,17 @@ void addAttribPairToTopo( const TriangleMesh& triMesh, TopologicalMesh* topoMesh
                           AttribManager::value_type attr, std::vector<PropPair<T>>& vprop,
                           std::vector<OpenMesh::HPropHandleT<T>>& pph ) {
     AttribHandle<T> h = triMesh.getAttribHandle<T>( attr->getName() );
-    OpenMesh::HPropHandleT<T> oh;
-    topoMesh->add_property( oh, attr->getName() );
-    vprop.push_back( std::make_pair( h, oh ) );
-    pph.push_back( oh );
+    if ( attr->getSize() == triMesh.vertices().size() )
+    {
+        OpenMesh::HPropHandleT<T> oh;
```
nmellado: add test

src/Core/Utils/Attribs.hpp
```diff
@@ -27,7 +27,7 @@ class AttribBase {
     void setName( std::string name ) { m_name = name; }
     virtual void resize( size_t s ) = 0;
 
-    virtual uint getSize() = 0;
+    virtual size_t getSize() = 0;
```
nmellado: add documentation

src/Core/Utils/Attribs.hpp
```diff
@@ -68,7 +68,7 @@ class Attrib : public AttribBase {
     inline const Container& data() const { return m_data; }
 
     virtual ~Attrib() { m_data.clear(); }
-    uint getSize() override { return Container::Matrix::RowsAtCompileTime; }
+    size_t getSize() override { return m_data.size(); }
     int getStride() override { return sizeof( typename Container::value_type ); }
```
nmellado: add warning: does not work for dynamic and sparse matrices

Plugins/Skinning/src/SkinningComponent.cpp
hoshiryu: works fine without

nmellado: Great, please remove the comment then.

src/Core/Utils/Attribs.hpp
dlyr: well this one is to be used along with stride to send data to the gpu. 
Actually it is not used yet ... (size is getted from VecArray::Vector::RowsAtCompileTime with the sendGL method of Engine::Mesh).

Nevertheless we could imagine having both meth (get size : number of attributes and get Whatever : dim of each value).

But this function is clearly related to gpu stuff, and size should always be the number of vertices in the mesh (so user don't have to worry about one attrib specific size).

Plugins/Skinning/src/SkinningComponent.cpp
hoshiryu: Actually need it.

src/Core/Utils/Attribs.hpp
```diff
@@ -19,58 +19,109 @@ namespace Utils {
...
     virtual void resize( size_t s ) = 0;
 
-    virtual uint getSize() = 0;
+    /**
+     * Return the number of elements in the attribute content.
+     */
+    virtual size_t getSize() = 0;
+
+    /**
+     * Return the size, in bytes, of one attribute element.
```
dlyr: Well stide is the offset, in byte, to go from one attirbute address to ne next one.
As defined here https://www.khronos.org/opengl/wiki/Vertex_Specification

MathiasPaulin:
```suggestion
     * Return the stride, in bytes, from one attribute address to the next one.
```

PR #429, opened 2019-06-04
Fix to convert int to TaskId.

dlyr:
Hot fix, no doc ;)

Fix

Radium do not compile on my setup.
Apple LLVM version 10.0.0 (clang-1000.11.45.5)

It compiles (at least on my setup)

* **Other information**:


MathiasPaulin:
@dlyr Why closing this PR? The problem is real and not limited to clang. gcc has the same. But, as it appears only when compiling in DEBUG mode, we need to fix it quickly (perhaps with more un-tested debug mode effects.

hoshiryu:
For your info, I pushed a fix in the last commit of #427

MathiasPaulin:
@hoshiryu OK, thanks

PR #430, opened 2019-06-04
Core/Utils/Color "linear RGB" and "sRGB" conversion

MathiasPaulin:
- [X] Docs have been added / updated (for bug fixes / features)

This PR add two operators on Core::Utils::ColorBase to convert a color between sRGB and linear RGB color space.
This is then used, in the mainapp, to make the background color selection user friendly as, after selection, the renderer background will match exactly the chosen color.
 
The color chooser for the background does not show the real background color.
The background color chooser is not initialized from the actual background.

The background color chooser is initialized with the actual background and the colors matches.

* **Other information**:
There is still some strange behavior with the QSettings (at least on MacOs) that prevents to edit the QSettings from another app.

### DIFF
--- a/src/Core/Utils/Color.hpp
+++ b/src/Core/Utils/Color.hpp
@@ -6,16 +6,17 @@
 #include <Eigen/Geometry> //homogeneous
 #include <random>
 
+
+namespace Ra {
+namespace Core {
+namespace Utils {
+
 /**
  * gamma value for sRGB color space transfer function
  * @see https://en.wikipedia.org/wiki/SRGB
  * @see http://www.color.org/srgb.pdf
  */
-static constexpr Scalar SRGB_GAMMA=2.4_ra;
-
-namespace Ra {
-namespace Core {
-namespace Utils {
+  static constexpr Scalar SRGB_GAMMA=2.4_ra;
 
 /*!
  * Colors are defined as vector4, i.e. 4 Scalars in RGBA order.
--- a/src/Core/Utils/Color.hpp
+++ b/src/Core/Utils/Color.hpp
@@ -11,7 +11,7 @@
  * @see https://en.wikipedia.org/wiki/SRGB
  * @see http://www.color.org/srgb.pdf
  */
-#define SRGB_GAMMA 2.4_ra
+static constexpr Scalar SRGB_GAMMA=2.4_ra;
 
 namespace Ra {
 namespace Core {
@@ -45,25 +45,29 @@ class ColorBase : public Eigen::Matrix<_Scalar, 4, 1> {
     operator VectorType() { return *this; }
 
     /// convert the color expressed in sRGB color space to linear RGB
-    inline void toLinRGB( _Scalar gamma = SRGB_GAMMA) {
-        for (auto &u : rgb()) {
+    inline ColorBase toLinRGB( _Scalar gamma = SRGB_GAMMA) const {
+        ColorBase<_Scalar> c( *this );
+        for (auto &u : c.rgb()) {
             if (u < 0.04045_ra) {
                 u /=12.92_ra;
             } else {
                 u = std::pow((u+0.055_ra)/1.055_ra, gamma);
             }
         }
+        return c;
     }
 
     /// convert the color expressed in linear RGB color space to sRGB
-    inline void tosRGB( _Scalar gamma = SRGB_GAMMA) {
-        for (auto &u : rgb()) {
+    inline ColorBase tosRGB( _Scalar gamma = SRGB_GAMMA) const {
+        ColorBase<_Scalar> c( *this );
+        for (auto &u : c.rgb()) {
             if (u < 0.0031308_ra) {
                 u *=12.92_ra;
             } else {
                 u = 1.055_ra*std::pow(u, 1_ra/gamma)-0.055_ra;
             }
         }
+        return c;
     }
 
     template <typename Derived>
--- a/src/Core/Utils/Color.hpp
+++ b/src/Core/Utils/Color.hpp
@@ -11,13 +11,6 @@ namespace Ra {
 namespace Core {
 namespace Utils {
 
-/**
- * gamma value for sRGB color space transfer function
- * @see https://en.wikipedia.org/wiki/SRGB
- * @see http://www.color.org/srgb.pdf
- */
-  static constexpr Scalar SRGB_GAMMA=2.4_ra;
-
 /*!
  * Colors are defined as vector4, i.e. 4 Scalars in RGBA order.
  * displayable colors should have all their coordinates between 0 and 1.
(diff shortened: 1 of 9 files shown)


src/Core/Utils/Color.hpp
```diff
@@ -37,6 +44,28 @@ class ColorBase : public Eigen::Matrix<_Scalar, 4, 1> {
     /// cast operator, mandatory to use Vector arithmetic
     operator VectorType() { return *this; }
 
+    /// convert the color expressed in sRGB color space to linear RGB
+    inline void toLinRGB( _Scalar gamma = SRGB_GAMMA) {
```
nmellado: As the method starts by `to`, I would note make it in place, and return a new instance.
Then it sounds more readable when using it:
```c++
auto sColor =  m_viewer->getBackgroundColor().tosRGB();
```
Of course you can also add `toLinRGBInPlace`, but the least we can do is to offer the aforementioned version.

src/Core/Utils/Color.hpp
```diff
@@ -6,6 +6,13 @@
 #include <Eigen/Geometry> //homogeneous
 #include <random>
 
+/**
+ * gamma value for sRGB color space transfer function
+ * @see https://en.wikipedia.org/wiki/SRGB
+ * @see http://www.color.org/srgb.pdf
+ */
+#define SRGB_GAMMA 2.4_ra
```
nmellado: Please use a `static constexpr` function in Color classe instead of a macro.

src/Engine/Renderer/Texture/Texture.hpp
```diff
@@ -237,8 +237,9 @@ class RA_ENGINE_API Texture final {
      * @param gamma the gama value to use (sRGB is 2.4)
      * @note only 8 bit (GL_UNSIGNED_BYTE data format) textures are managed by this operator.
      */
+    // FIXME : replace the 2.4 constant by SRGB_GAMMA without requiring to include Eigen ...
     void sRGBToLinearRGB( uint8_t* texels, uint numCommponent, bool hasAlphaChannel,
-                          Scalar gamma = Scalar( 2.4 ) );
+                          Scalar gamma = 2.4_ra );
```
nmellado: Use value defined in Color.

MathiasPaulin: As mentioned in the PR summary, this will add a dependency on Texture over Eigen and then, increase the compilation time. If you think it is the better solution, let's go for that

MathiasPaulin:
@nmellado All requested changes done ...

dlyr:
Well, I think the gamma should be 2.2, while 2.4 still be used in computation. 
(as explained here https://en.wikipedia.org/wiki/SRGB  )
I would prefer static member function (as the other converters), 
srgbToLinear
LinearToSrgb

MathiasPaulin:
> Well, I think the gamma should be 2.2, while 2.4 still be used in computation.
> (as explained here https://en.wikipedia.org/wiki/SRGB )
Well, but how to define 2.4 an other way than 2.2+0.2 that is arbitrary and do not ensure consistancy ?

> I would prefer static member function (as the other converters),
> srgbToLinear
> LinearToSrgb
3 developpers, 3 point of view. to be discussed then ...

MathiasPaulin:
@dlyr @nmellado could you review the new version ...

src/Core/Utils/Color.hpp
```diff
@@ -37,6 +38,32 @@ class ColorBase : public Eigen::Matrix<_Scalar, 4, 1> {
     /// cast operator, mandatory to use Vector arithmetic
     operator VectorType() { return *this; }
 
+    /// convert the color expressed in sRGB color space to linear RGB
+    static inline ColorBase toLinRGB(const ColorBase &srgb) {
```
nmellado:
```suggestion
    static inline ColorBase sRGBToLinear(const ColorBase &srgb) {
```

src/Core/Utils/Color.hpp
```diff
@@ -37,6 +38,32 @@ class ColorBase : public Eigen::Matrix<_Scalar, 4, 1> {
...
+                u /=12.92_ra;
+            } else {
+                u = std::pow((u+0.055_ra)/1.055_ra, 2.4_ra);
+            }
+        }
+        return c;
+    }
+
+    /// convert the color expressed in linear RGB color space to sRGB
+    static inline ColorBase tosRGB(const ColorBase &lrgb) {
```
nmellado:
```suggestion
    static inline ColorBase linearTosRGB(const ColorBase &lrgb) {
```

src/Engine/Renderer/Texture/Texture.hpp
```diff
@@ -186,9 +188,9 @@ class RA_ENGINE_API Texture final {
      * This will transform the internal representation of the texture to GL_FLOAT.
      * Only GL_RGB[8, 16, 16F, 32F] and GL_RGBA[8, 16, 16F, 32F] are managed.
      * Full transformation as described at https://en.wikipedia.org/wiki/SRGB
-     * @param gamma the gama value to use (sRGB is 2.4)
+     * The gama value  used is the sRGB one  2.4
```
dlyr: 2.2 ;)

nmellado:
@dlyr about handling multiple color space, I think we might consider using the library we developed for Constrained Color Palette Exploration.
This could require a refactoring for V2.

src/Engine/Renderer/Texture/Texture.hpp
MathiasPaulin: from sRGB wikipedia page : 
> Unlike most other RGB color spaces, the sRGB gamma cannot be expressed as a single numerical value. The overall gamma is approximately 2.2, consisting of a linear (gamma 1.0) section near black, and a non-linear section elsewhere involving a 2.4 exponent and a gamma (slope of log output versus log input) changing from 1.0 through about 2.3.

So, the gamma value 2.4 is for the non linear part, the gamma value 1 is the value of the lienear part

Concl. thereis no parameter and this comment will be updated to remove all numeric precision and to give only the link.

src/Core/Utils/Color.hpp
MathiasPaulin: the proposal of @dlyr for the naming is better. I'll integrate ASAP

MathiasPaulin:
> Again gamma for sRGB is assumed to be 2.2, even if 2.4 is used in the computation (but it is not the gamma).

Then exact sentence is (cf my previous post and https://en.wikipedia.org/wiki/SRGB
> Unlike most other RGB color spaces, the sRGB gamma cannot be expressed as a single numerical value. The overall gamma is approximately 2.2, consisting of a linear (gamma 1.0) section near black, and a non-linear section elsewhere involving a 2.4 exponent

src/Core/Utils/Color.hpp
MathiasPaulin:
```suggestion
    static inline ColorBase sRGBToLinearRGB(const ColorBase &srgb) {
```

MathiasPaulin:
> Our shader can keep gamma as a parameters (it is independent of the sRGB one, and could be adjusted depending on hardware ...)

If we use something else than 2.4, all the conversion function must be changed.  cf section Theory of the transformation of the wikipedia page
>Solving with {\displaystyle \gamma =2.4} \gamma =2.4 and ...
So, we do not allow to change the gamma, all the shader must be changed to adapt to the hardware.

MathiasPaulin:
> Reviews seem to have converged to an accept, right ?
Right :)

PR #436, opened 2019-06-14
Loop

dlyr:

Small fix + 1 change on event management.

`main()` function explicitly manage `processEvent` (but not it minimalApp by the way)

Some small fix (clang-tidy)
One fix for Qt 5.12, since the WindowQt is actually used as a Widget, the call to `create` in ctor just mix up context.

One other change is to use the usual `app.exec()` to manage`processEvent`.
I have added two boolean to `BaseApplication` one manage the update, one the continuous update.

To trigger an update, any widget, component or system can call `askForUpdate`, then next time the update event is processed, if asked for, `radiumFrame` is trigger.

* **Other information**:
Comments and testing welcome


src/Engine/Renderer/Mesh/Mesh.hpp
```diff
@@ -169,7 +169,7 @@ class RA_ENGINE_API Mesh : public Displayable
     /// Update the picking render mode according to the object render mode
     void updatePickingRenderMode();
 
-  private:
+  protected:
```
MathiasPaulin: According to cpp guidelines C.133 (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c133-avoid-protected-data).
> protected data is a source of complexity and errors. protected data complicates the statement of invariants. protected data inherently violates the guidance against putting data in base classes, which usually leads to having to deal with virtual inheritance as well. 

They make difficult to ensure invariant on the objects each derived class must ensure the correct use and update of such members. Either let this private with a protected accessor (in charge of verifying eventual invariant) or stricly limit the protected data section (here, you pour everything below m_vao as protected, Is this intentional and mandatory ?)

src/GuiBase/BaseApplication.cpp
```diff
@@ -356,77 +362,75 @@ void BaseApplication::addBasicShaders() {
...
-        emit( updateFrameStats( m_timerData ) );
-        m_timerData.clear();
+        if ( m_frameCounter % m_frameCountBeforeUpdate == 0 )
+        {
+            emit( updateFrameStats( m_timerData ) );
+            m_timerData.clear();
+        }
     }
-
     m_mainWindow->onFrameComplete();
```
MathiasPaulin: Are you sure this must be called outside of the frame update block (the if above) ?

src/GuiBase/BaseApplication.hpp
```diff
@@ -68,7 +69,7 @@ class RA_GUIBASE_API BaseApplication : public QApplication
                      const WindowFactory& factory,
                      QString applicationName  = "RadiumEngine",
                      QString organizationName = "STORM-IRIT" );
-    virtual ~BaseApplication();
+    ~BaseApplication() override;
```
MathiasPaulin: According to Cpp guidelines C.128: (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rh-override)
> If a base class destructor is declared virtual, one should avoid declaring derived class destructors virtual or override.
This is because destructors of a derived class does not override of the base destructor but is used in a chain call.

src/GuiBase/BaseApplication.hpp
```diff
@@ -175,6 +191,12 @@ class RA_GUIBASE_API BaseApplication : public QApplication
     bool m_recordGraph;
```
MathiasPaulin: When editing files that lacks doc and inline intialization, just add what is simple and quick.

src/GuiBase/BaseApplication.cpp
dlyr: Oops my comments were not taken into account when I click resolve conversation. 
So here again : well for this one I don't know but I remark I double check to update, so I simply remove the test here.

src/GuiBase/BaseApplication.hpp
dlyr: That was a clang tidy suggestion, I revert back.

src/GuiBase/BaseApplication.hpp
dlyr: Don't know what this bool is for, help appreciated.

src/Engine/Renderer/Mesh/Mesh.hpp
dlyr: Well it was a mistake to make all the above protected. 
Nevertheless giving access directly or even thru a protected accessor allows to mess thing up with OpenGL, I keep this modification for myself.

src/GuiBase/BaseApplication.hpp
MathiasPaulin: Done. I' commit the comment.

dlyr:
Is that ok to merge ?

Applications/MainApplication/Gui/MainWindow.cpp
```diff
@@ -473,6 +474,7 @@ void Gui::MainWindow::editRO() {
     {
         m_materialEditor->changeRenderObject( item.m_roIndex );
         m_materialEditor->show();
+        // \fixme check here autoupdate
```
MathiasPaulin: I'ts the job of the m_materialEditor to ask for update if needed. There is nothing to add here.

nmellado:
@MathiasPaulin I have addressed your comments about code style
@dlyr I've changed the autoupdate management to handle multiple requests from different components.

MathiasPaulin:
@nmellado you have to approve the changes made after your change request.

nmellado:
@hoshiryu could you please test this PR and check their is no side effect on your side, or further changes to make ? As the rendering updates are not made all the time now, you might need to trigger them in your plugins (I have made some changes, but I have probably missed some others).
For instance, I don’t know how to check it the scalar field visualization needs to trigger the updates.
Thanks !

src/PluginBase/RadiumPluginInterface.hpp
```diff
@@ -34,16 +27,8 @@ class FileLoaderInterface;
...
-    Engine::RadiumEngine* m_engine{nullptr};
-    GuiBase::SelectionManager* m_selectionManager{nullptr};
-    Gui::PickingManager* m_pickingManager{nullptr};
-    Gui::Viewer* m_viewer{nullptr};
-    std::string m_exportDir{"."};
-};
-
 namespace Plugins {
+
 class RadiumPluginInterface
```
hoshiryu: Don't we need `RA_PLUGINBASE_API` here?

src/PluginBase/PluginContext.hpp
```diff
@@ -0,0 +1,64 @@
...
+     * {
+     *   void MyPlugin::registerPlugin( const Ra::Plugins::Context& context ) {
+     *     connect( this, &MyPlugin::askForUpdate, &context, &Ra::Plugins::Context::askForUpdate );
+     *     connect( this, &MyPlugin::setContinuousUpdate, &context,
+     * &Ra::Plugins::Context::setContinuousUpdate );
+     *   }
+     *
+     * signals:
+     *   void setContinuousUpdate( bool b );
+     *   void askForUpdate();
```
hoshiryu: Shall we have these in the `RadiumPluginInterface` ?

nmellado: No, you cannot have signal and slots in an interface, because you need MOC to parse the class.

nmellado:
@hoshiryu regarding your comment:
> The MeshPaint plugin has issues: the material colors of the mesh are not automatically set to per-vertex, when disabling color painting and reactivating it, the mesh is all black.

No they are not. This behavior has been introduced when adding the per-vertex option.
Also, if this PR does not introduce a change of behavior, this comment is out of the scope of this review. Please open an issue if you believe thereis something wrong.

nmellado:
Closing. See PR #439 .

PR #438, opened 2019-06-18
Fix paintplugin

### LINKED ISSUE #437: Plugin MeshPaint is buggy
The MeshPaint plugin has some behavior issues.
When disabling color painting and reactivating it, the mesh is all black with `per-vertex` diffuse color material.
This might be due to missing attribute removal on painting disabling.

hoshiryu:

Fix #437 

When de-activating mesh paint and re-activating it on a mesh with no previous `per-vertex` colors, the mesh ends up all black and painting colors does not work.
Actually this is due to the fact that, when de-activating colors painting, the colors are set back to what they were before painting, which in this case is `none`.
Thus when re-activating painting, since the mesh has the color attribute, the plugin just copies it and no validity check is done.
Hence when painting the colors, colors are set in an empty container (this should crash but somehow does not).

- On color painting activation, the plugin request the addition of a color attribute to the mesh.
If there is only one, this one is selected as the `to be painted` and previous colors are saved.
Otherwise, a new one is created and filled with a default color.
- On color painting de-activation, the colors are set to what they were beforehand, i.e. either the saved ones or none.
Note that removing the created attribute would be preferred, but the actual management of attributes doesn't clear the data on GPU side.

* **Other information**:

### DIFF
--- a/Plugins/MeshPaint/src/MeshPaintComponent.cpp
+++ b/Plugins/MeshPaint/src/MeshPaintComponent.cpp
@@ -72,18 +72,12 @@ void MeshPaintComponent::startPaint( bool on ) {
 
         // If any, save colors from the CPU object, otherwise set default color and notify GPU
         // object
-        m_currentColorAttribHdl = triangleMesh.getAttribHandle<Ra::Core::Vector4>( colAttribName );
-        if ( m_currentColorAttribHdl.idx().isInvalid() )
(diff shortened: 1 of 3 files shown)


nmellado:
If I understand well, to get a proper fix here, we need a mechanism to update the data on the GPU when an attribute handle is deleted in `TriangleMesh`, right ?

hoshiryu:
Yes, that would be appreciated

hoshiryu:
fixed by #451

PR #440, opened 2019-06-19
Refactor KeyMappingManager.

dlyr:
Simpler conf file, simpler call.
Action are sent to camera and gizmoManager instead of letting them call the keybindingManager

THIS PR is to start to discuss the design of a new keyMappingManager, along with some simplification about interaction.

I think to add a context to keyMapping (so you can call getAction("CameraContext" ...)
And the context will be determine by the viewer according to either some state or what is under the mouse (with a pickingQuery directly from the mouse press event).

Comments are welcom, then I will create the real PR.

### DIFF
--- a/Configs/default.xml
+++ b/Configs/default.xml
@@ -1,56 +1,22 @@
 <?xml version="1.0" encoding="ISO-8859-1" ?>
 
+<!--
+<keymap key="" modifiers="" buttons="" action="" />
+-->
+
 <keymaps>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="mouse" value="MiddleButton"></information>
-        <information id="action" value="TRACKBALLCAMERA_MANIPULATION"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_A"></information>
-        <information id="action" value="TRACKBALLCAMERA_ROTATE_AROUND"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="mouse" value="LeftButton"></information>
-        <information id="action" value="GIZMOMANAGER_MANIPULATION"></information>
-    </keymap>
-   <keymap>
-        <information id="type" modifier="ControlModifier" type="mouse" value="LeftButton"></information>
-        <information id="action" value="GIZMOMANAGER_STEP"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="mouse" value="RightButton"></information>
-        <information id="action" value="VIEWER_BUTTON_PICKING_QUERY"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_V"></information>
-        <information id="action" value="FEATUREPICKING_VERTEX"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_E"></information>
-        <information id="action" value="FEATUREPICKING_EDGE"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_T"></information>
-        <information id="action" value="FEATUREPICKING_TRIANGLE"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_C"></information>
-        <information id="action" value="FEATUREPICKING_MULTI_CIRCLE"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="AltModifier" type="mouse" value="LeftButton"></information>
-        <information id="action" value="VIEWER_BUTTON_CAST_RAY_QUERY"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_R"></information>
-        <information id="action" value="VIEWER_RAYCAST_QUERY"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_Z"></information>
-        <information id="action" value="VIEWER_TOGGLE_WIREFRAME"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="ShiftModifier" type="mouse" value="LeftButton"></information>
-        <information id="action" value="COLORWIDGET_PRESSBUTTON"></information>
-    </keymap>
+    <keymap key="" modifiers="" buttons="MiddleButton" action="TRACKBALLCAMERA_MANIPULATION"/>
+    <keymap key="" modifiers="" buttons="MiddleButton" action="TRACKBALLCAMERA_ROTATE"/>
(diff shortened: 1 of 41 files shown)


hoshiryu:
It would be nice to allow plugins to add their own shortcuts.
But then we also might need to deal with multiple bindings, and here the context makes sense.

dlyr:
This PR is not ready yet, but review are welcome.

I have somehow broken something with gizmo.
For instance the first time I translate is fine, but if I change translate plane, the edit is weird.
I may avec remove some kind of update or so ... So if you have any idea where it can come from so I can fix it quickly, I will appreciate ;)

Others comments on design are welcome, as refactoring or help.

MathiasPaulin:
I'll give a look. 
I suppose that your PR #442 fix your gizmo problem ?

dlyr:
Nope :( #442 solve the gizmo flat looking bug bit not the weird interaction problem. 
I miss something on the actual gizmo pipeline in master. 
If someone can tell what are the mandatory step to have a gizmo working like :
Call select constraints one because it setup the start bool. 
Call mouse press 
... with a description of the member variables and side effects of gizmo manager and say translation gizmo 
Then i think i could adapt the PR.

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
```diff
@@ -1,22 +1,50 @@
...
+    m_keyMappingContext = Gui::KeyMappingManager::getInstance()->getContext( "GizmoContext" );
+    if ( m_keyMappingContext.isInvalid() )
+    {
+        LOG( Ra::Core::Utils::logINFO )
+            << "GizmoContext not defined (maybe the configuration file do not contains it";
+    }
+#define KMA_VALUE( XX ) \
+    XX = Gui::KeyMappingManager::getInstance()->getActionIndex( m_keyMappingContext, #XX );
+    KeyMappingGizmo
+#undef KMA_VALUE
```
hoshiryu: In case the context is invalid we should return and not try to access it, otherwise this will crash

src/GuiBase/Viewer/TrackballCamera.cpp
```diff
@@ -13,8 +14,28 @@
...
+    if ( m_keyMappingContext.isInvalid() )
+    {
+        LOG( Ra::Core::Utils::logINFO )
+            << "ViewerContext not defined (maybe the configuration file do not contains it";
+    }
+
+#define KMA_VALUE( XX ) \
+    XX = Gui::KeyMappingManager::getInstance()->getActionIndex( m_keyMappingContext, #XX );
+    KeyMappingCamera
+#undef KMA_VALUE
```
hoshiryu: In case the context is invalid we should return and not try to access it, otherwise this will crash.
Also, please update the log message from `ViewerContext` to `CameraContext`

src/GuiBase/Viewer/Viewer.cpp
```diff
@@ -55,6 +55,24 @@ namespace Ra {
...
+    if ( m_keyMappingContext.isInvalid() )
+    {
+        LOG( logINFO )
+            << "ViewerContext not defined (maybe the configuration file do not contains it";
+    }
+
+#define KMA_VALUE( XX ) \
+    XX = Gui::KeyMappingManager::getInstance()->getActionIndex( m_keyMappingContext, #XX );
+    KeyMappingViewer
+#undef KMA_VALUE
```
hoshiryu: In case the context is invalid we should return and not try to access it, otherwise this will crash

Configs/default.xml
```diff
@@ -1,56 +1,26 @@
...
-    <keymap>
-        <information id="type" modifier="NoModifier" type="key" value="Key_Z"></information>
-        <information id="action" value="VIEWER_TOGGLE_WIREFRAME"></information>
-    </keymap>
-    <keymap>
-        <information id="type" modifier="ShiftModifier" type="mouse" value="LeftButton"></information>
-        <information id="action" value="COLORWIDGET_PRESSBUTTON"></information>
-    </keymap>
+    <keymap context="CameraContext" key="" modifiers="" buttons="MiddleButton" action="TRACKBALLCAMERA_MANIPULATION"/>
+    <keymap context="CameraContext" key="" modifiers="" buttons="MiddleButton" action="TRACKBALLCAMERA_ROTATE"/>
```
hoshiryu: I missed than one: why having `TRACKBALLCAMERA_MANIPULATION` ?

src/GuiBase/Utils/KeyMappingManager.cpp
```diff
@@ -7,73 +7,113 @@ namespace Gui {
...
+            std::end( m_actionNameToIndex[contextIndex] ),
+            [&]( const ActionNameMap::value_type& pair ) { return f->second == actionIndex; } );
+
+        LOG( logWARNING ) << "Binding action " << findResult->first << " to "
+                          << "buttons [" << enumNamesFromMouseButtons( binding.m_buttons ) << "] "
+                          << "modifiers [" << enumNamesFromKeyboardModifiers( binding.m_modifiers )
+                          << "] "
+                          << "keycode [" << binding.m_key << "]"
+                          << "wheel [" << binding.m_wheel << "]"
+                          << ", which is already used for action " << findResult2->first << ".";
```
hoshiryu: Please check the validity of `findResult2` before using it

hoshiryu: Please also check the validity of `findResult2` before using it

hoshiryu:
For the gizmo latency, this is due to the fact that the tranform of the gizmo RO is updated only after the frame has been rendered (cf `MainWindow::onFrameComplete`).
When updating the gizmo transform in `Viewer::mouseMoveEvent`, right after calling `GizmoManager::handleMouseMoveEvent`, all goes right.

src/GuiBase/Utils/KeyMappingManager.cpp
dlyr: Done (btw this code was erroneous ... some bad copy/paste).

src/GuiBase/Viewer/Gizmo/GizmoManager.cpp
```diff
@@ -1,22 +1,52 @@
...
+}
+
+bool GizmoManager::isValidAction( const Gui::KeyMappingManager::KeyMappingAction& action ) const {
+
+    bool res = false;
```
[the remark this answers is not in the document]
dlyr: that's not true, see the macro codacy-bot !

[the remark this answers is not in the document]
MathiasPaulin: I think that cpp-check has problem with the macro expansion.
In the code, `res` is used in a macro that is not directly instantiated but only indirectly instantiated through the call of the macro `KeyMappingGizmo`.
We can forget this.

dlyr:
Are we ok to merge ?

nmellado:
@hoshiryu could you check the changes you requested have been done ?

nmellado:
@dlyr do you want to rewrite your commit history for a rebase, or do we stash this PR ?

PR #444, opened 2019-07-01
Fix #443 

### LINKED ISSUE #443: Recording during n frames
I would like Radium to open a PLY file and a camera file, to record one frame, and then to close.

I tried `./main-app -f mesh.ply -c camera.cam -n 1`, but no frame is recorded and Radium does not close.

Is there something wrong, or did I misunderstand the option `-n`?

I quickly checked `BaseApplication.cpp` and saw that 
- the option `-n`  does not set `m_recordFrames` to `true`
- when the number of rendered frames reaches the number of frame given by `-n`, the method `appNeedsToQuit()` is called but Radium continues to run

hoshiryu:

Fix #443 
Fix gizmo update on animation play / reset.

Due to the rework of the app loop, the app do not close after having run the requested number of frame through the option `-n`/`-numframes`.
There is no option for recording frames.
When resetting the animation skeleton pose, the gizmos do not update, neither do they when playing the animation.

After running the update for the current frame, the application checks if it needs to quit, and does so in this case.
Add an option for recording frames from launch (unfortunately, there is no easy way for now to update the MainApplication's UI at start, so the button stays unchecked while we are actually recording).
Also add a button to take a snapshot (i.e. recording only the current display).
Gizmos are updated right before rendering.

* **Other information**:


dlyr:
I have created a PR on your branch to correct one segfault I have on quit.
Not sure it's the best way to handle it, but I do not find the deletion path to ensure ItemManager is valid when ItemEntry is deleted.

nmellado:
What’s the status regarding @dlyr’s comment ?

hoshiryu:
The PR is ready for reviews

PR #445, opened 2019-07-09
Aabb

dlyr:
This PR is still WIP, need testing, and cleanup.

Change the computation of aabb, so that derived Entity, Component or RenderObject could have specific computation.

The aabb is computed by RenderObjectManager, querying the render object geometry

The object hierarchy is used to ask for aabb, with virtual function. 

nope

* **Other information**:
Also provide some code style correction and KeyMappingManager guards.

### DIFF
--- a/Applications/MainApplication/Gui/MainWindow.cpp
+++ b/Applications/MainApplication/Gui/MainWindow.cpp
@@ -623,7 +623,7 @@ void MainWindow::resetScene() {
 }
 
 void MainWindow::fitCamera() {
-    auto aabb = Engine::RadiumEngine::getInstance()->getRenderObjectManager()->getSceneAabb();
+    auto aabb = Engine::RadiumEngine::getInstance()->computeSceneAabb();
     if ( aabb.isEmpty() )
     {
         m_viewer->getCameraInterface()->resetCamera();
--- a/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
+++ b/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
@@ -584,7 +584,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
     // manage picking mode
     auto ro = getRoMgr()->getRenderObject( m_pickedRoIdx );
     if ( m_pickedMesh->getRenderMode() == MeshRenderMode::RM_POINTS )
-    { return ro->getAabb().sizes().norm() / 500; }
+    { return ro->computeAabb().sizes().norm() / 500_ra; }
     const auto& v = m_pickedMesh->getTriangleMesh().vertices();
     switch ( m_data.m_mode )
     {
@@ -622,7 +622,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
                2.0;
     }
     default:
-        return ro->getAabb().diagonal().norm() / 100_ra;
+        return ro->computeAabb().diagonal().norm() / 100_ra;
     }
 }
 
--- a/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
+++ b/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
@@ -14,6 +14,8 @@
 #include <Engine/Renderer/RenderObject/RenderObject.hpp>
 #include <Engine/Renderer/RenderObject/RenderObjectManager.hpp>
 
+#define POINT_SCALE_FACTOR 1_ra / 500
+
 using Ra::Engine::ComponentMessenger;
 using MeshRenderMode = Ra::Engine::Mesh::MeshRenderMode;
 using PickingMode    = Ra::Engine::Renderer::PickingMode;
@@ -584,7 +586,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
     // manage picking mode
     auto ro = getRoMgr()->getRenderObject( m_pickedRoIdx );
     if ( m_pickedMesh->getRenderMode() == MeshRenderMode::RM_POINTS )
-    { return ro->computeAabb().sizes().norm() / 500_ra; }
+    { return ro->computeAabb().sizes().norm() * POINT_SCALE_FACTOR; }
     const auto& v = m_pickedMesh->getTriangleMesh().vertices();
     switch ( m_data.m_mode )
     {
@@ -622,7 +624,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
                2.0;
     }
     default:
-        return ro->computeAabb().diagonal().norm() / 100_ra;
+        return 1_ra;
     }
 }
 
--- a/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
+++ b/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
@@ -14,7 +14,7 @@
 #include <Engine/Renderer/RenderObject/RenderObject.hpp>
 #include <Engine/Renderer/RenderObject/RenderObjectManager.hpp>
 
-#define POINT_SCALE_FACTOR 1_ra / 500
+constexpr Scalar POINT_SCALE_FACTOR = 1_ra / 500;
 
 using Ra::Engine::ComponentMessenger;
 using MeshRenderMode = Ra::Engine::Mesh::MeshRenderMode;
--- a/src/Engine/Component/Component.cpp
+++ b/src/Engine/Component/Component.cpp
@@ -54,5 +54,17 @@ void Component::notifyRenderObjectExpired( const Core::Utils::Index& idx ) {
     if ( found != m_renderObjects.cend() ) { m_renderObjects.erase( found ); }
 }
 
+Core::Aabb Component::computeAabb() const {
+
+    // component have not transformation, so ask render objects own bbox.
+    Core::Aabb aabb;
+    for ( const auto& roIndex : m_renderObjects )
+    {
+        auto ro = RadiumEngine::getInstance()->getRenderObjectManager()->getRenderObject( roIndex );
+        if ( ro->isVisible() ) { aabb.extend( ro->computeAabb() ); }
+    }
+    return aabb;
+}
+
 } // namespace Engine
 } // namespace Ra
--- a/src/Engine/Component/Component.cpp
+++ b/src/Engine/Component/Component.cpp
@@ -56,7 +56,7 @@ void Component::notifyRenderObjectExpired( const Core::Utils::Index& idx ) {
 
 Core::Aabb Component::computeAabb() const {
 
-    // component have not transformation, so ask render objects own bbox.
+    // component haven't any transformation, so ask render objects own aabb.
     Core::Aabb aabb;
     for ( const auto& roIndex : m_renderObjects )
     {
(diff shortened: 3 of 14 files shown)


src/GuiBase/Utils/KeyMappingManager.hpp
```diff
@@ -45,9 +45,14 @@ class RA_GUIBASE_API KeyMappingManager
     static std::string enumNamesFromKeyboardModifiers( const Qt::KeyboardModifiers& modifiers );
 
     Context getContext( const std::string& contextName ) {
-        return m_contextNameToIndex[contextName];
+        // use find so that it do not insert invalid context
+        auto itr = m_contextNameToIndex.find( contextName );
+        if ( itr != m_contextNameToIndex.end() ) return m_contextNameToIndex[contextName];
```
MathiasPaulin: return `itr->second` instead so that the search is done once.

src/GuiBase/Utils/KeyMappingManager.hpp
```diff
@@ -45,9 +45,14 @@ class RA_GUIBASE_API KeyMappingManager
...
+        return Context{};
```
MathiasPaulin: Not sure that default context is the better here.
Maybe a specific `ErrorContext` that will capture (and eventually log) all the operations might be more helpful.

src/GuiBase/Utils/KeyMappingManager.hpp
```diff
@@ -45,9 +45,14 @@ class RA_GUIBASE_API KeyMappingManager
...
     }
     KeyMappingAction getActionIndex( const Context& context, const std::string& actionName ) {
+        if ( context >= m_actionNameToIndex.size() || context.isInvalid() )
+            return KeyMappingAction{};
```
MathiasPaulin: Same as above ? A specific `ErrorAction` ?

src/GuiBase/Utils/KeyMappingManager.hpp
dlyr: Default ctor is Invalid index.

src/GuiBase/Utils/KeyMappingManager.hpp
dlyr: Same as above, default ctor is invalid.

src/Engine/Renderer/RenderObject/RenderObjectManager.cpp
```diff
@@ -130,6 +130,9 @@ size_t RenderObjectManager::getNumVertices() const {
     return result;
 }
 
+///\todo allow to have aabb, per component ...
+/// Well in fact I think getSceneAabb should not be done by RenderObjectManager
```
hoshiryu: I think we should just remove this method.

src/Engine/Renderer/RenderObject/RenderObjectManager.hpp
```diff
@@ -36,6 +36,7 @@ class RA_ENGINE_API RenderObjectManager final
 
     /**
      * @brief Get all render objects, the vector is assumed to be empty when called
+     * \todo what does the previous sentence means ....
```
hoshiryu: Should be some remains from the old version which took the list as output parameter.
Can you update the doc?

dlyr:
@hoshiryu not sure to catch your question.
The Entity transform applies to the entity, so that it is placed in world coordinates.

hoshiryu:
I mean when computing the aabb, should the RO use the complete transform (Entity * RO) or  just use its "local" transform and let the entity use its own on top?

dlyr:
Well, it seems that I have deleted the application of RO/entity transformations ... need to check this.
So if we think of AABB, either the RO return OBB, align in object frame, and entity extends its AABB with OBB, or RO return AABB in world coordinates ... 
But not AABB in entity frame, that will then extend AABB in world space, because this will make AABB bigger than needed.

src/Engine/Entity/Entity.cpp
```diff
@@ -70,6 +72,20 @@ void Entity::swapTransformBuffers() {
...
+Core::Aabb Entity::computeAabb() const {
+
+    using Ra::Core::Transform;
+    Core::Aabb aabb;
+
+    for ( const auto& component : m_components )
+    {
+        // transform is applied in ro aabb computation ...
+        aabb.extend( component->computeAabb() );
+        //  RA_DISPLAY_AABB( aabb, Ra::Core::Utils::Color( 0., 0., 1. ) );
```
hoshiryu: could you remove the commented display code please?

src/Engine/RadiumEngine.cpp
```diff
@@ -268,5 +269,22 @@ RadiumEngine::SystemContainer::iterator RadiumEngine::findSystem( const std::str
...
+    Core::Aabb aabb;
+
+    const auto& systemEntity = Engine::SystemEntity::getInstance();
+    auto entities            = m_entityManager->getEntities();
+    for ( const auto& entity : entities )
+    {
+        if ( entity != systemEntity )
+        {
+            aabb.extend( entity->computeAabb() );
+            // RA_DISPLAY_AABB( aabb, Ra::Core::Utils::Color( 1., 0., 1. ) );
```
hoshiryu: same here

Applications/MainApplication/Gui/MainWindow.cpp
```diff
@@ -41,6 +41,8 @@ namespace Gui {
 
 using namespace Core::Utils; // log
 
+///\todo import read/write settings from color point app
```
nmellado: Unrelated comment ?

src/Engine/Renderer/RenderObject/RenderObject.cpp
```diff
@@ -166,8 +166,8 @@ Core::Matrix4 RenderObject::getTransformAsMatrix() const {
     return getTransform().matrix();
 }
 
-Core::Aabb RenderObject::getAabb() const {
-    Core::Aabb aabb = m_mesh->getGeometry().computeAabb();
+Core::Aabb RenderObject::computeAabb() const {
+    const Core::Aabb& aabb = m_mesh->getGeometry().computeAabb();
```
nmellado: `computeAabb` returns a rvalue, you cannot use a reference here (at least, I don't understand what it does).
```suggestion
    Core::Aabb aabb = m_mesh->getGeometry().computeAabb();
```

nmellado:
Please also update from `master` (I let you do it in case you want to rewrite history)

dlyr:
Is everything ok with this PR ? please update your reviews.

src/Engine/Component/Component.cpp
```diff
@@ -54,5 +54,17 @@ void Component::notifyRenderObjectExpired( const Core::Utils::Index& idx ) {
     if ( found != m_renderObjects.cend() ) { m_renderObjects.erase( found ); }
 }
 
+Core::Aabb Component::computeAabb() const {
+
+    // component have not transformation, so ask render objects own bbox.
```
hoshiryu: typo: `have no`

src/Engine/RadiumEngine.hpp
```diff
@@ -165,6 +166,12 @@ class RA_ENGINE_API RadiumEngine
      */
     const std::vector<std::shared_ptr<Core::Asset::FileLoaderInterface>>& getFileLoaders() const;
 
+    /**
+     * Return the aabb of the scene by aggregating aabb from entity.
+     * @return the AABB
```
hoshiryu: no need for the second line right?

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -584,7 +584,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
     // manage picking mode
     auto ro = getRoMgr()->getRenderObject( m_pickedRoIdx );
     if ( m_pickedMesh->getRenderMode() == MeshRenderMode::RM_POINTS )
-    { return ro->getAabb().sizes().norm() / 500; }
+    { return ro->computeAabb().sizes().norm() / 500_ra; }
```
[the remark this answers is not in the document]
MathiasPaulin: Remarks was not resolved. Why 500_ra ? What is it ? Could be better to use a `constexpr Scalar TheMagnificientConstant = 500_ra;` at the top of the file and uses this constant here.

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -622,7 +622,7 @@ Scalar MeshFeatureTrackingComponent::getFeatureScale() const {
                2.0;
     }
     default:
-        return ro->getAabb().diagonal().norm() / 100_ra;
+        return ro->computeAabb().diagonal().norm() / 100_ra;
```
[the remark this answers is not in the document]
MathiasPaulin: NYR (Not Yet Resolved)

src/Engine/RadiumEngine.hpp
MathiasPaulin: This allows doxygen to generate a well formated documentation. The second line is mandatory. The first could be rewritten to be less similar ...

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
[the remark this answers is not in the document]
dlyr: Well, i thought I commented on this, but I surely click resolve before confirm my comment.
This is out of the scope of the PR, it's orthogonal to aabb computation, it's from MeshFeatureTrackingCompoenent.
The only thing I have do is to fix the constant to use correct suffix.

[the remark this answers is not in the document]
MathiasPaulin: Out of scope and not your code, I agree. But when fixing code, why not take 30s more to fix it _definitively_ or, at least for a long time ;) if the fix is trivial.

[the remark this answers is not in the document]
nmellado: I do agree this is out of the scope of this PR.

But I also strongly agree that moving this constant to a dedicated variable on which we can add a `\fixme` comment is a very good idea, and that is a good time to do it (low energy consumption for high benefit). We will fix the use of that variable later on, but at least please identify it.

src/Engine/RadiumEngine.hpp
hoshiryu: Then just use one line with `@returns the aabb of ...`

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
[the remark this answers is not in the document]
dlyr: Please PR on my code if you have a suggestion, I will accept it.

hoshiryu:
@dlyr I did a PR on your branch.

Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp
```diff
@@ -14,6 +14,8 @@
 #include <Engine/Renderer/RenderObject/RenderObject.hpp>
 #include <Engine/Renderer/RenderObject/RenderObjectManager.hpp>
 
+#define POINT_SCALE_FACTOR 1_ra / 500
```
MathiasPaulin: Prefer a constexpr rather than a macro. Radium is (mainly) modern C++, not C.
constexpr are typed expression, macro are not !
```suggestion
constexpr Scalar POINT_SCALE_FACTOR = 1_ra / 500;
```

dlyr:
@mathiaspaulin, could you accept the changes if it's fine for you ?

PR #451, opened 2019-07-14
Delete buffers, should fix #438

### LINKED ISSUE #437: Plugin MeshPaint is buggy

dlyr:
Replacement proposal for #438, but I'm not sure to catch all the issues, so please check ;)

I also reformat all code (in a separate commit), according to last script version.
So please check by commit to not get disturbed by the numerous changes.

---------------------- Update pull request manifest -----------------------

Fix #437 

When de-activating mesh paint and re-activating it on a mesh with no previous `per-vertex` colors, the mesh ends up all black and painting colors does not work.
Actually this is due to the fact that, when de-activating colors painting, the colors are set back to what they were before painting, which in this case is `none`.
Thus when re-activating painting, since the mesh has the color attribute, the plugin just copies it and no validity check is done.
Hence when painting the colors, colors are set in an empty container (this should crash but somehow does not).

- On color painting activation, the plugin request the addition of a color attribute to the mesh.
If there is only one, this one is selected as the `to be painted` and previous colors are saved.
Otherwise, a new one is created and filled with a default color.
- On color painting de-activation, the colors are set to what they were beforehand, i.e. either the saved ones or the created attribute is removed.

### DIFF
--- a/Applications/MainApplication/Gui/ColorWidget.cpp
+++ b/Applications/MainApplication/Gui/ColorWidget.cpp
@@ -27,14 +27,14 @@ void ColorWidget::colorChanged() {
 
 void ColorWidget::mousePressEvent( QMouseEvent* /*event*/ ) {
 
-        QColor color = QColorDialog::getColor( m_currentColor );
-        if ( color != m_currentColor )
-        {
-            m_currentColor = color;
-            colorChanged();
+    QColor color = QColorDialog::getColor( m_currentColor );
+    if ( color != m_currentColor )
+    {
+        m_currentColor = color;
+        colorChanged();
 
-            emit newColorPicked( m_currentColor );
-        }
+        emit newColorPicked( m_currentColor );
+    }
 }
 } // namespace Gui
 } // namespace Ra
--- a/Applications/MainApplication/Gui/MainWindow.cpp
+++ b/Applications/MainApplication/Gui/MainWindow.cpp
@@ -98,7 +98,7 @@ void MainWindow::createConnections() {
     connect( actionToggle_Local_Global,
              &QAction::toggled,
              mainApp,
-             &Ra::GuiBase::BaseApplication::askForUpdate);
+             &Ra::GuiBase::BaseApplication::askForUpdate );
 
     connect( actionGizmoOff, &QAction::triggered, this, &MainWindow::gizmoShowNone );
     connect( actionGizmoTranslate, &QAction::triggered, this, &MainWindow::gizmoShowTranslate );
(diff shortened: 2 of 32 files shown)


hoshiryu:
There is an issue with Travis-CI: the compilation is marked as `passed` while there was a compilation error:
```
/home/travis/build/STORM-IRIT/Radium-Engine/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp: In member function ‘Scalar MeshFeatureTrackingPlugin::MeshFeatureTrackingComponent::getFeatureScale() const’:
/home/travis/build/STORM-IRIT/Radium-Engine/Plugins/MeshFeatureTracking/src/MeshFeatureTrackingComponent.cpp:587:18: error: ‘using element_type = class Ra::Engine::RenderObject {aka class Ra::Engine::RenderObject}’ has no member named ‘computeAabb’
     { return ro->computeAabb().sizes().norm() / 500; }
                  ^~~~~~~~~~~
```

hoshiryu:
Regarding the code for paint activation, it seems the code should be changed, as proposed in #438 :
```[cpp]
if ( m_currentColorAttribHdl.idx().isInvalid() )
{
    m_baseColors.clear();
    m_isBaseColorValid = false;
    triangleMesh.colorize( Ra::Core::Utils::Color::Skin() );
    m_currentColorAttribHdl =
        triangleMesh.getAttribHandle<Ra::Core::Vector4>( colAttribName );
    m_mesh->setDirty( Ra::Engine::Mesh::VERTEX_COLOR, true );
}
else
{
    m_isBaseColorValid = true;

    m_baseColors = triangleMesh.getAttrib( m_currentColorAttribHdl ).data(); // copy
}
```
should be simplified as:
```[cpp]
m_baseColors  = triangleMesh.getAttrib( m_currentColorAttribHdl ).data();
m_isBaseColorValid = m_baseColors.size() != 0;
if (  !m_isBaseColorValid )
{
    triangleMesh.colorize( Ra::Core::Utils::Color::Skin() );
    m_mesh->setDirty( Ra::Engine::Mesh::VERTEX_COLOR, true );
}
```
where the first line creates the attribute if it doesn't exist, while clears the colors otherwise.

dlyr:
Well, reading your proposal, I think it is based on the fact that addAttrib do not allocate space to store the attribute, which might change.
I prefer the explicit version

hoshiryu:
fair

PR #467, opened 2019-09-04
Observer

dlyr:

Add a new observable class, so that other class can derive and benefit from a observer mecanism.
You can attach a function to the class, and the class has to call notify to call those function.
It's some kind of minimal signal/slot implementation.

KeyMappingManager has been updated to use Observable as a base class.

* **Other information**:

### DIFF
--- a/src/Core/Utils/Observable.hpp
+++ b/src/Core/Utils/Observable.hpp
@@ -19,6 +19,7 @@ class Observable
     Observable()   = default;
 
     /// attach an \p observer that will be call on subsecant notify
+    /// \return an unique int to identify the observer, could be used to pass to Obeservable::detach
     int attach( Observer observer ) {
         m_observers.insert( std::make_pair( ++m_currentId, observer ) );
         return m_currentId;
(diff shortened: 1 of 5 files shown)


src/Core/Utils/Observable.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <Core/Utils/Log.hpp>
+
+#include <functional>
+#include <map>
+
+namespace Ra {
+namespace Core {
+namespace Utils {
+
+/// Simple observable implementation with void observer() notification
```
MathiasPaulin: ... `void observer(Arg...)` ...

MathiasPaulin:
```suggestion
/// Simple observable implementation with void observer( Args... ) notification
```

tests/CoreTests/src/observer.cpp
```diff
@@ -0,0 +1,115 @@
...
+    RA_VERIFY( a.m_a == 2, "Notify failed." );
+    RA_VERIFY( A::m_b == 1, "Detach failed." );
+
+    test.attach( A::g );
+    test.attach( A::g );
+    test.attach( bf );
+    test.attach( [&c]() { c++; } );
+    test.notify();
+
+    RA_VERIFY( c == 4, "Notify failed." );
```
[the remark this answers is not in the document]
dlyr: that's not true

PR #471, opened 2019-09-26
Fixup for AttribManager move ;

hoshiryu:
TODO: use unique_ptr instead of raw pointers for value_type ?

UPDATE the form below to describe your PR.

* **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Partial fix of #470 

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

* **Other information**:

PR #472, opened 2019-09-26
Use unique_ptr in AttribManager.

### LINKED ISSUE #470: Memory Leak on std::move( AttribManager )
I ran into a memory leak when moving a `TriangleMesh`.
Within a plugin of mine, I do the following:
```
auto mesh = Ra::Core::Geometry::makePlaneGrid( height, width, scale );
engineMesh->loadGeometry( std::move( plane ) ); // there was already another mesh
```
This unfortunately leads to a memory leak.
After looking into the code of both `Engine::Mesh` and `Core::TriangleMesh` I found:
```
Mesh::loagGeometry( Core::TriangleMesh&& mesh ) {
m_mesh = std::move( mesh ); // which is legit
...
}
```
and
```
inline TriangleMesh& TriangleMesh::operator=( TriangleMesh&& other ) {
    m_triangles      = std::move( other.m_triangles );
    m_faces          = std::move( other.m_faces );
    m_vertexAttribs  = std::move( other.m_vertexAttribs );
    m_verticesHandle = std::move( other.m_verticesHandle );
    m_normalsHandle  = std::move( other.m_normalsHandle );
    return *this;
}
```
which is legit too!

Thus I went to `AttribManager`:
```
AttribManager& operator=( AttribManager&& m ) {
    m_attribs      = std::move( m.m_attribs ); // which is a std::vector<AttribBase*>
    m_attribsIndex = std::move( m.m_attribsIndex ); // which is a std::map<std::string, Index>
    return *this;
}
```
all seems fine there, except that we are moving raw pointers, and thus these are not deleted.

So a simple fix would be to clear `m_attribs`before moving.

A better solution would be to use `unique_ptr` instead of raw pointers for `AttribManager::value_type`

dlyr:

Bug fix

memory leak

use unique_ptr to have auto delete

* **Other information**:
Supersede #471, resolve #470 
To be tested

### DIFF
--- a/src/Core/Utils/Attribs.hpp
+++ b/src/Core/Utils/Attribs.hpp
@@ -124,7 +124,7 @@ class Attrib : public AttribBase
     size_t getSize() override { return m_data.size(); }
 
     /// \warning Does not work for dynamic and sparse Eigen matrices.
-    int getStride() override { return sizeof( typename Container::value_type ); }
+    int getStride() override { return sizeof( value_type ); }
 
     bool isFloat() const override { return std::is_same<float, T>::value; }
     bool isVec2() const override { return std::is_same<Vector2, T>::value; }
@@ -139,9 +139,9 @@ template <typename T>
 class AttribHandle
 {
   public:
-    typedef T value_type;
-    using Container = typename Attrib<T>::Container;
-
+    using value_type = T;
+    using AttribType = Attrib<T>;
+    using Container  = typename Attrib<T>::Container;
     /// compare two handle, there are the same if they both represent the same
     /// attrib (type and value).
     template <typename U>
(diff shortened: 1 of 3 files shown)


nmellado:
@hoshiryu can you test this PR actually solve your memory leak please ?

src/Core/Utils/Attribs.hpp
```diff
@@ -124,7 +124,7 @@ class Attrib : public AttribBase
     size_t getSize() override { return m_data.size(); }
 
     /// \warning Does not work for dynamic and sparse Eigen matrices.
```
hoshiryu: What about this warning?

dlyr: still true

dlyr:
Regarding the doc comment, for a user point of view, it do not change if container use unique_ptr or raw ptr. No need to change this.

PR #474, opened 2019-10-27
Conf fix

dlyr:
Add dot file to git ignore
Add a compile flag to remove some eigen warnings (should be temporary)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,11 +7,9 @@
 *.orig
 *.swp
 *.old
-.idea/
 \#*\#
-.\#*\#
-.vs
-.clang-format
+# All dot files, if you wish to track one, just add it
+.*
 
 # Misc
 *Thumbs.db
@@ -20,10 +18,6 @@
 Bundle*/
 build*/
 cmake-build-*/
-
-# Mercurial stuff (e.g from Eigen)
-.hg*
-
 
 # external libraries
 external/eigen


.gitignore
```diff
@@ -12,6 +12,7 @@
 .\#*\#
 .vs
 .clang-format
+.*
```
hoshiryu: won't that make git ignore all `.git*` files as well as the `.ci` and `.travis.yml` files?

nmellado: Files that are already tracked are not considered by .gitignore.

hoshiryu: ok, then the two lines are useless right?

nmellado: Yes.

.gitignore
```diff
@@ -12,6 +12,7 @@
 .\#*\#
 .vs
 .clang-format
```
MathiasPaulin: instead of ignoring .clang-format, it could be useful to track this file here. Thisis due to how clang-format work. When launched without other arguments than files, clang-format search the directories up until .clang-format is found. Tracking this file here instead of in the 'script' subdir will ensure the right configuration file is used when formating files.

nmellado: I dont't know about tracking `.clang-format`, but we need to track the script directory anyway.

MathiasPaulin: Sure, the script directory must be tracked.
But, to ease IDE integration, having the file `.clang-format`at the root of the Radium dir could be helpful.
From the `clang-format --help`
```
  -style=<string>           - Coding style, currently supports:
LLVM, Google, Chromium, Mozilla, WebKit.
    Use -style=file to load style configuration from
    .clang-format file located in one of the parent
    directories of the source file (or current
    directory for stdin).
    Use -style="{key: value, ...}" to set specific
    parameters, e.g.:
        -style="{BasedOnStyle: llvm, IndentWidth: 8}"
```
Note that the `.clang-format` file is search by default if nothing is given using this option.

dlyr: well, I have a symlink `.clang-format -> scripts/clang-format`
and  afaik symlink are not a so good idea in git repo (some windows file system do not support them)
I think it's user responsability to install the script, with `scripts/install-scripts-linux.sh` for instance

nmellado:
@MathiasPaulin can you dismiss you review / approve the changes ?

PR #477, opened 2019-11-04
Unique box checking

ThibaultLejemble:

New ui feature : show one single entity with a simple _ctrl + left click_ on its checkbox (as in Meshlab).

To show one single entity, one has to hide all entities, and then click on the checkbox.  

One _ctrl + left click_ on a checkbox hide all entities except this one.

None.

* **Other information**:
It works only for entities.
When a _ctrl + left click_ is done, the entity and all of its components and render objects are shown, even if some of them where hidden before.
I only tested it on simple situation with one render object and one component per entity.


hoshiryu:
Interesting improvement of the visibility Checkboxes, though I think it should be extended to every item in the tree. I tested it on scenes with several entities, with several components and ROs, working fine.

PR #486, opened 2020-01-06
Fix 454

hoshiryu:

- Fix bug on `IndexedGeometry<T>::loadGeometry`,
- Fix bug on Subdivider destructor.

- When replacing the `Ra::Core::Geometry::IndexedGeometry` of a `Ra::Engine::IndexedGeometry` with a new one by calling `Ra::Engine::IndexedGeometry<T>::loadGeometry()`, indices are not tagged as dirty, which is only done on the first call to `Ra::Engine::IndexedGeometry<T>::updateGL_specific_impl()`. Thus the new `Ra::Core::Geometry::IndexedGeometry` is not displayed correctly since indices are not send on OpenGL side.
- When calling the destructor of a `Ra::Core::Geometry::*Subdivider`, nothing specific is done. Thus when the destructor of `OpenMesh::Subdivider::Uniform::SubdividerT` (which it publicly inherits) is called, this one calls `OpenMesh::Subdivider::Uniform::SubdividerT::detach()`, which might call the `cleanup` method, which is pure virtual in `OpenMesh::Subdivider::Uniform::SubdividerT::detach()`.
This causes a crash.

- When calling `Ra::Engine::IndexedGeometry<T>::loadGeometry()`, indices are checked and tagged as dirty if necessary. Doing so they will be updated on OpenGL side when calling `Ra::Engine::IndexedGeometry<T>::updateGL_specific_impl()`.
- The `Ra::Core::Geometry::*Subdivider`'s destructor now calls `detach`, which calls the overriden `cleanup` method and makes the last call to `detach` (in`OpenMesh::Subdivider::Uniform::SubdividerT`'s destructor) to do nothing.

* **Other information**:

### DIFF
--- a/src/Engine/Renderer/Mesh/Mesh.inl
+++ b/src/Engine/Renderer/Mesh/Mesh.inl
@@ -387,13 +387,13 @@ IndexedGeometry<T>::IndexedGeometry( const std::string& name,
 
 template <typename T>
 void IndexedGeometry<T>::loadGeometry( T&& mesh ) {
+    if ( base::m_mesh.m_indices != mesh.m_indices ) { m_indicesDirty = true; }
     m_numElements = mesh.m_indices.size() * base::CoreGeometry::IndexType::RowsAtCompileTime;
     base::loadGeometry_common( std::move( mesh ) );
 }
 
 template <typename T>
 void IndexedGeometry<T>::updateGL_specific_impl() {
-
     if ( !m_indices )
     {
         m_indices      = globjects::Buffer::create();
--- a/src/Engine/Renderer/Mesh/Mesh.inl
+++ b/src/Engine/Renderer/Mesh/Mesh.inl
@@ -38,6 +38,11 @@ std::string AttribArrayDisplayable::getAttribName( MeshData type ) {
     if ( type == VERTEX_WEIGHT_IDX ) return {"in_weight_idx"};
     return {"invalid mesh data attr name"};
 }
+///////////////// VaoIndices  ///////////////////////
+
+void VaoIndices::setIndicesDirty() {
+    m_indicesDirty = true;
+}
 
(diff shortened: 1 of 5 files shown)


src/Engine/Renderer/Mesh/Mesh.inl
```diff
@@ -387,13 +387,13 @@ IndexedGeometry<T>::IndexedGeometry( const std::string& name,
 
 template <typename T>
 void IndexedGeometry<T>::loadGeometry( T&& mesh ) {
+    if ( base::m_mesh.m_indices != mesh.m_indices ) { m_indicesDirty = true; }
```
MathiasPaulin: This could be extremely costly to compare the contents of two vectors. just think about comparing the indices of two triangle meshes with 500k triangles (medium complexity) - 1.5M of comparisons if indices are the same ... ? What about an other mean, very less costly, to set indices dirty (a topologyChanged method ?). @dlyr What do you think of that ?

dlyr: Well, we will discuss that point AFK.
But the loadGeometry do change topology since it changes the associated CoreGeometry.

I think we don't have to test for equality, loadGeometry should be only called to move another CoreGeometry to the Render::Mesh (and release the current one).
If you want to update attributes, use getCoreGeometry() and modify the attributes.

Right way to do is to have similar observer thing on indices, small patch could be to add a setDirty to the VaoIndices class.

MathiasPaulin: I agree on the usage restriction of loadGeometry. And this restriction must be mentioned in the method documentation.
It could be helpful also to add documentation about this use case as it is quite current.

dlyr:
On my side it's ready to accept and merge.

PR #476, opened 2019-10-29
Toward v2 - Improvement of the material interface (both C++ and GLSL)

MathiasPaulin:
UPDATE the form below to describe your PR.

* **Please check if the PR fulfills these requirements**
- [X] The commit message follows our guidelines
- [X] Docs have been added / updated (for bug fixes / features)

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This is a first step toward Material(BSDF), Technique(Configuration of the renderer) and GLSL code generation that will be available in radium-v2.

Documentation update (Radium Material)
    Updated
Improve Material interface
   GLSL part : allow to dynamically assemble shaders, see Render-App on Gitlab

Make Plain a usable material (lambertian BSDF)
   Not loadable. Only configurable by program
   Primitives and gizmos use this material

PR #506, opened 2020-03-27
Todos doc and comment.

dlyr:
Fixes some of the forgotten review of #476

### DIFF
--- a/doc/developer/conventions.md
+++ b/doc/developer/conventions.md
@@ -50,7 +50,7 @@ use `Scalar()` or `_ra` suffix when defining numbers from literals (e.g. `auto a
 * Brace style : keep it consistent across files.
 * Case style : CamelCase
 * Only classes have their first letter capitalized. Functions and variables don't.
-* Class members have the `m_` prefix. Other prefixes (apart from `g_`) are discouraged.
+* Class members have the `m_` prefix. Other prefixes (apart from `g_` for globals, and `s_` for static classe members) are discouraged.
 * no ifs / for / while one-liners. Braces everywhere. Even for cases in a switch.
 * separate different clauses in a boolean expression with parens.
 `( (a || b) && ( c+d < 0) )`
--- a/src/Core/Utils/Index.inl
+++ b/src/Core/Utils/Index.inl
@@ -40,7 +40,8 @@ constexpr void Index::setValue( const IntegerType i ) {
 }
 
 /// CAST
-constexpr Index::operator int() const {
+/// \todo manage other types ? like unsigned with an assert if invalid ?
+constexpr Index::operator IntegerType() const {
     return m_idx;
 }
 
--- a/src/Engine/Component/GeometryComponent.cpp
+++ b/src/Engine/Component/GeometryComponent.cpp
@@ -8,34 +8,28 @@
 #include <Core/Geometry/Normal.hpp>
 #include <Core/Resources/Resources.hpp>
 #include <Core/Utils/Color.hpp>
+#include <Core/Utils/Log.hpp>
 
 #include <Engine/Managers/ComponentMessenger/ComponentMessenger.hpp>
-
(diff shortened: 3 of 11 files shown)


nmellado:
Great thanks @dlyr 
@hoshiryu are all your recommended changes included ?

MathiasPaulin:
> We still do not agree with @MathiasPaulin about the OpenGL error check (can raise `GL_INVALID_OPERATION`).
Which and why ?

hoshiryu:
Just a question @MathiasPaulin why is there a `Material` in a `RenderObject` since we have it as the `ShaderParameterProvider` of the `RenderTechnique` ?

MathiasPaulin:
> Just a question @MathiasPaulin why is there a `Material` in a `RenderObject` since we have it as the `ShaderParameterProvider` of the `RenderTechnique` ?

Because nothing says that only materials could be `ShaderParameterProvider`. 
We can define rendertechnique (we have some on our renderers) that need a much more advanced `ShaderParameterProvider` than materials but that still use the material for defining a subset of its parameter. Materials is a BSDF calculator. If we need to compute the appearance of an object in a different way, we migh create RenderTechnique with `ShaderParameterProvider` that is not a meterial.

hoshiryu:
> > We still do not agree with @MathiasPaulin about the OpenGL error check (can raise `GL_INVALID_OPERATION`).
> > Which and why ?

ok I was looking to an older version's doc, my mistake

hoshiryu:
> > Just a question @MathiasPaulin why is there a `Material` in a `RenderObject` since we have it as the `ShaderParameterProvider` of the `RenderTechnique` ?
> 
> Because nothing says that only materials could be `ShaderParameterProvider`.
> We can define rendertechnique (we have some on our renderers) that need a much more advanced `ShaderParameterProvider` than materials but that still use the material for defining a subset of its parameter. Materials is a BSDF calculator. If we need to compute the appearance of an object in a different way, we migh create RenderTechnique with `ShaderParameterProvider` that is not a meterial.

Then what's the better way to change the `Material` if we don't know what's in there? Through the `RenderObject` or the `RenderTechnique`?

MathiasPaulin:
> > > We still do not agree with @MathiasPaulin about the OpenGL error check (can raise `GL_INVALID_OPERATION`).
> > > Which and why ?
> 
> ok I was looking to an older version's doc, my mistake

FYI : https://www.khronos.org/registry/OpenGL-Refpages/gl4/

MathiasPaulin:
> > > Just a question @MathiasPaulin why is there a `Material` in a `RenderObject` since we have it as the `ShaderParameterProvider` of the `RenderTechnique` ?
> > 
> > 
> > Because nothing says that only materials could be `ShaderParameterProvider`.
> > We can define rendertechnique (we have some on our renderers) that need a much more advanced `ShaderParameterProvider` than materials but that still use the material for defining a subset of its parameter. Materials is a BSDF calculator. If we need to compute the appearance of an object in a different way, we migh create RenderTechnique with `ShaderParameterProvider` that is not a meterial.
> 
> Then what's the better way to change the `Material` if we don't know what's in there? Through the `RenderObject` or the `RenderTechnique`?

The `RenderTechnique` has no knowledge of any material! They know only `ShaderParameterProvider`.
So, you have your answer.

But remember, it is the renderer that know how to consgtruct the rendertechnique he needs. So, if you change the material, you have to ask the renderer to build the renderTechnique for the RenderObject through a call to the method `Renderer::buildRenderTechnique( RenderObject* ro )`

nmellado:
@MathiasPaulin are you fine with the changes ?

nmellado:
@hoshiryu are you fine with these changes (please update your review accordingly)

hoshiryu:
already up to date

nmellado:
@MathiasPaulin / @dlyr  can you answer @hoshiryu comments please ?

PR #507, opened 2020-03-27
 No need to set draw or read buffers …, fix #504

### LINKED ISSUE #504: [GL] Add support for single buffering
Using the master branches of both `Radium-Engine` and `Radium-Apps` I have two bugs on my personal laptop that I didn't have on another computer:
 - On Debug, line `src/Engine/Renderer/Renderer.cpp:557` crashes (b.t.w. there is no OpenGL error check on that line or right after):
```glDrawBuffer( GL_BACK );```
Strangely it doesn't occur on another of my projects that use both the Viewer and Renderer from Radium.
Commenting line  `src/Engine/Renderer/Renderer.cpp:557`, the app runs well in debug.

Just for info:
```
> uname -a
Linux hoshiryu 4.19.0-8-amd64 #1 SMP Debian 4.19.98-1 (2020-01-26) x86_64 GNU/Linux
> g++ --version
g++ (Debian 8.3.0-6) 8.3.0
> cmake --version
cmake version 3.13.4
```

MathiasPaulin:
…on default framebuffers

Application crashes when used with a single-buffered default framebuffer. 

Application will no crash and keep the default framebuffer with its default configuration according to the openGL Spec.

* **Other information**:

### DIFF
--- a/src/Engine/Renderer/Renderer.cpp
+++ b/src/Engine/Renderer/Renderer.cpp
@@ -177,8 +177,6 @@ void Renderer::initialize( uint width, uint height ) {
 
     resize( m_width, m_height );
 
-    glDrawBuffer( GL_BACK );
-    glReadBuffer( GL_BACK );
 }
 
 Renderer::PickingResult Renderer::doPickingNow( const PickingQuery& query,
@@ -563,12 +561,10 @@ void Renderer::restoreExternalFBOInternal() {
     if ( m_qtPlz == 0 )
     {
         GL_ASSERT( glBindFramebuffer( GL_FRAMEBUFFER, 0 ) );
-        glDrawBuffer( GL_BACK );
     }
     else
     {
         GL_ASSERT( glBindFramebuffer( GL_FRAMEBUFFER, m_qtPlz ) );
-        GL_ASSERT( glDrawBuffers( 1, buffers ) );
     }
 }
 


dlyr:
I have no way to test if this PR fix #504 , but don't we need to also test if double buffer is available, and change the requested surface (e.g. [here](https://github.com/STORM-IRIT/Radium-Engine/blob/ac1a8565bd30a2136e0d70fa10aad3f10de88863/tests/ExampleApps/DrawPrimitivesApp/main.cpp#L24))

MathiasPaulin:
> I have no way to test if this PR fix #504 , but don't we need to also test if double buffer is available, and change the requested surface (e.g. [here](https://github.com/STORM-IRIT/Radium-Engine/blob/ac1a8565bd30a2136e0d70fa10aad3f10de88863/tests/ExampleApps/DrawPrimitivesApp/main.cpp#L24))

The issue #504 relates an `INVALID_OPERATION` on the call of glDrawBuffer. If we don't call this function, the error disappear.
But the core of this PR is that we have to left untouched the default frame-buffer configurations as it is independent on our own frame-buffers and only used for final display. 

According to the FB configurations sent on the issue, the request for a double-buffer context might succeed and the PR will remove the error on single-buffered FB.

@hoshiryu could you test and report success or failure ?

MathiasPaulin:
> Hmm, I still have the issue on resources location:
This have nothing to do with the current PR and the issue it is linked. If the PR solve the issue #504 and correspond to the Radium desired code quality : approved. Else, changes must be requested only on the PR and its goals.

PR #520, opened 2020-04-06
Fix #519

### LINKED ISSUE #519: Wrong state on rendertechnique passes after OpenGL update
When a shader is compiled in a render technique, its state (dirty bit) is not reset and the system search for a compiled shader with the same configuration at each frame. As the configuration has not changed, this implies superfluous computation time.

MathiasPaulin:
dirty bit was not reset on a pass of a renderTechnique after OpenGL Update.

Bug fix (see issue #519)

dirty bit on RenderTechnique/pass was not reset

dirty bit on RenderTechnique/pass  is reset


MathiasPaulin:
> I think these flag stuff could be controlled by inline function like isDirty(pass) and setDirty(pass).
> add to TODO list ;)

Done in the PR #521

Issue #529, opened 2020-04-21
Radium Engine UUID


dlyr:
Following discussion from #527 , it's time to think of the name of the organization.

This name is used in default argument of BaseApplication, Doxygen configuration and plugin metadata.

First suggestions: 
  - use lab name, optionally with team name e.g. fr.irit or fr.irit.storm
  - use a name dedicated to radium e.g. org.radium-engine (the app have to take something like main-app

Issue #44, opened 2015-09-24
Z-buffering of UI drawables


louen:
At the moment UI drawables ignore the z buffer, which allows us to keep them visible above other normal scene objects.

We should try to clear the normal z buffer and draw the UI elements with a fresh buffer to keep them above the rest but still have some depth in their displau

PR #531, opened 2020-04-21
Small setting editor available for BaseApplication derived applications.

MathiasPaulin:
This add a "Setting editor" to any BaseApplication derived apps.
This PR is far for being ready to review but define a first, minimalistic setting editor.

* **Other information**:


nmellado:
This PR is very old. Does it make sense to keep it open ? Should we move it as an issue or a discussion ?

From what I see the setting editor would still make sense today. Shall we ask a developer to update/rebase the proposed code against current `master` ?

MathiasPaulin:
> This PR is very old. Does it make sense to keep it open ?

It make sense. I'll rebase on master as, as far as I know, this PR only interact with Gui ...

MathiasPaulin:
The proposal for a setting editor is updated and rebased on the master branch. 
Usage demonstration (and question) is on the PR on Radium-Apps #44 (https://github.com/STORM-IRIT/Radium-Apps/pull/44)
For now, the editor is quite limited and unfriendly but allow to change the path from where is loaded the KeyMappingManager configuration file.

Feel free to test/comment/develop/....

nmellado (dismissed):

nmellado:
For me the code seems to be ready to be merged.

@MathiasPaulin in your initial PR description you were mentioning:

> Several problems must be solved before integrating this PR into Radium :
> - Issue #529 that discuss about the Radium Engine UUID. This is related to the name and access of the QSettings file that allow application-based, organization-based or domain-based configuration
>  - Settings organization as there are some inconsistancy in the management of keymapping/config files and the way they are loaded ...
>  - The need for persistant settings ...

Are these comments still up to date ? Shall we move them in dedicated issues ?

dlyr (dismissed):
I'm fine with the PR (and with the new style, "ça claque" more or less ...)
But I don't get why the help button since the signal is not catch ...
Moreover it can be a good idea to have settings "reset to default" and a ligne by possible settings (here only modified settings are shown, e.g. background color)
Last, settings edits are not applied on the fly, only at startup, I don't know how easy is to link settings editor to the corresponding slot in the application.
Nevertheless, the PR could be accepted as is,

nmellado:
Fine for me as well (need rebase on top of `master`.

MathiasPaulin:
> I'm fine with the PR (and with the new style, "ça claque" more or less ...)
> But I don't get why the help button since the signal is not catch ...

This let the application to define its own help. 

> Moreover it can be a good idea to have settings "reset to default" and a ligne by possible settings (here only modified settings are shown, e.g. background color)

I agree.

> Last, settings edits are not applied on the fly, only at startup, I don't know how easy is to link settings editor to the corresponding slot in the application.

I'll search how to do that

> Nevertheless, the PR could be accepted as is,

I don't think so. Some "style" question should be resolved before that (see the PR to come)

nmellado:
I agree with @MathiasPaulin that we should not have a custom style for a given window: style must be defined at the application level for better consistency.
Also, we need to be sure that a style is supported on all operating systems (e.g. fonts).

nmellado (changes requested):
Commit 6209d19 introduce non portable look and field.
Please reset to default.

MathiasPaulin:
Closed as this. Will be splitted in 2.

PR #535, opened 2020-04-22
Fix normals for TriangleMeshComponent ;

hoshiryu:

Bug fix.
The doc about `AttribArrayGeometry` says that `Points and Normals, defining the geometry, are always present.`
But this is not always the case with the `TriangleMeshComponent`.

When building a `TriangleMeshComponent` from a `GeometryData`, the created `TriangleMesh` might not have normals if the `GeometryData` does not provide them.

When building a `TriangleMeshComponent` from a `GeometryData`, the created `TriangleMesh` will have normals initialized to 0 if the `GeometryData` does not provide them.

* **Other information**:


dlyr:
Clearly, please update the doc and discard these changes, we will update mesh data structures such that normals will be treated as other attirbutes (and hence easy to check if there is normals).

PR #537, opened 2020-04-23
Fix Mesh pre-transform problem.

### LINKED ISSUE #536: Local transformation is applied once on geometries
When loading a point cloud or a triangle mesh, the geometric data are transformed once using the transformation matrix associated with the geometry part and the resulting renderobject store an identity matrix used at runtime.
This cause two problems : 
1. Only Vertices and normals are transformed. If tangents are provided, they are not transformed, resulting in incoherency between tangents and normals.
2. This break the load/process/save interoperability workflow with other tools as transformations are lost on the Radium side.

MathiasPaulin:
resolves #536 

Store the local transformation matrix of a geometric component into the associated renderobject instead of applying the transformation once at load-time.

See issue #536

Mesh (and point cloud) are not transformed once. The transformation matrix of the associated renderObject store now the mesh local transformation.

* **Other information**:

### DIFF
--- a/src/Engine/Component/GeometryComponent.cpp
+++ b/src/Engine/Component/GeometryComponent.cpp
@@ -59,7 +59,7 @@ TriangleMeshComponent::TriangleMeshComponent( const std::string& name,
     GeometryComponent( name, entity ), m_displayMesh( new Engine::Mesh( name ) ) {
     setContentName( name );
     m_displayMesh->loadGeometry( std::move( mesh ) );
-    finalizeROFromGeometry( mat );
+    finalizeROFromGeometry( mat, Core::Transform::Identity() );
 }
 
 //////////// STORE Mesh/TriangleMesh here instead of an index, so don't need to request the ROMgr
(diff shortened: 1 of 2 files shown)


src/Engine/Component/GeometryComponent.cpp
```diff
@@ -73,33 +73,32 @@ void TriangleMeshComponent::generateTriangleMesh( const Ra::Core::Asset::Geometr
     std::string name( m_name );
     name.append( "_" + m_contentName );
 
-    std::string meshName = name;
-    meshName.append( "_Mesh" );
+    std::string meshName = m_contentName;
```
nmellado, line 76: This change is not related to the PR, please remove.

src/Engine/Component/GeometryComponent.cpp
```diff
@@ -73,33 +73,32 @@ void TriangleMeshComponent::generateTriangleMesh( const Ra::Core::Asset::Geometr
...
     Ra::Core::Geometry::TriangleMesh::PointAttribHandle::Container vertices;
     Ra::Core::Geometry::TriangleMesh::NormalAttribHandle::Container normals;
-
-    const auto T = data->getFrame();
-    const Ra::Core::Transform N( ( T.matrix() ).inverse().transpose() );
-
+    // data->getFrame()
     vertices.resize( data->getVerticesSize(), Ra::Core::Vector3::Zero() );
 
+    // Use std::copy ?
```
nmellado, line 87: Yes, please update this code with algorithms from the `std` library, like copy and transform.

src/Engine/Component/GeometryComponent.cpp
MathiasPaulin: This change allow also to keep consistency in a load/process/save workflow as the name of the low levels components (mesh, materials, ... ) will remains the same instead of being pre and post fixed by some Old-style Radium-specific informations (once apon a time in the Radium life, it seems that pre and postfix of the name - e.g _RO, GEOM_, _MESH_ and so on - where used to identify the type or semantic of these component. Now, we have others informations that allow to identify that and, at least on input/output element, keeping the name unchanged is important in a cooperation workflow.

src/Engine/Component/GeometryComponent.cpp
MathiasPaulin: The replaced code was equivalent of a std::transform. Here, we have to copy the data without modification. std::copy will do that but parallel execution policy are not yet well supported on several systems. Well, this code being used once, at the creation of a component, if its execution is sequential instead of parallel, the time penalty will not be so much. So, OK, I'll change this, trying the std::par policy.

src/Engine/Component/GeometryComponent.cpp
dlyr: Why not, but maybe in another PR, taking point and vol into account, and remove unnecessary code then (do the `name` variable still in use ?)

nmellado: I do agree.
Let’s go step by step

src/Engine/Component/GeometryComponent.cpp
nmellado: STD::transform can be used for the normals.

PR #539, opened 2020-04-24
Keyframedvalue

hoshiryu:

This PR introduce the KeyFramedValue class that manages time-dependent values of any type.
It stores a set of time-value keyframes that can be interpolated using any kind of interpolator, such as the new linear interpolators.
This PR also cleans up some documentation about the Animation data and process.

The class `Ra::Core::Animation::Animation` does not exist anymore.

* **Other information**:


hoshiryu:
Ready for review

hoshiryu:
Ready for review

hoshiryu:
I am not sure about what you want me to put in there.
There is no new system in this PR.
If you are talking about the online keyframing of data such as color, this should be considered with the PR #541 and #542, for which I already provide examples in dedicated branches.

MathiasPaulin:
At least a test that validates the storage and interpolation and demonstrate how to use the proposed functionality.

hoshiryu:
I forgot to mention the PR on the [SkeletonBasedAnimation Plugin](https://gitlab.com/Storm-IRIT/radium-official-plugins/skeletonbasedanimation/-/merge_requests/10) related to the new keyframes, which might be what you are looking for

src/Core/Animation/KeyFramedValueInterpolators.hpp
```diff
@@ -0,0 +1,114 @@
...
+    // exact match
+    if ( it != keyframes.end() )
+    {
+        t0 = it->first;
+        t1 = t0;
+        dt = 0.0;
+        return;
+    }
+    // in-between
+    auto upper = keyframes.upper_bound( t );
```
nmellado, line 59: Avoid traversing the whole collection twice: 
```suggestion
    auto upper = std::upper_bound(it, keyframes.end(), t );
```

src/Core/Animation/KeyFramedValueInterpolators.hpp
```diff
@@ -0,0 +1,114 @@
...
+ * \param[out] dt Linear parameter, \f$ \in [0;1] \f$, of \p t between \p t0 and \p t1.
+ * \note If \p t is lower than the first keyframe's time tf,
+ *       then \p t0 = \p t1 = tf and \p dt = 0.
+ *       If \p t is higher than the last keyframe's time tl,
+ *       then \p t0 = \p t1 = tl and \p dt = 0.
+ *       If \p t is an exact match on a keyframe's time tt,
+ *       then \p t0 = \p t1 = tt and \p dt = 0.
+ */
+template <typename T>
+void findRange( const typename KeyFramedValue<T>::KeyFrames& keyframes,
```
nmellado, line 28: I think this function should be moved in a class called `KeyFrames` (I would rename it as `KeyFrameCollection`), which would inherit `std::map` (would replace the `typedef` you have in `KeyFramedValue`) and add this specific lookup method.

src/Core/Animation/KeyFramedValueInterpolators.hpp
hoshiryu: If we don't find `t` in `keyframes` then we will have `it == keyframes.end()`, so your proposal must not work as expected

nmellado: This is case is detected few lines before
https://github.com/hoshiryu/Radium-Engine/blob/keyframedvalue/src/Core/Animation/KeyFramedValueInterpolators.hpp#L51

hoshiryu: what I meant is that when we come to this line, `it == keyframes.end`, so your proposal doesn't work since it will be the same as 
```std::upper_bound( keyframes.end(), keyframes.end(), t );```

Comment (hoshiryu) [reply src/Core/Animation/KeyFramedValueInterpolators.hpp:59]: what I meant is that when we come to this line, `it == keyframes.end()`, so your proposal doesn't work since it will be the same as  
```std::upper_bound( keyframes.end(), keyframes.end(), t );```

File: src/Core/Animation/KeyFramedValueInterpolators.hpp [hunk 3]
```diff
@@ -0,0 +1,128 @@
...
+}
+
+/** \name Predefined KeyFramedValue interpolators.
+ * Linearly interpolates the keyframes right before and right after time \p t.
+ * \note If there is no keyframe before \p t, then returns the first keyframe.
+ * \note If there is no keyframe after \p t, then returns the last keyframe.
+ */
+/// \{
+template <typename T>
+inline T linearInterpolate( const typename KeyFramedValue<T>::KeyFrames& keyframes, Scalar t ) {
```
dlyr, line 84: This seems inconsistent with the keyframe ctor that have an interpolator as parameter.
I prefer this version : KeyFramedValue do not know how to be interpolate, but just handles the values. Then linearInterpolate functions do the job.

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,212 @@
...
+    /// The type for interpolators.
+    using Interpolator = std::function<VALUE_TYPE( const KeyFrames& /*frames*/, Scalar /*t*/ )>;
+
+    /**
+     * Creates a KeyFramedValue from a first keyframe.
+     * \param frame the first keyframe value.
+     * \param t the first keyframe point in time.
+     * \param interpolator the function used to interpolate between keyframes.
+     */
+    KeyFramedValue( VALUE_TYPE frame,
```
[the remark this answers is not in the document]
hoshiryu: I am not sure either. What if you want to implement a specific interpolator that takes as additional parameter some value? Then having it as part of the KeyFramedValue "hides" it from callers who might not be interested in how you interpolate your values.
b.t.w. one can get/set the interpolator if needed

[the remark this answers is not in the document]
nmellado: I think I agree with @dlyr on the fact that changing the interpolator might be a desirable thing, and templating the `.at` method makes sense.
> What if you want to implement a specific interpolator that takes as additional parameter some value?

Do you have something specific in mind ? I can't find any use case for this.

For me, the data (`KeyFramedValue`) and it usage (`Interpolator`) must be kept separated, it is the job of the caller to ensure that the interpolator used for computation is the same than the one used in UI.

src/Core/Asset/AnimationTime.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * Return the animation end time in \p end.
+     */
+    inline void setEnd( const Time& end ) { m_end = std::max( end, m_start ); }
+
+    /**
+     * Extends *this so that \f$ t \in [start,end] \f$.
+     */
+    inline void merge( const Time& t ) {
```
nmellado, line 58:
```suggestion
    inline void extends( const Time& t ) {
```

src/Core/Asset/AnimationTime.hpp
```diff
@@ -0,0 +1,143 @@
...
+        setStart( t );
+        setEnd( t );
+    }
+
+    /**
+     * Performs the merge of *this with \p time, resulting in *this
+     * being the smallest AnimationTime containing both *this and \p time.
+     */
+    inline void merge( const AnimationTime& time ) {
```
nmellado, line 67:
```suggestion
    inline void extends( const AnimationTime& time ) {
```

src/Core/Animation/KeyFramedValueInterpolators.hpp
hoshiryu: see answer above

src/Core/Asset/AnimationTime.hpp
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Return the animation duration.
+     * \note if the AnimationTime is invalid, returns 0.
+     */
+    inline Time duration() const { return isValid() ? m_end - m_start : 0; }
+
+    /**
+     * Return true if \f$ t \in [start,end] \f$.
+     */
+    inline bool contain( const Time& t ) const { return ( ( t >= m_start ) && ( t <= m_end ) ); }
```
nmellado, line 90:
```suggestion
    inline bool contains( const Time& t ) const { return ( ( t >= m_start ) && ( t <= m_end ) ); }
```

src/Core/Asset/AnimationTime.hpp
```diff
@@ -0,0 +1,143 @@
...
+ */
+using Time = Scalar;
+
+/**
+ * The AnimationTime class stores the start and end times of an animation.
+ */
+class RA_CORE_API AnimationTime
+{
+  public:
+    AnimationTime( const Time& start = std::numeric_limits<Time>::max(),
```
nmellado, line 23: Do not pollute the namespace..

```suggestion
/**
 * The AnimationTime class stores the start and end times of an animation.
 */
class RA_CORE_API AnimationTime
{
  public:
    using Time = Scalar;
    AnimationTime( const Time& start = std::numeric_limits<Time>::max(),
```

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
hoshiryu: I really dont' see why you would want it to be separated, it brings more work on client side to ensure the interpolator used is the same in all places for no reason.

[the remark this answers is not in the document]
nmellado: Because I want to be able to use two different interpolators at the same time.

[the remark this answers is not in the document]
dlyr: Ok, let's say the interpolator is something attached to the keyFramedValue,
What kind of interpolator do we think of ? what are the needed data ?

- Nearest
- Linear
- Bezier
- ? (slerp, hermit cubic ? are they usefull)

The needed data is value, in and out tangent, as far as I know. we can also think of a weight value.

I think of a use case like 
```
KeyFramedValue v;
v.insertValue(t0, v0);
v.insertValue(t2, v1, inTgt, outTgt);
v.insertValue(t2, v2);

v.setInterpolator(&LinearInterpolation)

v.value(time);
```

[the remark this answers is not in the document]
MathiasPaulin: +1 for templated at method

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+        return time;
+    }
+
+    /// \name KeyFrame Management
+    /// \{
+
+    /**
+     * \returns the \p i-th keyframe.
+     */
+    inline const KeyFrame& getKeyFrame( size_t i ) { return m_keyframes[i]; }
```
MathiasPaulin, line 119: Why not an operator [] ?

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+        { lower->second = frame; }
+        else
+        { m_keyframes.insert( upper, kf ); }
+    }
+
+    /**
+    ￼ * Inserts a keyframe at time \p t corresponding to the value interpolated at \p t.
+    ￼ */
+    inline void insertKeyFrame( const Scalar& t ) override final { insertKeyFrame( t, at( t ) ); }
+
```
MathiasPaulin, line 146: See comment above (rename refineKeyFrame ?) and, as for `at`, this method could be templated by the interpolator. One can refine a key-frame set using an interpolator different from the one used to sample values at time t. (eg. spline when refining, linear when sampling)

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+     *          false otherwise.
+     */
+    inline bool operator!=( const KeyFramedValue& keyframe ) const {
+        return !( *this == keyframe );
+    }
+    /// \}
+
+    /// \name Interpolator
+    /// \{
+
```
MathiasPaulin, line 191: Seems that this get/set is superfluous.

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * \returns the number of keyframes.
+     */
+    virtual size_t size() const = 0;
+
+    /**
+     * \returns the ordered list of the points in time where a keyframe is defined.
+     */
+    virtual inline std::vector<Scalar> getTimes() const = 0;
```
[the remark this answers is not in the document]
hoshiryu: well this was an helper function, but it could be removed

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Inserts a keyframe at time \p t.
+     */
+    virtual void insertKeyFrame( const Scalar& t ) = 0;
+
```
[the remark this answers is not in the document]
hoshiryu: Might be misnamed.
one might want to first add a keyframe for time t before adjusting it to its needs

src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -0,0 +1,215 @@
...
+    /**
+     * Removes the \p i-th keyframe.
+     * \returns true if the keyframe has been removed, false otherwise.
+     */
+    virtual bool removeKeyFrame( size_t i ) = 0;
```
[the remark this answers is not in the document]
hoshiryu: There must be at least one keyframe so that the KeyFramedValue is well defined, will enhance the doc

src/Core/Animation/KeyFramedValue.hpp
hoshiryu: can do

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
MathiasPaulin: Keeping this function could introduce some data incoherency between usage of the KeyFramedValues. Imagine you get a copy to, e.g. build an interface but refine only the original data making the interface to diverge from the original data. This will not be robust.
And if it is a helper function, it is then not mandatory. So, it should'nt be pure virtual. And as it can't be something else, remove it and let the user build its own helpers function.

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
MathiasPaulin: So, this is refine ... this method should be renamed.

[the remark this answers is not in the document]
dlyr: I think this function is usefull (it is present in 3D software as "set keyframe", that stores the current interpolated value at current time as a keyframe)
Maybe rename "addKeyFrameFromInterpolatedValueForTime" ... or something a bit less verbose.

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
MathiasPaulin: Sure you need at least some values for the interpolator to function. But the number will depend on the interpolator (1 for nearest, 2 for linear, 3 for cubic, ...). So, this is related on the way the interpolator uses the values, not to the values themselves. And an empty keyframed value is valid as long as we consider it is a collection of <time-value> pair even if it can't be used because it is empty.
So, the only case of failure will be on an empty collection, and this is better manage by a precondition/assertion.

src/Core/Animation/KeyFramedValueInterpolators.hpp
```diff
@@ -0,0 +1,125 @@
...
+ *       then \p i0 = \p i1 = the index of the last keyframe and \p dt = 0.
+ *       If \p t is an exact match on a keyframe's time tt,
+ *       then \p i0 = \p i1 = the index of the keyframe and \p dt = 0.
+ */
+template <typename T>
+void findRange( const typename KeyFramedValue<T>::KeyFrames& keyframes,
+                const Scalar& t,
+                size_t& i0,
+                size_t& i1,
+                Scalar& dt ) {
```
MathiasPaulin, line 33: Why ref to indices (can we modify the indices ? Is so, what about the others keyframevalues of the collection ?) and not const ref to the keyframes (or even non const if wa allow to modifyt some par of the keys?

src/Core/Animation/KeyFramedValueInterpolators.hpp
MathiasPaulin: This is not resolved. KeyFramedValues are just an arrangement (possibly with constraints) of <time/value> pairs that only manage the storage and the coherency of the data. I'm not sure a keyframedvalue must hold an interpolator.

src/Core/Asset/AnimationData.hpp
```diff
@@ -4,69 +4,102 @@
...
+    /**
+     * \returns the number of HandleAnimations.
+     */
     inline uint getFramesSize() const;
 
-    /// Return the HandleAnimation, i.e. the whole animation frames.
+    /**
+     * \returns the list of HandleAnimations, i.e. the whole animation frames.
+     */
     inline std::vector<HandleAnimation> getFrames() const;
```
MathiasPaulin, line 84: Not sure the name is well chosen. HandleAnimation own a KeyFramedValues, defining all the animation frames wrt a handle. This method should be renamed getAllTheAnimatedParameter (or something less verbose as said previously by @dlyr)

src/Core/Animation/KeyFramedValueInterpolators.hpp
hoshiryu: indices are out parameters, so is dt, since the goal of this method is to find them ;)

dlyr:
I think the discussion is quite hard to follow for the PR.
I found my old gist about the key frame design [here](https://gist.github.com/dlyr/63ee167bb1a540866ec9de36465f3ec1)
I suggest that reviewers make their own with the expected use of keyframes, so we can fix the design of keyframedvalue

src/Core/Animation/KeyFramedValueInterpolators.hpp
MathiasPaulin: so, at least, const ref ...

hoshiryu: why would you want const since we want to compute them?
using tuple is overdoing it

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
hoshiryu: this method was needed to get all the time points where a frame is defined without having to cast from the base class (which the ui container is built on) to the derived class (which is templated w.r.t. the `VALUE_TYPE`)

src/Core/Animation/KeyFramedValueInterpolators.hpp
hoshiryu: As of now, keyframes are stored in a vector, would you inherit from std::vector to add this method?

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
hoshiryu: made it templated

src/Core/Animation/KeyFramedValue.hpp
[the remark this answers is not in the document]
hoshiryu: renamed as insertInterpolatedKeyFrame

src/Core/Animation/KeyFramedValue.hpp
hoshiryu: done

src/Core/Asset/AnimationData.hpp
hoshiryu: done

hoshiryu:
Ready for review

MathiasPaulin:
@dlyr Here is my gist about the key frame design.
You can have a look [here](https://gist.github.com/MathiasPaulin/39cdf8f7f56eaeff95a4ad87f302b517)

MathiasPaulin:
> @dlyr Here is my gist about the key frame design.
> You can have a look [here](https://gist.github.com/MathiasPaulin/39cdf8f7f56eaeff95a4ad87f302b517)

Some discussion about design was added to the gist

MathiasPaulin:
After discussing with @dlyr on our gist, we converge toward a proposal that correspond to an MVC pattern and that exhibits : 
  - a low level `KeyFramedValues` management class (the Model)  that mainly consists in storage and basic functionalities such that `find_range`. The aim of this class is to add invariant over a vector so that it always store the keyframe ordered by time to have both contiguous storage and log time access to any part of the data and to implement in a very efficient way the API. This low level class is templated and exhibits an API similar to the one proposed in this PR or in the gist [here](https://gist.github.com/MathiasPaulin/39cdf8f7f56eaeff95a4ad87f302b517). `find_range` could be refactored to return a range that match the interpolator degree (by returning the interval in which lies the value `t` but also some neighbors so that one can build higher degree interpolators (see CatmullRom in the gist). For the API, and thanks to the structural binding allowed by C++, I prefer a functional approach rather than a procedural one (e.g. functions returning tuples, like in the gist instead of procedures taking "by reference" parameters) but this could be discussed and arguments are welcome.
  - a higher level `KeyFrameController` (the controller, see the gist) that encapsulate a `KeyFramedValues` and an interpolator to ease the use of coherent keyframe/interpolator by any user/programmer. Note that it could be interesting that such a controller only store a reference (or smart pointer to avoid some memory problems) to a `KeyFrameValues` so that the same data could be used by several controller or view. The controller might be connected to any view that interact with the controller using an "update" method and that get signaled of a result/data change by the use of a callback to, e.g. call a function that will used the interpolated value computed at a coordinate `t`. 

To ease the use of both the model and the controller, we can also define a collection of predefined functions for interpolation such that `nearest`, `lerp`, `slerp`, `catmullrom`, `hermite`, `polynomial`, ... (non exhaustive nor mandatory list).

hoshiryu:
Regarding the first point, it seems we are on the good road.
For the second, would it ressemble to the `KeyFrameManipulator` introduced in file Timeline.hpp in PR #542?

MathiasPaulin:
For the first point, you are on the good way now, just, try to not compute something that is not needed, prefer constant time operator than potentially linear time ones and try to have functional profiles instead of procedural ones. And don't forget to write a unit test of this part. (and not an integration test such in the plugin you references once). 
For the second point, yes, `KeyFrameManipulator` is somewhat similar to what we want but it has nothing to do in another part of Radium. The controller must be also in the Radium::Core and in this PR as it is directly related to KeyFramedValues and could be used independently of any Gui or Engine application. 
In the [gist](https://gist.github.com/MathiasPaulin/39cdf8f7f56eaeff95a4ad87f302b517), we proposed `KeyFrameController`. The view is associated to the controller by setting the callback. The view ask for an update on the model by calling the update(t) method on the controller, who interact with the model and signal the view of the result by calling the callback. In this example, there is a simple textual view but you can define more than one signal/callback between view and controller according to the need we have on Radium. 

Think of a MVC pattern implemented as a chain "View <--> Controller <--> Model" instead of being implemented as a loop "View --> Controller --> Model --> View"

nmellado:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- src/Core/Animation/KeyFramedValueInterpolators.hpp  2
- tests/unittest/Core/animation.cpp  47
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/STORM/Radium-Engine/pullRequest?prid=5415618&bid=17756107)

hoshiryu:
> just, try to not compute something that is not needed, prefer constant time operator than potentially linear time ones and try to have functional profiles instead of procedural ones.
@MathiasPaulin I don't undestand what you mean by that

Also, I am not sure about your design for the `KeyFrameController` vs my `KeyFrameManipulator` , we might need to discuss it live.

nmellado:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- src/Core/Animation/KeyFramedValueInterpolators.hpp  2
- tests/unittest/Core/animation.cpp  47
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/STORM/Radium-Engine/pullRequest?prid=5415618&bid=17756107)

nmellado:
@hoshiryu , please update the tests according to the discussion we had on slack.
Cheers

nmellado:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- src/Core/Animation/KeyFramedValueInterpolators.hpp  2
- tests/unittest/Core/animation.cpp  46
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/STORM/Radium-Engine/pullRequest?prid=5415618&bid=17756107)

nmellado:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- tests/ExampleApps/SimpleAnimationApp/minimalapp.cpp  5
- tests/ExampleApps/SimpleAnimationApp/main.cpp  3
- src/Core/Animation/KeyFramedValueInterpolators.hpp  2
- tests/ExampleApps/SimpleAnimationApp/minimalradium.hpp  1
- tests/ExampleApps/SimpleAnimationApp/minimalapp.hpp  2
- tests/unittest/Core/animation.cpp  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/STORM/Radium-Engine/pullRequest?prid=5415618&bid=17756107)

nmellado:
@dlyr and @MathiasPaulin , could one of you review the PR for confirmation ?

nmellado:
Will be rebased once CI is done.

nmellado:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- tests/ExampleApps/SimpleAnimationApp/main.cpp  3
- tests/ExampleApps/SimpleAnimationApp/minimalradium.hpp  1
- tests/unittest/Core/animation.cpp  7
- tests/ExampleApps/SimpleAnimationApp/minimalapp.cpp  5
- src/Core/Animation/KeyFramedValueInterpolators.hpp  2
- tests/ExampleApps/SimpleAnimationApp/minimalapp.hpp  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/STORM/Radium-Engine/pullRequest?prid=5415618&bid=17756107)

PR #541, opened 2020-04-24
Time system

hoshiryu:

The CoupledTimedSystem is reshaped as the TimeSystem, main system managing time related events and dispatching them to AbstractTimedSystems.
Radium's main application now registers a TimeSystem, so that every plugin needing time management can plug its own system on it (such as the Animation Plugin).

The Animation Plugin is in charge of time management, other plugins may have either to develop their own time management process or plug into this one, which is not a good design.

Time management is performed on application side.

* **Other information**:


hoshiryu:
can someone relaunch Travis please?

nmellado:
Yesterday we discussed this PR with @dlyr and @MathiasPaulin. 
We observe that the task system is supposed to handle dependencies between processes, and somehow the `TimedSystem` is trying to do the same. 

From our discussions and your code, I propose the following:


The struct `FrameInfo` offers a good place to expose the animation time: it is accessible by all the systems. As such, we propose to add a field `Scalar m_animTime{0};` to `FrameInfo`, and use the `TimedSystem` to control its values.

The current `FrameInfo` is transmitted as read-only to the systems when generating the tasks, but then ignored when executing the tasks. I propose to make it accessible and modifiable by the tasks (we may want to rename `FrameInfo` as `FrameContext` ?), e.g.:
 1. `FrameInfo` is not transmitted when generating the tasks anymore:
 ```cpp
/// System.hpp
class RA_ENGINE_API System
{
/// ....
virtual void generateTasks( Core::TaskQueue* taskQueue ) = 0;
};


/// RadiumEngine.cpp
void RadiumEngine::getTasks( Core::TaskQueue* taskQueue ) {
    for ( auto& syst : m_systems )
    {
        syst.second->generateTasks( taskQueue );
    }
}
FrameInfo RadiumEngine::getFrameInfo( Scalar dt ) {
    static uint frameCounter = 0;
    return FrameInfo frameInfo{dt, frameCounter++};
}
```
 2. Tasks now have write access to the `FrameInfo` during processing:
```cpp
class RA_CORE_API Task
{
  public:
    /// Destructor
    virtual ~Task() {}

    /// Return the name of the task.
    virtual std::string getName() const = 0;
    
    /// Do the task job. Will be called from the task queue threads.
    virtual void process(FrameInfo& frameInfo) = 0;
};
```
 3. `FrameInfo` is now given when processing the tasks:
```cpp
/// BaseApplication.cpp
m_engine->getTasks( m_taskQueue.get() );

// Run one frame of tasks
m_taskQueue->startTasks( m_engine->getFrameInfo( dt ) );
m_taskQueue->waitForTasks();
timerData.taskData = m_taskQueue->getTimerData();
m_taskQueue->flushTaskQueue();
```
The `FrameInfo` is stored in the task queue when calling `startTasks`, and it is the reference to this local copy which is shared and modified by the `Tasks`. The method `waitForTasks` could return the computed `FrameInfo`, for debug or monitoring purposes.

Following this proposal, the TimedSystem` must update the value in a dedicate task, and 
   3. any system needing to read the animation time must do it into a task that depends on `TimedSystem` one.

**Open question**: do we need `FrameInfo` for task generation, e.g. can the animation time change the type or the layout of the generated tasks ? This proposal does not allow this, we must find an alternative in that case.

Note that this proposal also allows to move all computations in the tasks, and thus (if needed only) to generate the tasks only once, and then tasks for successive frames without recomputing the dependencies.

hoshiryu:
From the quick discussion we had with @MathiasPaulin this morning, it seems better to have the `RadiumEngine` directly manage time rather than a `TimeSystem` doing so with attached sub-systems.

I am ok with points 1 and 3, but I would not allow systems to modify the `FrameInfo`.
Regarding the tasks layout, I can't find any example of it depending on the animation time, but in such a case we could let the dev manage that in the tasks (like adding all tasks, then activate only the ones for the given animation time).

Generating all tasks at once (and not every frame) seems promising since systems always do the same generation. But then we must allow systems to ask for a task layer rebuild <hen they need to (for example when a new Component is created - either at load or through UI - 
and the system needs to add tasks for it).

nmellado:
Ok, so, do we still need something in this PR ?

hoshiryu:
Should we keep this PR or open a new one with my proposal for time management in the `RadiumEngine` ?

PR #548, opened 2020-04-27
glbinding and globjects use the same compiler than Radium

### LINKED ISSUE #547: Unable to compile in debug mode on MacOsX
On  MacOs, glbinding and globjects are not compiled with the same compiler than the Radium libs. These two external libraries are compiled using the system default compiler (Apple Clang) whatever the compiler used by Radium (set using -DCMAKE_CXX_COMPILER= ...).

This results in an unexpected ABI conflict and link error related to CXX11 string ABI.

MathiasPaulin:
closes #547

Set the compiler for externals glbinding and globject to be the same than the one used  by Radium


hoshiryu:
Can you make sure that this is done for every external lib?

MathiasPaulin:
> This change must also be applied to all the other external packages.
> Please add this fix to this PR.

All others package are automatically compiled with the good compiler.

nmellado:
No, they are not: https://github.com/STORM-IRIT/Radium-Engine/blob/master/src/Core/external/CMakeLists.txt#L34

MathiasPaulin:
All other packages are compiled (in Release, Debug and RelWithDebInfo) with the good compiler, i.e. the CMAKE_CXX_COMPILER chosen for the main configuration file. 
Even cpplocate, that comes also from cginternals and has the same cmake configuration system, is well compiled. There is no need to tell them again. 
There is problem only for glbinding and globject, and only in debug mode, and I can't investigate why nor if replicable on several systems.

PR #554, opened 2020-06-03
Allow to use locally installed dependencies instead of the one bundled with Radium

### LINKED ISSUE #544: Use local external lib instead of cmake comfigured one
To reduce compile time, or to test specific version of an external lib, could it be possible to give an explicit external directory.
something like if LibDIR is provided during cmake config,  try to load the lib conf from this directory, if config fail, warn the user, if config succeed message the user that a local version is used, warning for compatibility.
if no LibDIR is provided, use ExternalProject_add, as done right now.

### LINKED ISSUE #613: Is there a way to fetch and install  in one command all the needed Radium deps ?
Is there a way to fetch and install  in one command all the needed Radium deps ? 
It could be fine to have a `CMakeLists.txt` somewhere (e.g. at the root of the 'external' folder) whose aim is only to fetch, compile and install (in any `CMAKE_INSTALL_PREFIX`) all the required dependencies in the correct version such that a user might install them at once, possibly outside of the Radium bundle, and use them in several Projects.

_Originally posted by @MathiasPaulin in https://github.com/STORM-IRIT/Radium-Engine/pull/554#discussion_r523929482_

MathiasPaulin:
First try to fix issue #544
closes #544
closes #613 

This PR is ready to review and merge now.

This PR aims at discussing on how to allow user to have their own local installation of Radium external dependencies instead of configuring/building/installing them with the Radium bundle.

In this first try, only Eigen3 is supported.
Beware that the version used in Radium is not the version found in several package managers (e.g. macport on Macos has the last released Eigen3 version and Radium use another tag)

The proposal succeed in using locally installed Eigen3 or Radium bundled Eigen3.

The specific way of doing this should be made more generic (perhaps)

When configuring the project, use the option of cmake `-DEigen3_DIR = /path/to/local/Eigen3/cmake` or set all your local dependencies into a file (see localDependencies.cmake at the project root) and use the cmake option "-C /path/to/localDependencies.cmake" to populate an initial cache.

Lot of actions remains : 
  - make the process more generic
  - document how to use locally installed dependencies (and update all the Radium doc to be clear on the version/tag needed)
  - Improve the process to detect the version of the local dependency proposed by the user. If the version is not the good-one, use Radium external dependency process after signaling the user (or stop on an error with a clear message)
 - Add all the others dependencies


cmake/externalFunc.cmake
```diff
@@ -2,7 +2,7 @@ set( CMAKE_EXECUTE_PROCESS_COMMAND_ECHO STDOUT)
 
 macro(addExternalFolder NAME FOLDER )
     # External resources/repositories are downloaded and built at configuration stage
-    message(INFO "[addExternalFolder] process ${NAME} ${FOLDER}")
+    message("[addExternalFolder] process ${NAME} ${FOLDER}")
```
dlyr, line 5:
```suggestion
    message(STATUS "[addExternalFolder] process ${NAME} ${FOLDER}")
```

localDependencies.cmake
```diff
@@ -0,0 +1,6 @@
+# this file set variables pointing to local installation of dependencies.
```
dlyr, line 1: I don't think this file is needed, it's part of the cmake usage, and depends on the user setup.
We may state that in the documentation, but don't need to include it in Radium sources.

src/Core/CMakeLists.txt
```diff
@@ -31,16 +31,21 @@ add_library(
     ${core_inlines}
     ${RA_VERSION_CPP}
 )
-addExternalFolder( Core ${CMAKE_CURRENT_SOURCE_DIR}/external )
+
+set(LocalDependencies)
+if (DEFINED Eigen3_DIR)
+    set(LocalDependencies "-DEigen3_DIR=${Eigen3_DIR}")
+endif ()
+addExternalFolder(Core ${CMAKE_CURRENT_SOURCE_DIR}/external ${LocalDependencies})
```
dlyr, line 39: don't get the point here, (but I do not try without) 
should Eigen3_DIR def be propagated to addExternalFolder automatically ?
Is it to give Eigen3_DIR to other dependencies ? if so how it is working currently without this def ?

MathiasPaulin: As `addExternalFolder` creates a cmake sub-process, the current cache is not propagated to the child process. So, we must pass explicitely somme additional args to this cmake child process. 
The macro `addExternalFolder` (i do not change the macro)  takes any number of parameters. The first two are used internally by the macro, the others are passed to the cmake child process. This is the only way I found to give the child cmake process (which configure and build the externals) informations about locally installed package that it might build or that it depends on.

localDependencies.cmake
MathiasPaulin: I agree.
For now, it is here just for the discussion and some tests, it is easier to have just 1 parameter to the cmake command line than several ones (and less error prone)). It will be removed when the PR wil be in final state.

cmake/externalFunc.cmake
MathiasPaulin: OK for that. But, according to the cmake message documentation (https://cmake.org/cmake/help/v3.13/command/message.html), thereis a difference between `none` and `STATUS` for the message type. Even more on current cmake version (https://cmake.org/cmake/help/latest/command/message.html).
This is clearly out of scope of the current PR/discussion but we perhaps have to be more coherent on our cmake `message` usage and identification.

localDependencies.cmake
nmellado: A proper way to that with modern cmake is to use Toolchain files, which contains all the configuration option that you want to locally set to cmake. If we want to push this approach, we may have a default toolchain file which uses dependencies loaded dynamically by cmake, and maybe one or two examples of libraries on standard platforms (e.g., vanilla ubuntu).

nmellado (commented):
Another approach (maybe complementary) to this proposal would be to allow the user to specify a cache directory, where the dependencies would be downloaded. If this cache dir is out of the build dir, it can be shared across several builds.

localDependencies.cmake
MathiasPaulin: I'm not sure that TOOL_CHAIN_FILE is the proper way to do that. It is intended for cross-compilation if I understand well the doc : 
> This variable is specified on the command line when cross-compiling with CMake.

dlyr: Well, it's up to the user to define how he wants the different options, with tool chaine, with -C FILE, or directly with -D.
So I propose : we do not say anyting about that in the doc, just document the options.
I have finished on Core, create a script that tests the different combinations of local installed vs fetch... seems to work (but the script is tailored on my own install, not for sharing).

dlyr:
FYI guideline check fail is normal, and updated according to the change in external structure

doc/developer/buildchain.md
```diff
@@ -89,3 +89,40 @@ OPTION( RADIUM_SKIP_${NAME_UPPER}_EXTERNAL "[addExternalFolder] Skip updating ${
...
+ *  glm\_DIR
+ *  glbinding\_DIR
+ *  globjects\_DIR
+ *  stb\_DIR
+ *  Eigen3\_DIR
+ *  OpenMesh\_DIR
+ *  cpplocate\_DIR
+
+Radium in compiled and tested with specific version of dependencies, as given in the external's folder CMakeLists.txt and state here for the record
+
```
nmellado, line 111: Could’t we use snippets from the cmake files directly ?

doc/developer/buildchain.md
```diff
@@ -89,3 +89,40 @@ OPTION( RADIUM_SKIP_${NAME_UPPER}_EXTERNAL "[addExternalFolder] Skip updating ${
 ~~~
 where `$NAME_UPPER` is `CORE`, `ENGINE`, `GUIBASE`, `PLUGINBASE` or `IO`.
 \note The option `RADIUM_SKIP_${NAME_UPPER}_EXTERNAL` is generated only if `RADIUM_GENERATE_LIB_${NAME_UPPER}` is `ON`.
+
+## User provided external dependencies
+
+You can  use your own installation of a local dependency instead of letting cmake fetch and comiple it.
+To this end, just provide the corresponding '*_DIR' to cmake at configuration time (with '-D' option, configuration file or toolchain file, please refer to cmake documentation).
+
+Currently supported:
```
nmellado, line 98:
```suggestion
Currently supported (note that these path must refer tot installation the directory of the corresponding library):
```

doc/developer/buildchain.md
```diff
@@ -89,3 +89,40 @@ OPTION( RADIUM_SKIP_${NAME_UPPER}_EXTERNAL "[addExternalFolder] Skip updating ${
...
+
+ *  assimp\_DIR
```
nmellado, line 100:
```suggestion
 *  `assimp_DIRˋ
```

nmellado (commented):
First round of comments.
Looks great anyway !

doc/developer/buildchain.md
dlyr: not sure what do you mean ? including the cmake files here ? actually I've done a python script to generate this text from the cmake file.

cmake/radium_setup_functions.cmake
```diff
@@ -761,3 +761,15 @@ function(configure_radium_library)
...
 
+
+
+# Simple macro to populate LocalDependecies variable according to cmake args
+# NAME is the name of the variable to test, e.g. "Eigen3_DIR"
+macro(populateLocalDependencies)
+    set(oneValueArgs NAME)
+    cmake_parse_arguments(CHECKDEP "DUMMY_OPTION" "${oneValueArgs}" "DUMMY_MULTI" ${ARGN} )   
+    if (${CHECKDEP_NAME})
+        set(LocalDependencies "-D${CHECKDEP_NAME}=${${CHECKDEP_NAME}}")
```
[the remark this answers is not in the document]
dlyr: right, in next update

doc/developer/buildchain.md
nmellado: Ok fine then 👍

doc/basics/compilation.md
```diff
@@ -9,7 +9,7 @@ All the dependencies are automatically fetched and compiled at build time.
  * [GuiBase] Qt Core, Qt Widgets and Qt OpenGL v5.5+ (5.10 recommended)
  * stb_image
 
-See @ref develbuildchain for technical details on dependencies management.
+See @ref develbuildchain for technical details on dependencies management and how to provide locally installed dependencies.
```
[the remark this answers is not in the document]
nmellado: Hum good idea... 
But, ... can we keep that for another PR ? I mean the current one is large enough, and what you suggest is to add another functionality, so can be done after.

nmellado:
@MathiasPaulin  regarding this https://github.com/STORM-IRIT/Radium-Engine/pull/554#discussion_r523925771 about snippets and oxygen: while this is easy for code files, I am not sure doxygen will be able to produce similar results for cmake files.

doc/developer/buildchain.md
nmellado: While this is easy for code files, I am not sure doxygen will be able to produce similar results for cmake files.

dlyr: well, feel free to try, the script is below (see scripts/list_dep.py)
if you find a easy way to obtain something similar using snippets in cmakelists without redundancies.

MathiasPaulin: OK, just stay like this. Even if doxygen handle cmake scripts quite well, it is not so crucial to add cmake snipets. The script does the job well.

doc/basics/compilation.md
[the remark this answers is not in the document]
MathiasPaulin: I think we can do that on this PR. Just 1 file more (and a small documentation adds). I'll try to do that this morning.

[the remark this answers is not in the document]
dlyr: Well, it's not so hard, but then we will have to maintain two files when doing update of dependencies ...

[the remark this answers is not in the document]
MathiasPaulin: I've written a `CMakeLists.txt` that just include the one in Core, Engine and IO. 
As far as I know, we can update the dependencies of Radium without changing something in this file (but only in external/Core, external/Engine and external/IO), so, this will not change the work to do when updating a dependency.
This file will be alone in one commit, so that, if you don't agree, we can remove this commit.
I'm testing for now and I'll push very quickly.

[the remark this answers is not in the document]
nmellado: As I said I prefer that we add this in a dedicated PR.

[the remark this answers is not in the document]
MathiasPaulin: > As I said I prefer that we add this in a dedicated PR.

Give a look on the commit 4d49506878bf6f337a440e8b8bf518484751805d and feel free to remove it if you think it is too intrusive. 
I do not think so but ...
I tried it with all or only some of the dependencies, everything goes fine.

external/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+      message( "Set default build type to ${CMAKE_BUILD_TYPE}" )
+  endif()
+
+  #-----------------------------------------------------------------------------
+
+  # Set default install location to RADIUM_EXTERNAL_INSTALL
+  if (CMAKE_BUILD_TYPE STREQUAL "Release")
+      set(RADIUM_EXTERNAL_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/RadiumExternals-${CMAKE_CXX_COMPILER_ID})
+  else ()
+      set(RADIUM_EXTERNAL_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/RadiumExternals-${CMAKE_CXX_COMPILER_ID}-${CMAKE_BUILD_TYPE})
```
dlyr, line 53:
```suggestion
      set(RADIUM_EXTERNAL_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/Bundle-Externals-${CMAKE_CXX_COMPILER_ID}-${CMAKE_BUILD_TYPE})
```

external/CMakeLists.txt
```diff
@@ -0,0 +1,73 @@
...
+  if (NOT CMAKE_BUILD_TYPE)
+      set (CMAKE_BUILD_TYPE Release )
+      message( "Set default build type to ${CMAKE_BUILD_TYPE}" )
+  endif()
+
+  #-----------------------------------------------------------------------------
+
+  # Set default install location to RADIUM_EXTERNAL_INSTALL
+  if (CMAKE_BUILD_TYPE STREQUAL "Release")
+      set(RADIUM_EXTERNAL_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/RadiumExternals-${CMAKE_CXX_COMPILER_ID})
```
dlyr, line 51:
```suggestion
      set(RADIUM_EXTERNAL_INSTALL ${CMAKE_CURRENT_SOURCE_DIR}/Bundle-Externals-${CMAKE_CXX_COMPILER_ID})
```

nmellado (approved):

nmellado:
Approved.
@dlyr please fix the checks, and we are good to go.
Thanks !

dlyr:
> FYI guideline check fail is normal, and updated according to the change in external structure

Well, the error in checks are because the script refer removed files, so the PR propose a fix of the check ;)

nmellado:
Hum m’y bad, I didn’t catch that...

nmellado (approved):

PR #556, opened 2020-06-08
[CI] Clean cmake for externals

nmellado:

Change cmake options in travis for shorter logs + to avoid inconsistencies between Radium and external cmake options.

1. `CMAKE_INSTALL_MESSAGE` is set to default value, which prints update logs when installing twice. This produces very long log when installating the documentation
https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_MESSAGE.html
2. the options passed to cmake for Radium are only propagated to externals, manually for each external project.

1. Set `CMAKE_INSTALL_MESSAGE` to `LAZY` to avoid printing update messages.
2. Save the cmake options that need to be propagated to external using an intermediate cmake file shared between Radium project and the externals (`cmake/ExternalInclude`). This file populates a variable `RADIUM_EXTERNAL_CMAKE_OPTIONS` with current cmake option, that can be passed directly to `ExternalProject_add`.

1. Minor change on CI.
2. Now the externals are built with the same default variables than Radium (e.g. `DCMAKE_CXX_STANDARD`). Depending of the externals, this variables are either used as is, or overridden internally. At least we know that we suggested the same compilation environment that the one used to compile Radium.

* **Other information**:


nmellado:
Ready to review.

nmellado:
We can also set `-DCMAKE_INSTALL_MESSAGE` to `NEVER` when compiling dependencies, what do you think ?

MathiasPaulin:
> We can also set `-DCMAKE_INSTALL_MESSAGE` to `NEVER` when compiling dependencies, what do you think ?

If one can nevertheless have a rough information about the dependencies building progression, yes, it is a good idea.

nmellado:
It is just affecting the installation, not the compilation.
We can also add an option to control this for externals, and set it by default as `NEVER`

nmellado:
@dlyr I can't see the impact of `RADIUM_EXTERNAL_CMAKE_INSTALL_MESSAGE` on the CI, as you have a good connection, could you try locally ?

dlyr:
@nmellado what should I do ?

nmellado:
@dlyr nothing I am on it thanks

nmellado:
Ready to review, I have extended a bit the scope of the PR and updated the description accordingly

nmellado:
Now ready for review and (hopefully) merge

cmake/externalFunc.cmake
```diff
@@ -1,8 +1,10 @@
 set( CMAKE_EXECUTE_PROCESS_COMMAND_ECHO STDOUT)
 
+include(ExternalInclude)
+
 macro(addExternalFolder NAME FOLDER )
     # External resources/repositories are downloaded and built at configuration stage
-    message(INFO "[addExternalFolder] process ${NAME} ${FOLDER}")
+    message(VERBOSE "[addExternalFolder] process ${NAME} ${FOLDER}")
```
MathiasPaulin, line 7: The `VERBOSE` message mode was introduced in cmake 3.15. But the cmake minimal version set in the main `CMakeLists.txt` is 3.13. On several `CMakeLists.txt`, the minimum version is even set to 3.6 ... Should be useful to have coherent cmake minimum version along al the project and to ensure functionalities used are supported.

cmake/ExternalInclude.cmake
```diff
@@ -0,0 +1,19 @@
+set(RADIUM_EXTERNAL_CMAKE_OPTIONS
+    -DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}
+    -DCMAKE_CXX_STANDARD_REQUIRED=${CMAKE_CXX_STANDARD_REQUIRED}
+    -DCMAKE_GENERATOR=${CMAKE_GENERATOR}
+    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
+    -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX}
+    -DCMAKE_INSTALL_MESSAGE=${RADIUM_EXTERNAL_CMAKE_INSTALL_MESSAGE}
```
MathiasPaulin, line 7: At the first run on any external folder, this makes cmake to emit a warning message : 
```
CMake Warning:
    Manually-specified variables were not used by the project:
       CMAKE_INSTALL_MESSAGE
```

cmake/externalFunc.cmake
```diff
@@ -51,12 +41,12 @@ macro(addExternalFolder NAME FOLDER )
         string(TOLOWER ${CMAKE_GENERATOR} generator_lower)
         if (MSVC AND NOT ${generator_lower} STREQUAL "ninja")
             set(RadiumExternalMakeTarget ALL_BUILD)
-            message(INFO "[addExternalFolder] Enable compatibility mode for VS Generator")
+            message(VERBOSE "[addExternalFolder] Enable compatibility mode for VS Generator")
```
MathiasPaulin, line 44: I found  it's a good idea to differentiate messages emit by the build chain. For now, except those few `VERBOSE`, all messages are set to `STATUS` in the various cmake files we used. It could be helpful to ensure this PR is efficient to define a clean policy about the messages mode and to express this policy in the developer doc along Radium programming guidelines.

cmake/ExternalInclude.cmake
```diff
@@ -1,3 +1,14 @@
+#set the compile defdinition for current directory (i.e. externals)
+if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  add_compile_options(/w )
+else()
+  add_compile_options(-w)
+endif()
+
+#create a sepefic options for the external build
+get_directory_property(EXTERNAL_FLAGS COMPILE_OPTIONS)
```
MathiasPaulin, line 9: Where is this directory property set ?

CMakeLists.txt
```diff
@@ -28,6 +32,8 @@ option(RADIUM_GENERATE_LIB_PLUGINBASE "Include Radium::PluginBase in CMake proje
 option(RADIUM_UPDATE_VERSION          "Update version file each time the project is compiled (update compilation time in version.cpp)" ON)
 option(RADIUM_INSTALL_DOC             "Install documentation. If RadiumDoc is compiled, install documentation to bundle directory for install target" ON)
 
+set(RADIUM_EXTERNAL_CMAKE_INSTALL_MESSAGE NEVER)
```
MathiasPaulin, line 35: When installing externals, the install messages are still there. (installing, updating, ...). I tested on Linux with cmake 3.16.3

cmake/externalFunc.cmake
nmellado: I changed it as INFO does not exists. I can use any other.
The other changes you require are out of the scope of this PR, and can be fixed in another PR

cmake/ExternalInclude.cmake
nmellado: Yes, but there is no typo and this option exists in the doc. It is however rarely used and maybe it’s implementation is not available on all version (the option has been introduced in cmake 3.1, and is still available).
So I suggest to keep it.

cmake/ExternalInclude.cmake
nmellado: `COMPILE_OPTIONS` is a cmake variable.

CMakeLists.txt
nmellado: Yes, see my other comment about the option.

cmake/externalFunc.cmake
nmellado: Out of the scope.
This PR fixes cmake in order to remove the  errors that make the CI broken on master and radium-release. If we want a message policy we can do in another PR. 
We must stop adding more and more content to any PR submitted to Radium.

cmake/ExternalInclude.cmake
dlyr: line 3 or 5

cmake/externalFunc.cmake
MathiasPaulin: `INFO` mode has never existed.
See cmake doc about message :
https://cmake.org/cmake/help/v3.13/command/message.html for cmake version 3.13
or
https://cmake.org/cmake/help/v3.15/command/message.html for cmake version 3.15
3.15 and later allow much more message modes

And I think that cmake-minimum version check and coherency, even if not the initial goal of this PR could be fixed in the PR as modifications will be only on some already modified files.

cmake/ExternalInclude.cmake
MathiasPaulin: I do not ask to remove `CMAKE_INSTALL_MESSAGE `but to find a way to silent this Warning. 
Having something in red that appears in a small log can be very distuubing if one think that everything goes well. And more, if this is not related to the user (but to the developer of the buildchain), why to perturb the user with such messages ? Perhaps there is a cmake policy that silent this ...

cmake/externalFunc.cmake
MathiasPaulin: this PR introduce a way to choose which level of message log to display, then allowing to shorten the log on the CI, then allowing the CI to success instead of stopping with "log too long" message. So, we might have coherent message mode to allow user to choose the log level they want instead of either printing everything (because every message is STATUS except those few) or nothing but the errors.

CMakeLists.txt
dlyr: well it seems there still is an issue here. If I run install in the external project subdir (e.g. /build/src/Core/external/OpenMesh-prefix/src/OpenMesh-build) with CMAKE_INSTALL_MESSAGE to NEVER (cmake -DCMAKE_INSTALL_MESSAGE=NEVER . && make install ), I get the expected behavior, but not if I run from the external subdir ( e.g. build/Core/external ), as there is no install target, just OpenMesh-install, don't know if it's linked.
BTW after reconfiguring the external in its build dir, there is no install messages ... maybe something went wrong during configuration.

cmake/ExternalInclude.cmake
MathiasPaulin: COMPILE_OPTIONS is not a cmake variable but a property name. but yes, it is set by `add_compile_option` and i missed it. but has it does nothing on my runs, i suspected a not set property

nmellado:
Thanks @dlyr !

dlyr:
Yes, there still is some overriding from /w and /W4 but some externals sets their own debug level.

nmellado:
I am fine with the current version. We need review/approvals for merging.

PR #559, opened 2020-06-16
update documentation about Rendering without using Materials

MathiasPaulin:
This PR updates the documentation on how to define shader parameters when rendering without using materials.

- [X] Docs have been added / updated (for bug fixes / features)


doc/developer/material.md
```diff
@@ -407,23 +407,43 @@ builder.second( rt, isMaterialTransparent );
...
-_Bidirectional Scattering Distribution function (BSDF)_.
+The _Radium Material Library_ and related components are mainly designed to manage Materials as a representation of a _Bidirectional Scattering Distribution function (BSDF)_.
 
-When rendering, it is sometime useful to compute the final color of an object that do not rely on a bsdf but just 
-on a specific color for each geometry fragment.
+When rendering, it is sometime useful to compute the final color of an object that do not rely on a bsdf but just on a specific color for each geometry fragment.
 
-Even if the Ra::Engine::PlainMaterial could be used for this and assuming that such a way to define material is object 
-specific, the following steps are required :
+Even if the Ra::Engine::PlainMaterial could be used for this and assuming that such a way to define material is object specific, the following steps are required :
```
nmellado, line 414: It is not clear to me if I can use `PlainMaterial` for this, or not.
Do you mean that it is not recommended ? or that it cannot be used ?

MathiasPaulin: I add this line because i had the following question from another reviewer : "Why allowing to render without material. I can do the example using PlainMaterial" ...  But doing such is an ugly hack even if the following example is similar in its computation and limitation than the plain material. Feel free to propose the sentence that will make clear that such hacks must be prevented.

dlyr: I'm not sure to catch the point here, what is the good way for one to have some specific rendering computation.
Let's say I'm doing a display shader to display an hdr texture on a quad for debug purposes. 
I want to control some simple parameter like color scale.
Is it a good idea to define a material ? or should I use the materialess approach explained here ?

dlyr: (and I do not find the reference to the other reviewer question in history ... if you have a link it could help to understand the changes).

doc/developer/material.md
```diff
@@ -407,23 +407,43 @@ builder.second( rt, isMaterialTransparent );
...
+  ~MyParameterProvider() {}
+  void updateGL() override {
+    // Method called before drawing each frame in Renderer::updateRenderObjectsInternal.
+    // The name of the parameter corresponds to the shader's uniform name.
+    m_renderParameters.addParameter( "aColorUniform", m_colorParameter );
+    m_renderParameters.addParameter( "aScalarUniform", m_scalarParameter );
+  }
+
+  void setOrComputeTheParameterValues() {
+	// client side computation of the parameters, e.g.
```
dlyr, line 437: After some experimentation, why do we need to have double storage of m_colorParameter.
Could this method directly set `m_renderParameter.addParameter("aColorUniform", value)` and the updateGL do nothing ...

nmellado: Hum, I guess we could see this as a CPU/GPU transfer ?
I mean, we should be able to use material even without GL.
I didn't check the code, but maybe that's why this is needed.

dlyr: This class is a `ShaderParameterProvider` so let's say it's linked to GPU Shader.
The `ShaderParameterProvider` class stores parameters on the CPU side as uniform value, ready to be transfered.
We can think of a different design, but as is, Materials are already on the GPU side, even if they can give access to material values without a shader, they don't need to store the value twice.

MathiasPaulin: It is not mandatory, when you write a class that derive from `ShaderParameterProvider` to store the parameter twice. All the parameters can be stored only once in the `m_renderParameter` object and the UpdateGL method will do nothing (or only some render-state dependent actions). And this can simplify and improve the performance of many rendering operations.
Meanwhile, random access to a named value (a uniform) from `m_renderParameter` is O(log(n)) while accessing the same named value from an object attribute is O(1). As there are not so many parameters for a shader, it will not be so important but no one know what can happen.
This was also designed to allow different representation between CPU and GPU version of a parameter. This is clearly the case for textures.
If you're refactoring RenderParameter to store also the type (vec3, float, ...), definition domain ([-2, 5], RGB subdomain, ...) and semantic (vector, point, color, ...) into the association name<->value while allowing constant time access, all the Engine classes inheriting from `ShaderParameterProvider` might be refactored as well to remove this double storage. In the current Engine implementation, the CPU version is used to load/save the assets or to edit/change a value while ensuring some constraints related to definition domain or relations between parameters (e.g. energy conservation for bsdf)

PR #562, opened 2020-06-18
Move time management to RadiumEngine ;

hoshiryu:

Removes `TimedSystem` and moves time management into the `RadiumEngine`.

The Animation Plugin is in charge of time management, other plugins may have either to develop their own time management process or plug into this one, which is not a good design.

 Time management is performed on Engine side.

* **Other information**:
Working with PR time_system on Radium-Apps and PR time_system on SkeletonBasedAnimation


src/Engine/RadiumEngine.hpp
```diff
@@ -171,6 +171,83 @@ class RA_ENGINE_API RadiumEngine
...
+     */
+    void setEndTime( Scalar t );
+
+    /**
+     * \Returns the `end` time for the time window.
+     */
+    Scalar getEndTime() const;
+
+    /**
+     * Toggles the PingPong mode.
```
MathiasPaulin, line 225: What is "the PingPong" mode ? 
If I understood well the code, and even if longer, I suggest to use instead ForwardBackward

src/Engine/RadiumEngine.hpp
```diff
@@ -171,6 +171,83 @@ class RA_ENGINE_API RadiumEngine
      */
     Core::Aabb computeSceneAabb() const;
 
+    /// \name Time Management.
+    /// {
+
+    /**
+     * Toggles effective time flow or constant time flow.
```
MathiasPaulin, line 178: Even if experienced users will understand this comment, give a little explanation of what is "effective time" or "constant time" and state the relation of these with the "RealTime" name of the method.

src/Engine/RadiumEngine.hpp
```diff
@@ -171,6 +171,83 @@ class RA_ENGINE_API RadiumEngine
...
+     */
+    void setRealTime( bool realTime );
+
+    /**
+     * Sets the time delta between two frames for constant time flow.
+     */
+    void setConstantTimeFlowDt( Scalar dt );
```
MathiasPaulin, line 185: `setConstantTimeStep` instead of  `setConstantTimeFlowDt`?

src/Engine/RadiumEngine.hpp
```diff
@@ -198,6 +275,16 @@ class RA_ENGINE_API RadiumEngine
 
     /// For internal resources management in a filesystem
     std::string m_resourcesRootDir;
+
+    Scalar m_dt;                ///< The time delta between 2 consecutive frames.
+    Scalar m_startTime{0_ra};   ///< The `start` time for the time window.
+    Scalar m_endTime;           ///< The `end` time for the time window.
+    Scalar m_time{0_ra};        ///< The current time.
```
MathiasPaulin, line 282: It could be more easy to maintain if strongly related members are packed into a local struct. So the above time related members might be packed together ?

src/Engine/RadiumEngine.hpp
```diff
@@ -198,6 +275,16 @@ class RA_ENGINE_API RadiumEngine
...
+    bool m_realTime{false};     ///< Whether we use the effective time flow or the constant one.
+    bool m_play{false};         ///< Shall time flow.
+    bool m_step{true};          ///< Shall time flow for only one frame.
+    bool m_pingPongMode{false}; ///< Is PingPong mode enabled.
```
MathiasPaulin, line 286: All of the above time related members might be packed together into an internal struct TimeMode or something similar ?

src/Engine/FrameInfo.hpp
```diff
@@ -7,9 +7,15 @@ namespace Ra {
 namespace Engine {
 /// Structure passed to each system before they fill the task queue.
 struct RA_ENGINE_API FrameInfo {
+    /// The current time.
+    Scalar m_time{0};
+
     /// Time elapsed since the last frame in seconds.
     Scalar m_dt{0};
 
+    /// The frame corresponding to the current time.
```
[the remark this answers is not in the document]
hoshiryu: This is just a remnant from tests

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,84 @@ class RA_ENGINE_API RadiumEngine
...
+     * In Real-time mode, time flows according to the actually elapsed time
+     * between consecutive frames, while in constant mode time flows according
+     * to a fixed time step.
+     */
+    void setRealTime( bool realTime );
+
+    /**
+     * Sets the time delta between two frames for Constant time flow.
+     */
+    void setConstantTimeStep( Scalar dt );
```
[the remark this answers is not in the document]
hoshiryu: I don't understand the meaning of `toggleConstant` here: do you mean "toggle mode", i.e. if true then false and vice-versa, or "set to value" ?
b.t.w. I don't see the point of returning the if "Constant time" mode is on

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,96 @@ class RA_ENGINE_API RadiumEngine
...
+    /**
+     * \brief Indicates if Constant-time time flow is on (Real-time is off) .
+     * \see isRealTime
+     */
+    bool isConstantTime() const;
+
+    /**
+     * \brief Sets the time delta between two frames for Constant-time time flow.
+     */
+    void setConstantTimeStep( Scalar dt );
```
nmellado, line 215: My point here is that you are not sure, when you call this function, that ConstantTime flow is enabled.
So I propose to add a boolean to force enabling contant time flow while setting the time step, to avoid that the user need to call two functions successively.

src/Engine/RadiumEngine.hpp
```diff
@@ -214,18 +304,40 @@ class RA_ENGINE_API RadiumEngine
...
+         */
+        void updateTime();
+
+        Scalar m_dt;                ///< The time delta between 2 consecutive frames.
+        Scalar m_startTime{0_ra};   ///< The `start` time for the time window.
+        Scalar m_endTime;           ///< The `end` time for the time window.
+        Scalar m_time{0_ra};        ///< The current time.
+        bool m_realTime{false};     ///< Whether we use the effective time flow or the constant one.
+        bool m_play{false};         ///< Shall time flow.
+        bool m_step{true};          ///< Shall time flow for only one frame.
```
nmellado, line 335:
```suggestion
        bool m_singleStep{true};          ///< Shall time flow for only one frame.
```
?

nmellado:
Please update on top of `master` branch next time you force push.

src/Engine/RadiumEngine.hpp
hoshiryu: hmm, so would you set the default value to `true` or `false`?
And why returning the activation state?

nmellado:
```c++
bool setConstantTimeStep( Scalar dt, bool forceConstantTime = false );
```
permet de garder un code lisible dans les cas généraux (et sans effet de bord) ou l'on souhaite simplement changer le timestep, ie:
```c++
foo.setConstantTimeStep( .1_ra );
```
Cela peut par être par exemple connecté à un wdget qui permet à l'utilisateur de changer le timestep.

Autre usage, un plugin de simulation qui veut lancer une simulation avec ses propres paramètres, et qui souhaite donc activer le mode constant:
```c++
foo.setConstantTimeStep( .1_ra, true ); // will always return true
```

Concernant la valeur de retour, elle peut être utile pour informer l'utilisateur de certains comportements:
```c++
if( ! foo.setConstantTimeStep( .1_ra ) )
   // log: warning: constantTimeStep set but RealTime mode is enabled
```

It is just a matter of design (very minor point) and I will not hang on this if others don't like it,.

hoshiryu: ok

nmellado:
I am fine with this version.
@dlyr and/or @MathiasPaulin, could one of you confirm (or not) approval ?

nmellado:
After reading carefully, something is maybe missing: how do ask the `Engine` to reach a given animation time ? 
This will happen when we will add the timeline, which should be able to have a bit of control on the Engine (not only play and pause).

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,106 @@ class RA_ENGINE_API RadiumEngine
...
+     * in the time window.
+     * \note If `end` time is negative, the ForwardBackward mode doesn't apply
+     *       and time just flows indefinitely.
+     */
+    void setForwardBackward( bool mode );
+
+    /**
+     * Toggles on/off time flow.
+     */
+    void play( bool isPlaying );
```
MathiasPaulin, line 237: If this method toggle a state, why does it take a parameter. If you keep the method as this, change the comment. Else, write 2 methods without parameters : `play` and `stop` without

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,106 @@ class RA_ENGINE_API RadiumEngine
...
+
+    /**
+     * Toggles time flow for one frame.
```
MathiasPaulin, line 240: Not sure to understand what is toggled.

src/Engine/RadiumEngine.hpp
```diff
@@ -214,18 +314,40 @@ class RA_ENGINE_API RadiumEngine
...
 
     /// Stack of saved fbo and viewport values @see pushFboAndViewport popFboAndViewport
     std::stack<FboAndViewport> m_fboAndViewportStack;
+
+    struct TimeData {
+        /**
+         * Update the current time from the current frame number, according to the
+         * time flow policy (ForwardBackward or loop around).
+         */
+        void updateTime();
```
MathiasPaulin, line 337: As said in the comment (and in the code of the method), if this method ensures some constraints (or invariant) between members, why are they public ? 
public members can be changed without ensuring any constraint (or invariant) between them.
If there is some constraints between members (and there are), put the constrained members (e.g. `m_time`) in a private section.
Even if this class is private to the Engine, it is important to ensure correct usage and data coherency.

hoshiryu: There is no specific constraint, the time window can be whatever, same for the modes.
Only *m_time*, which can still be whatever when time flow is on hold, has to be remapped into the time window when time flows. I don't see the point of having it private with setter and getter.

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,106 @@ class RA_ENGINE_API RadiumEngine
...
+     */
+    void step();
+
+    /**
+     * Sets time to \p t.
+     */
+    void goTo( Scalar t );
+
+    /**
+     * Resets time to 0, same as \code goTo( uint( 0 ) ); \endCode.
```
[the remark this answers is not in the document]
hoshiryu: just not updated the doc, actually resetting to start time

src/Engine/RadiumEngine.hpp
```diff
@@ -185,6 +185,106 @@ class RA_ENGINE_API RadiumEngine
...
+     */
+    void play( bool isPlaying );
+
+    /**
+     * Toggles time flow for one frame.
+     */
+    void step();
+
+    /**
+     * Sets time to \p t.
```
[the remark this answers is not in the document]
hoshiryu: Yes it is allowed as soon as time flow is stopped. When time flow is activated again, then the time will be mapped into the time window according to the flow mode

src/Engine/RadiumEngine.hpp
hoshiryu: well, we just allow time to flow for one frame

src/Engine/RadiumEngine.hpp
MathiasPaulin: I see one reason (from the code of the function update time) : `m_time` is always in the interval `[start, end]`. This is an invariant that must be ensured when setting time (or start, or end) and not at each update.

hoshiryu: This invariant has to be met only when time flows.
Let's say I have an animation from t=0 to t=10 (in seconds), but I just want to play it from t=2 to t=8 and edit that part.
I should still be able to see what happens outside [2,8] for convenience checks or whatever, for example using the keyframe from t=0 for t=7.
That's why I can set the time to any value, as soon as time flow is stopped.

PR #566, opened 2020-06-29
Fix topomesh compilation issue.

MathiasPaulin:
Alow to compile topomesh with older version than on CI of compiler

Compilation fix

error on non-instanciated template method returning ref to temporary

the template method compile and rais a static_assert if instantiated with an unsupported type.


src/Core/Geometry/TopologicalMesh.inl
```diff
@@ -409,14 +409,15 @@ GET_NAME_ARRAY_HELPER( Vector3, vector3 )
 GET_NAME_ARRAY_HELPER( Vector4, vector4 )
 
 #undef GET_NAME_ARRAY_HELPER
-
+// These template functions, for supported types are explicitly instanciated above. For unsupported
+// types, generates a compile error.
```
hoshiryu, line 413:
```suggestion
// These template functions are defined above for supported types.
// For unsupported types they simply generate a compile error.
```

PR #567, opened 2020-06-29
Add a way to access to the emissivity of a material.

MathiasPaulin:

This PR define a GLSL interface to access the emissivity of a surfacic material.

For now, emissive materials are not supported in Radium and there is a need, when compositing shaders, to access the individual components of a material : microgeometry, BSDF (that are already managed by the GLSL interface of Materials) and emissivity.

This PR just add the definition of the interface in the documentation and implement this interface in BlinnPhong and Lambertian GLSL implementation. (For now, these materials can't be emissive so the function returns black color)

The GLSL shaders are modified so that the emissivity of materials is taken into account in the zPrepass of the forward renderer.

Specific renderers will be able to access to the emissivity of the material to compute some effects like glow.

### DIFF
--- a/doc/developer/material.md
+++ b/doc/developer/material.md
@@ -286,7 +286,7 @@ vec3 getSpecularColor(Material material, vec3 texCoord);
 vec3 evaluateBSDF(Material material, vec3 texCoord, vec3 wi, vec3 wo);
 ~~~
 
-### Emissivity interface {#bsdf-interface}
+### Emissivity interface {#emissivity-interface}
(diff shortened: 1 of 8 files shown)


doc/developer/material.md
```diff
@@ -286,6 +286,15 @@ vec3 getSpecularColor(Material material, vec3 texCoord);
 vec3 evaluateBSDF(Material material, vec3 texCoord, vec3 wi, vec3 wo);
 ~~~
 
+### Emissivity interface {#bsdf-interface}
```
MathiasPaulin, line 289:
```suggestion
### Emissivity interface {#emissivity-interface}
```

PR #568, opened 2020-07-01
Fix cmake config scripts to support several helper libs

MathiasPaulin:
When configuring plugins that rely on several helpers lib, there is a typo in the cmake scripts that generates an error.
Fixed by using the good varaible.

build-system Bug fix

error when configuring a plugin that uses several helper libs

The plugin is configured with success


nmellado:
Thanks for the proposal, but could you explain why changing a variable name fixes the bug ? I don't understand what is happening...

MathiasPaulin:
In the cmake function `configure_radium_plugin`, there is a loop over helper libs to install them as dependencies
```
foreach (helperLib ${ARGS_HELPER_LIBS})
   do something with ${helperLib} # <-- This is a good usage of individual elements in the ${ARGS_HELPER_LIBS} list
   do something else with ${ARGS_HELPER_LIBS} # <-- this consider the list as a single element, that is wrong for some configurations
endforeach()
```
By using `${ARGS_HELPER_LIBS}`, the configuration failed when the list as several elements.
I just replace in this line (in fact 2 lines) `${ARGS_HELPER_LIBS}` by `${helperLib}`

Issue #570, opened 2020-07-01
BaseApplication derived apps (e.g. main-app) crash at exit when using some plugins


MathiasPaulin:
When an application derived from BaseApplication uses several plugins that manipulates RenderTechniques (e.g. in a specific renderer and a specific material library), it crash at exit due to deletion of already deleted Materials or parameter providers.
If the scene is cleared (all render objects deleted) before exiting, the application does not crash.
This is due to some unspecified order of deletion of dynamic loaded plugins and ownership transfer through shared pointer that is not always robust.

This could be reproduced through the use of 2 private plugins (GLTF and node based renderer) that works well when used as plugins separately or when linked to the application as a library but that make the application to crash when used simultaneously as plugins.
Both manage rendertechniques and parameter providers with shared ownership (std::shared_ptr) with the Engine.

MathiasPaulin:
Fixed by #673

PR #573, opened 2020-07-02
Painty

### LINKED ISSUE #572: Invalid state of default CoreGeometry in Engine::Mesh
Currently, it seems mandatory to call the method `loadGeometry` to initialize the geometry of an `Engine::Mesh`. 
However, by default the `Engine::Mesh` contains an empty object that can be edited directly. Problem: when doing this, the communication between `Core` and `Engine` seems to be broken (GL is not updated when updating the geometry).

Here is an example in two parts, where I have a method `init_geometry` that is called once.
First, the version that works:
```c++
void init_geometry() {
    Ra::Core::Geometry::TriangleMesh mesh;

    mesh.addAttrib( Ra::Engine::Mesh::getAttribName( Ra::Engine::Mesh::VERTEX_POSITION ), someVector3Array );
    mesh.m_indices= /* something */;

    m_engineMesh->loadGeometry( std::move( mesh ) );
}
```

And here is the version that doesn't work, ie. gl is not updated:
```c++
void init_geometry() {
    Ra::Core::Geometry::TriangleMesh & mesh = m_engineMesh->getCoreGeometry();

    mesh.addAttrib( Ra::Engine::Mesh::getAttribName( Ra::Engine::Mesh::VERTEX_POSITION ), someVector3Array );
    mesh.m_indices= /* something */;
    m_engineMesh->setIndicesDirty();
}
```

I would prefer the second option as it more compact and does not require `move`.
Also in term of semantics, it means that, currently, the `Engine::Mesh` is not completely initialized until `loadGeometry` is called.

dlyr:
may fix #572 
If works, I'll rework on the PR to have something more selfcontained.
only three last commits are relevant to the issue.

### DIFF
--- a/src/Engine/Component/GeometryComponent.cpp
+++ b/src/Engine/Component/GeometryComponent.cpp
@@ -50,9 +50,9 @@ TriangleMeshComponent::TriangleMeshComponent( const std::string& name,
                                               Entity* entity,
                                               Core::Geometry::TriangleMesh&& mesh,
                                               Core::Asset::MaterialData* mat ) :
-    GeometryComponent( name, entity ), m_displayMesh( new Engine::Mesh( name ) ) {
+    GeometryComponent( name, entity ),
+    m_displayMesh( new Engine::Mesh( name, std::move( mesh ) ) ) {
     setContentName( name );
-    m_displayMesh->loadGeometry( std::move( mesh ) );
     finalizeROFromGeometry( mat, Core::Transform::Identity() );
 }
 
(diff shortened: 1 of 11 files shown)


nmellado:
The code of this PR looks fine and brings several improvements, however it does not fix the problem.

I think I have found why: when using `loadGeometry`, the attributes are already created in the input `CoreGeometry`. In contrast, using the default `CoreGeometry`, the attributes are created after the call to `setupCoreMeshObservers`, thus the observers are not set. 
I see two options: either the `addAttrib` method sets the observers, or the user needs to call `addAttribObserver`. In both cases we need to update the documentation.

_Side note_: There is duplicated code between `addAttribObserver` and `setupCoreMeshObservers`. You might consider calling `addAttribObserver` (another version without name resolution) in  `setupCoreMeshObservers`.

MathiasPaulin:
this migh also fix #570 as shared_pointer ownership is the base problem in this issue.

PR #615, opened 2020-11-17
fix issue #584

### LINKED ISSUE #584: system wide lib interfere with radium external
On debian bullseye, when `libglm-dev` is installed, then link fail for globjects.
```
[ 84%] Performing build step for 'globjects'
[  1%] Linking CXX shared library ../../libglobjects.so
/usr/bin/ld: cannot find -lglm
```

to reproduce:
```
sudo apt install libglm-dev
```
cmake configuration fail
```
sudo apt remove --purge libglm-dev
```
cmake configuration pass

So it seems that the library search pass is not set correctly during configuration of globjects regarding glm

MathiasPaulin:
Closes #584 

The issue was due to wrong `external_DIR` variable values for glm, glbinding and globjects.
as the variables are set in two different files, there was some errors not detected by the CI nor our tests : 
 - `external/Engine/CMakeLists.txt` has errors in its paths but is used only to configure/build/install dependencies. These errors were at the origin of the issue.
 - `external/Engine/package.cmake` was correct and used by other Radium component so that the issue was hard to identify.

This PR set the right paths and try to set the paths as simple as possible and add a WARNING comment at the beginning of the `external/Engine/package.cmake` to prevent future errors.

### DIFF
--- a/external/Engine/CMakeLists.txt
+++ b/external/Engine/CMakeLists.txt
@@ -26,6 +26,7 @@ if (NOT DEFINED glm_DIR)
         ${RADIUM_EXTERNAL_CMAKE_OPTIONS}
         -DGLM_TEST_ENABLE=OFF
         -DBUILD_STATIC_LIBS=OFF
+        -DCMAKE_INSTALL_LIBDIR=lib
         )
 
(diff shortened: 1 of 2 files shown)


dlyr (approved):

nmellado (approved):
Well spotted...

PR #626, opened 2020-12-10
Expose manipulator keys (protected)

nmellado:

Small change in camera manipulators API

Right now the keys used in Radium manipulators are defined as `private` fields, which prevent there use in child classes.
This limits the creation of new manipulators extending existing ones.

Define keys (`KeyMappingManager::KeyMappingAction`) as protected fields instead of private.

No change required.

* **Other information**:


nmellado:
Example of use case: https://github.com/STORM-IRIT/Radium-Apps/blob/add_painter_app/Painter/CameraManipulator.hpp
I know that is example only disable rotation in the trackball manipulator, which could be done in a better way by updating the trackball API.
However, my feeling is that the proposed design allows more complex extensions of the manipulators, which is beneficial. In the other end, I don't think that the exposed fields break encapsulation.

MathiasPaulin (approved):

PR #627, opened 2020-12-11
Add self debugging function StackTrace and better Debug messages

MathiasPaulin:
This PR add only a `Ra::Core::Utils` fonction to display the execution stack and to ease self debugging. This function is called, e.g. by the macro `GL_CHECK_ERROR` in debug mode to ease locating in which call sequence an OpenGL error is thrown.


nmellado (approved):

nmellado:
Nice tool, thanks !

PR #629, opened 2020-12-12
fix issue #628

### LINKED ISSUE #628: Compilation error on master using MacOS
The last PR introduce a compilation error due to namespaces ambiguities:
https://github.com/STORM-IRIT/Radium-Releases/runs/1539840285?check_suite_focus=true#step:7:106

MathiasPaulin:
Closes #628

### DIFF
--- a/src/Engine/Component/GeometryComponent.hpp
+++ b/src/Engine/Component/GeometryComponent.hpp
@@ -175,7 +175,7 @@ class RA_ENGINE_API PointCloudComponent : public GeometryComponent
     // directly hold a reference to the displayMesh to simplify accesses in handlers
     std::shared_ptr<PointCloud> m_displayMesh {nullptr};
     // The diameter of the splat when rendered
-    float m_splatSize {0.0025};
+    float m_splatSize {0.0025f};
 };
 
 /*-----------------------------------------------------------------------------------------------*/
--- a/src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
+++ b/src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
@@ -144,6 +144,7 @@ ComponentMessenger::registerInput( const Entity* entity,
                                    const std::string& id,
                                    const typename CallbackTypes<ReturnType>::Setter& cb ) {
     CORE_ASSERT( entity && comp->getEntity() == entity, "Component not added to entity" );
+    CORE_UNUSED( comp );
     // Will insert a new entity entry if it doesn't exist.
     CallbackMap& entityList = m_entitySetLists[entity];
 
--- a/src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
+++ b/src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
@@ -100,11 +100,10 @@ inline bool ComponentMessenger::canRw( const Entity* entity, const std::string& 
 template <typename ReturnType>
 inline void
 ComponentMessenger::registerOutput( const Entity* entity,
-                                    Component* comp,
+                                    Component*,
                                     const std::string& id,
                                     const typename CallbackTypes<ReturnType>::Getter& cb ) {
     CORE_ASSERT( entity && comp->getEntity() == entity, "Component not added to entity" );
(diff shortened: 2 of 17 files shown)


nmellado:
I suspect an issue with CI, which does not get the source code from the branch but rather from the `master`...

nmellado:
Ok this is confirmed, i’ve found the bug with the GitHub actions.
I am on my phone now, will fix it tomorrow.

nmellado (approved):

nmellado:
CI fixed. This PR can be merged when CI is completed.

src/Engine/Component/GeometryComponent.hpp
```diff
@@ -175,7 +175,7 @@ class RA_ENGINE_API PointCloudComponent : public GeometryComponent
     // directly hold a reference to the displayMesh to simplify accesses in handlers
     std::shared_ptr<PointCloud> m_displayMesh {nullptr};
     // The diameter of the splat when rendered
-    float m_splatSize {0.0025};
+    float m_splatSize {0.0025f};
```
nmellado, line 178: Why not using Scalar ?

nmellado (dismissed):

src/Engine/Component/GeometryComponent.hpp
dlyr: Well, I think most of the OpenGL related stuff is float ... but I just change the suffix to remove warning (error ?) from MSVC ...

dlyr: And since I do not know more about this class, I prefer to modify as little as possible.

src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
```diff
@@ -144,6 +144,7 @@ ComponentMessenger::registerInput( const Entity* entity,
                                    const std::string& id,
                                    const typename CallbackTypes<ReturnType>::Setter& cb ) {
     CORE_ASSERT( entity && comp->getEntity() == entity, "Component not added to entity" );
+    CORE_UNUSED( comp );
```
MathiasPaulin, line 147: I think this kind of macro is error prone. Instead of a macro, that expand to dead code while the name still remains usable after that, il might be safer to just omit the name of the parameter in the method/function definition.

src/Engine/Renderer/Renderer.hpp
```diff
@@ -428,8 +428,8 @@ class RA_ENGINE_API Renderer
   private:
     // Qt has the nice idea to bind an fbo before giving you the opengl context,
     // this flag is used to save it (and render the final screen on it)
-    int m_qtPlz;
-    int m_qtViewport[4];
+    //  int m_qtPlz;
+    //  int m_qtViewport[4];
```
MathiasPaulin, line 432: Why not removing these lines and the comment above if these are not needed anymore?

src/Engine/Component/GeometryComponent.hpp
nmellado: Fine !

src/Engine/Managers/ComponentMessenger/ComponentMessenger.inl
nmellado: Hum, true.

dlyr: Well, not sure how it works, but do this change a lot of user code ?
If not, well I'm fine changing the interface right now.

MathiasPaulin: Nothing to change in the user code. Just give only type of parameter that is not used (without name). eg, for here (can't make a suggestion ...)
```cpp
ComponentMessenger::registerInput( const Entity* entity,
                                   Component* , // This un-named parameter can't be used in the method and will not generate "unused" warning
                                   const std::string& id,
                                   ...
```

dlyr: ah, I though of removing the comp parameter ;)

nmellado (approved):

Issue #611, opened 2020-11-08
Malformed custom targets Install_... Build run out of memory in docker-based CI


MathiasPaulin:
When building Radium in memory-bounded environment (e.g. free Gitlab docker runners, limited to 2gb of memory)
the command `make Install_` used to compile configured components and install them in one command fail.

This command, as defined in the main `CMakeLists.txt`, line 124, launches as many compilation process as there are files to compile. 
This makes `gcc` to run out of memory as compiling Radium this way requires 16GB of memory (on Linux) due to template instantiation.

The way the target `Install_` is defined prevent the caller to set the max number of jobs for its compilation environment  (e.g. by using `make -j 4 Install_`) as the "infinite" number of job to use is set directly in the target definition.

This malformed target definition is also present in the file `cmake/Config.cmake.in`, line 6 and generates the same effect (run out of memory) when compiling some Radium client libraries or apps that rely on Radium's templated classes.

The solution to this is to remove the `-j` option in the definition of the custom target and to let the user, either to 
  - explicitly set the option when calling make (e.g. `make -j <jobs> Install_`) or by using `cmake --build <dir> -j <jobs> --target Install_` 
  - or define the cmake environment variable `CMAKE_BUILD_PARALLEL_LEVEL` (see https://cmake.org/cmake/help/v3.16/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html) before compiling using `cmake --build ...`

Note that if Radium is compiled and installed using 2 commands (i.e. `make && make Install_`) the problem can be avoided (but is still there) but this is not documented. This allow to compile in low memory environment but spawning an infinite number of threads for only installing everything is then useless.

PR #636, opened 2020-12-17
update doc w.r.t. tools version and CMAKE_BUILD_PARALLEL_LEVEL

hoshiryu:

Update the documentation about:
 - the minimum version of CMake (from 3.0.+ to 3.15.7+ for the need for at least one sub-command)
 - the minimum version of Qt (from 5.10 to 5.14, though 5.11.3 works with a slight code modification)
 - a warning about the parallel configuration of Radium on computers with low RAM capacities, i.e. less than 12G.
   (Though I only tested on my Debian buster 8G RAM laptop, more RAM might be needed)

* **Other information**:

### DIFF
--- a/doc/basics/compilation.md
+++ b/doc/basics/compilation.md
@@ -6,14 +6,14 @@ All the dependencies are automatically fetched and compiled at build time.
  * [Core] Eigen, OpenMesh
  * [Engine] glm, globjects, glbindings
  * [IO] Assimp
- * [GuiBase] Qt Core, Qt Widgets and Qt OpenGL v5.5+ (5.10 recommended)
+ * [GuiBase] Qt Core, Qt Widgets and Qt OpenGL v5.5+ (5.14 recommended)
  * stb_image
 
 See @ref develbuildchain for technical details on dependencies management and how to provide locally installed dependencies.
 
 Minimal requirements
  * OpenGL 3+ / GLSL 330+
- * CMake 3.0.+
+ * CMake 3.15.7+
 
 # Supported compiler and platforms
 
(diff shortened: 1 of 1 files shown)


doc/basics/compilation.md
```diff
@@ -116,6 +116,8 @@ RADIUM_UPDATE_VERSION:BOOL=ON
 
 All radium related cmake options (with their current values) can be printed with `cmake -LAH | grep -B1 RADIUM` (on linux like system)
 
+\warning For computers with low RAM capacities (under 12G) we recommend to set the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable to a reasonable value (i.e. 2) to prevent the computer from swapping.
```
nmellado, line 119:
```suggestion
\warning For computers with low RAM capacities (under 12G) we recommend to set the `CMAKE_BUILD_PARALLEL_LEVEL` variable to a reasonable value (i.e. 2) to prevent the computer from swapping.
```

nmellado:
Please update wrt to `master` (I have merged the other PR)

doc/basics/compilation.md
hoshiryu: The variable has to be set in the environment, setting it with `-DCMAKE_BUILD_PARALLEL_LEVEL` does not work since it is fetched with `ENV{CMAKE_BUILD_PARALLEL_LEVEL}`. I would rather not remove that precision.

MathiasPaulin: **TL;DR** : The choice of using ENV variable is explained in the issue #611 fixed by the PR #612.

As said in the doc, the `CMAKE_BUILD_PARALLEL_LEVEL` initial value is taken from the calling process environment
(https://cmake.org/cmake/help/v3.18/envvar/CMAKE_BUILD_PARALLEL_LEVEL.html) and, if this variable is not defined in the environment, the parallel build will use the default value of the native build tool.

These lines allows to change the default build tool parallel build level while still allowing the user to control this value.

Most of the native build tool have a default value to 1, that makes the compilation very long. But setting the value to the number of core whithout testing the ENV variable prevent to build and install the libs on low memory computer (e.g. some CI docker runners only offers 4 Gb) or prevent to do a `build -j xxx`, as the build tools does not accept that one overrides the jobs number.

MathiasPaulin (approved):

PR #641, opened 2020-12-19
[cmake] make doc CMakeLists.txt working standalone.

dlyr:

* **What kind of change does this PR introduce?** 
bug fix

* **What is the current behavior?**
github workflow use doc/CMakeLists.txt as a standalone project, but some cmake variables are then not set.
For instance there is no sources files listed when generating the doc, so no documentation for the classes ;)

(re-)define virables in the doc/CMakeLists.txt, so the documentation can be generated as a standalone cmake project.

no

* **Other information**:
I focus on documentation pages in the previous PRs, and just realize "something" is missing.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -44,6 +44,20 @@ if(DOXYGEN_FOUND)
     # configure doxygen target 
     set(RADIUM_DOC_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/)
     set(RADIUM_MAIN_DOC_FILE ${CMAKE_CURRENT_SOURCE_DIR}/main.md)
+
+    find_package(Git QUIET)
+    if(GIT_FOUND AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../.git")
+        execute_process(COMMAND ${GIT_EXECUTABLE} rev-parse HEAD
+            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+            OUTPUT_VARIABLE GIT_CHANGESET)
+        if ( GIT_CHANGESET )
+            # remove new line sometime appearing in git changeset
+            string(REGEX REPLACE "\n$" "" GIT_CHANGESET "${GIT_CHANGESET}")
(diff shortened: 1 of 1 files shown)


doc/CMakeLists.txt
```diff
@@ -44,6 +44,20 @@ if(DOXYGEN_FOUND)
     # configure doxygen target 
     set(RADIUM_DOC_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/)
     set(RADIUM_MAIN_DOC_FILE ${CMAKE_CURRENT_SOURCE_DIR}/main.md)
+
```
nmellado, line 47: Could you do that only if `GIT_CHANGESET` does not exists ?

nmellado (commented):
Did you check that all the @fields@ in doxyfile.in are all defined this way ?

doc/CMakeLists.txt
dlyr: Yes, but it works in both cases (and I found adding a IF here could be misleading ...)
So if you insists I'll do it ;)

dlyr:
I think you refer to an old way of generating the Doxyfile.
All fields are now given within cmake with set `DOXYGEN_`variables, and as far as I see there is no misising information, but a double check is welcome.

doc/CMakeLists.txt
nmellado: Thinking about it, I think it is better to keep the computation of this value in this file, so I am fine with the current version.

nmellado (approved):

dlyr:
To test on your side : 
```bash
mkdir build-main-doc
cmake -S . -B build-main-doc/ #this one installs external dependencies
cmake --build --target RadiumDoc -B build-main-doc/
# doc is in build-main-doc/doc/html


mkdir build-doc-only
cmake -S doc/ -B build-doc-only/
# there are Dev warning : no project, no min cmake, but fine I think
cmake --build --target RadiumDoc -B build-doc-only/
# doc is in build-doc-only/html/
```

dlyr:
(arg you're too fast ;))

PR #643, opened 2021-01-02
[io] TinyPlyLoader refactor and cppcheck fix.

dlyr:

Cppcheck reports a possible dereference of nullptr, which is right.
After trying to understand the code flow attended, I change the tests so that It works as I think it should.
Also this PR introduce code refactor to have less redundancy, and should be more efficient, but firsts tests do not show loading time diminution.

* **Other information**:
At least one real test before accepting the PR.
-> done

### DIFF
--- a/src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
+++ b/src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
@@ -116,47 +116,50 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
     // read buffer data from file content
     file.read( ss );
 
-    std::vector<Eigen::Matrix<float, 3, 1, Eigen::DontAlign>> verts( vertBuffer->count );
-    std::memcpy( verts.data(), vertBuffer->buffer.get(), vertBuffer->buffer.size_bytes() );
-    geometry->setVertices( verts );
+    auto copyBufferToGeometry = []( const std::shared_ptr<tinyply::PlyData>& buffer,
+                                    Ra::Core::Vector3Array& container ) {
+        if ( buffer && buffer->count != 0 )
+        {
+            auto floatBuffer = reinterpret_cast<float*>( buffer->buffer.get() );
+            container.reserve( buffer->count );
+            for ( size_t i = 0; i < buffer->count; ++i )
+            {
+                container.emplace_back(
+                    floatBuffer[i * 3 + 0], floatBuffer[i * 3 + 1], floatBuffer[i * 3 + 2] );
+            }
+        }
+    };
 
-    if ( normalBuffer && normalBuffer->count != 0 )
-    {
-        std::vector<Eigen::Matrix<float, 3, 1, Eigen::DontAlign>> normals( normalBuffer->count );
-        std::memcpy(
-            normals.data(), normalBuffer->buffer.get(), normalBuffer->buffer.size_bytes() );
-        geometry->setNormals( normals );
-    }
+    copyBufferToGeometry( vertBuffer, geometry->getVertices() );
+    copyBufferToGeometry( normalBuffer, geometry->getNormals() );
 
     size_t colorCount = colorBuffer ? colorBuffer->count : 0;
     if ( colorCount != 0 )
     {
-
-        std::vector<Eigen::Matrix<uint8_t, 3, 1, Eigen::DontAlign>> colors( colorBuffer->count );
-
-        std::memcpy( colors.data(), colorBuffer->buffer.get(), colorBuffer->buffer.size_bytes() );
-        auto* cols = colors.data();
-
         auto& container = geometry->getColors();
-        container.resize( colorCount );
+        container.reserve( colorCount );
 
         if ( alphaBuffer && alphaBuffer->count == colorCount )
         {
-            uint8_t* al = alphaBuffer->buffer.get();
-            for ( auto& c : container )
+            uint8_t* al     = alphaBuffer->buffer.get();
+            uint8_t* colors = colorBuffer->buffer.get();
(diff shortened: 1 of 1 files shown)


src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -134,6 +135,7 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
 
     file.read( ss );
 
+    if ( vertBuffer && vertBuffer->count != 0 )
```
MathiasPaulin, line 138: This test is always true if this line is reached as it is the negation of the test at line 93 that ends the function if true.
To remove the cppcheck warning, only the modification proposed at line 93 is needed.

dlyr: well right, I fist add this line to remove the warning, then modifiy the bogus test, we may add the following comment
```suggestion
    // here we know that  !( !vertBuffer || vertBuffer->count == 0 ), i.e. ( vertBuffer && vertBuffer->count != 0 )
```

MathiasPaulin (approved):
The initial bug is fixed. There is still a lot of code redundancy but this might evolve in the future. The PR could be merged.

dlyr:
okey here is a bit of "compactness". I do not see other things easy to factorize in this code. If you have any advice, comments welcome.

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -73,123 +71,106 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
-    { colorBuffer = file.request_properties_from_element( "vertex", {"red", "green", "blue"} ); }
-    catch ( const std::exception& e )
-    {
-        colorBuffer = nullptr;
-        LOG( logERROR ) << "[TinyPLY] " << e.what();
-    }
-
-    std::clock_t startTime;
-    startTime = std::clock();
+    std::shared_ptr<tinyply::PlyData> normalBuffer {initBuffer( "vertex", {"nx", "ny", "nz"} )};
```
MathiasPaulin, line 112:
```suggestion
    auto normalBuffer {initBuffer( "vertex", {"nx", "ny", "nz"} )};
```

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -73,123 +71,106 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
+    std::shared_ptr<tinyply::PlyData> alphaBuffer {initBuffer( "vertex", {"alpha"} )};
```
MathiasPaulin, line 113:
```suggestion
    auto alphaBuffer {initBuffer( "vertex", {"alpha"} )};
```

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -73,123 +71,106 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
+    std::shared_ptr<tinyply::PlyData> colorBuffer {
```
MathiasPaulin, line 114:
```suggestion
    auto colorBuffer {initBuffer( "vertex", {"red", "green", "blue"} )};
```

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -73,123 +71,106 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
+        initBuffer( "vertex", {"red", "green", "blue"} )};
```
MathiasPaulin, line 115:
```suggestion
```

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -73,123 +71,106 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
 
-    {
-        const size_t numVerticesBytes = vertBuffer->buffer.size_bytes();
-        std::vector<Eigen::Matrix<float, 3, 1, Eigen::DontAlign>> verts( vertBuffer->count );
-        std::memcpy( verts.data(), vertBuffer->buffer.get(), numVerticesBytes );
-        geometry->setVertices( verts );
-    }
+    std::vector<Eigen::Matrix<float, 3, 1, Eigen::DontAlign>> verts( vertBuffer->count );
+    std::memcpy( verts.data(), vertBuffer->buffer.get(), vertBuffer->buffer.size_bytes() );
+    geometry->setVertices( verts );
```
MathiasPaulin, line 122: Out of scope for this PR : data are copied twice here : 
- First by the `std::memcpy` from `vertBuffer` to `verts`
- Second by `setVertices` from`verts` to `geometry->m_vertex`

This can result in performance issues when loading huge dataset. I understand that the second copy takes care of alignment and type promotion (i.e. from float to Scalar) but this second copy is expensive in term of memory allocation/reallocation (it uses a `back_inserter` into a vector that is not correctly sized most of the time.

dlyr:
@nmellado I think at least one test by a "point cloud" guy to confirm nothing is broken is a plus.

nmellado:
ok, I was not sure that the PR was ready to review.
Good now ?

dlyr:
sure

dlyr:
After some "simple" tests:
 - loading time is the same
 - there is a bug in normal display (but not from this PR)

Upcoming PR with integration tests

nmellado (approved):

PR #644, opened 2021-01-06
Intergration/Cache/Coverage

### LINKED ISSUE #645: [CI] Cache dependencies source and build folders
CI fetch and recompile Radium dependencies at each runs.
As we now have a way to manage dependencies out of Radium, it could save a lot of time to cache the source and build directories of out-of-build dependencies. This would allow to:
 - save time by skipping compilation at each run
 - still allow dependencies updates by always calling the dependencies cmake project on the cached directories (should compile only in case of change).

dlyr:

Feature : add integration tests (that could not be done with unit tests only).
Example is given for tinyplyloader
Add also coverage computation (RADIUM_ENABLE_COVERAGE), and trigger codecov.io update action
Add external caches for github action.

no integration tests

add cmake and dir structures to add integration tests with one example
add lcov-* targets

nope
Fixes #645 

* **Other information**:
no doc for the moment, and work on linux only I think, due to some scripts.


nmellado:
Could you update GitHub Actions so they actually run the integration tests ?

dlyr:
Well, they do already (since the check target trigger all ctest tests) see https://github.com/STORM-IRIT/Radium-Engine/pull/644/checks?check_run_id=1657452456
```
1/3 Test #1: catch_unit_tests .................   Passed    0.01 sec
    Start 2: tinyply-info_create
2/3 Test #2: tinyply-info_create ..............   Passed    0.07 sec
    Start 3: tinyply-info_compare
3/3 Test #3: tinyply-info_compare .............   Passed    0.00 sec
```

nmellado:
I think that you need to reduce the number of compilation thread for the coverage build, the compiler crashed, probably running out of memory.

tests/CMakeLists.txt
```diff
@@ -17,41 +16,54 @@ else()
...
+    # TODO : find a way to that more efficiently
+    find_dependency(cpplocate REQUIRED NO_DEFAULT_PATH)
+    find_dependency(glbinding REQUIRED NO_DEFAULT_PATH)
+    find_dependency(globjects REQUIRED NO_DEFAULT_PATH)
+    find_dependency(tinyply REQUIRED NO_DEFAULT_PATH)
+    set(RadiumExternalDlls_location "")
+    addImportedDir( FROM cpplocate::cpplocate TO RadiumExternalDlls_location) 
+    addImportedDir( FROM glbinding::glbinding TO RadiumExternalDlls_location) 
+    addImportedDir( FROM globjects::globjects TO RadiumExternalDlls_location)
+    addImportedDir( FROM tinyply TO RadiumExternalDlls_location)
```
nmellado, line 31: The use of tinyply is garded by the option `RADIUM_IO_TINYPLY`, so I suggest:
```suggestion
    if (RADIUM_IO_TINYPLY)
        addImportedDir( FROM tinyply TO RadiumExternalDlls_location) 
    endif ()
```

dlyr: Should be done. If I didn't mixup with rebases.

nmellado:
Could we mention in the documentation that submodules are required (changes the compilation instructions)

dlyr:
Well in fact I have second thought, I will remove the submodule

dlyr:
There is an inconsistency when I run from my own install and when it runs from CI, the report is not the same.
Let's do some tests before accepting the PR.

nmellado (approved):

nmellado (approved):

PR #647, opened 2021-01-11
Update assimp

dlyr:

Update assimp external
TODO:
- [ ] tests if everything is loaded correctly
- [ ] specifically animation
- [ ] choose a commit as reference, right now it's master, but it's a bad idea since it can changes anytime, and introduce unexpected behavior

This update solves compilation problems under windows, and also when you have Release Application with Debug Lib (or the other way round).

Compilation fails on windows (with assimp, in CI), compilation fails if your application is compiled in Debug with Radium lib in Release.

Everything compiles fine.

No change is expected, but maybe some post file loading process will be needed.


nmellado:
Can you split this PR in two: update assimp on one side, and fix testing on windows on the other side ?

nmellado:
@MathiasPaulin did you have a specific commit in mind for this one ?
We have a specific tag for all our externals, I would prefer we give the sha of the current `master` instead of referring to the branch directly.

MathiasPaulin:
> @MathiasPaulin did you have a specific commit in mind for this one ?
Not really. Any one that allows to build and use assimp reliably is welcome. I need to test the current proposal.

external/IO/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if (RADIUM_IO_ASSIMP)
         # For now, only configure in Debug mode to work with xcode
         ExternalProject_Add(assimp
             GIT_REPOSITORY https://github.com/assimp/assimp.git
-            GIT_TAG tags/v5.0.1
+            GIT_TAG master
```
nmellado, line 28:
```suggestion
            GIT_TAG ec08092dbf559a9da352889317f07afa56c97bdb
```

nmellado: In case we want to stick to current `master`

external/IO/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if (RADIUM_IO_ASSIMP)
...
             GIT_SHALLOW TRUE
```
dlyr, line 29:
```suggestion
            GIT_SHALLOW FALSE
```
in case we set a commit instead of a tag or branch name

dlyr:
I think we could squash merge at the end, so I don't need to rewrite history.

MathiasPaulin (changes requested):
This assimp update breaks animation loading in Radium. Some scene does not load due to incomplete scene detection (all the scene is below an animated node), others loads but with wrong animation time, speed and matrices ... 
Those issues must be investigated and solved before upgrading assimp.

MathiasPaulin:
> This assimp update breaks animation loading in Radium. Some scene does not load due to incomplete scene detection (all the scene is below an animated node), others loads but with wrong animation time, speed and matrices ...
> Those issues must be investigated and solved before upgrading assimp.

After investigations, there are problems with this assimp update : 
 - from the assimp side. Issue https://github.com/assimp/assimp/issues/3584 refer to collada import problem on animation but even fixing this as indicated in the issue does not solve loading problems. There are issues about loading meshes attached to bones in the assimp collada loader (meshes not loaded, wrong face number, ...).
 - from the Radium side. Some problems arise that concerns the animation time and duration.

The first thing to do is to fix the assimp side problems (hard to understand where they come from), then the Radium-side one before accepting this PR.

MathiasPaulin:
> The first thing to do is to fix the assimp side problems (hard to understand where they come from), then the Radium-side one before accepting this PR.

Assimp issues related to this PR : 
https://github.com/assimp/assimp/issues/3584
https://github.com/assimp/assimp/issues/3576
https://github.com/assimp/assimp/issues/3477
https://github.com/assimp/assimp/issues/3468

kimkulling:
See https://github.com/assimp/assimp/pull/3619, is fixed on the assimp-master,

nmellado:
Thanks @kimkulling for updating us on this.
We will come back to you if we find other problems.

MathiasPaulin:
With assimp master from 05-25-2021 : 
Radium compiles fine but there is huge problems when loading animations : 
  - from dae files : no binding from skeleton to meshes,
  - from fbx files : wrong matrices and time settings.
 
The first thing to do is to verify our transformations from assimp data to Radium data is correct, if so, understand what is incorrect on the assimp side and open a issue on assimp about that. But note that as the issues 
assimp/assimp#3576
assimp/assimp#3477
assimp/assimp#3468
that might solve part of the problem are still opened, perhaps our problem will be fixed soon.

PR #655, opened 2021-01-18
Resources

dlyr:

Refactor resrouces using std filesystem
also cleanup and simplify

path are post processed to replace \ with / and remove '../' to make canonical path
replace the dummy class ResourcesLocator by one function

using std::filesystem path and canonical is robust to os and filesystem's

ResourcesLocator has to be replace by the direct call to the corresponding function

* **Other information**:


dlyr:
damn seems like ubuntu 18.04 do not come with std::filesystem :(

dlyr:
I have added the base file for tests, but not sure what to test
if the search subdir do not exists, then the path return, for Resources for instance, is "Resources/", instead of "/my/bin/dir/Resources/"

MathiasPaulin:
> damn seems like ubuntu 18.04 do not come with std::filesystem :(

well, not ubuntu, gcc.
according to this https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017, std::filesystem is supported only from gcc 8.1 and you you use gcc 7.5 in this CI ...

MathiasPaulin:
> I have added the base file for tests, but not sure what to test
> if the search subdir do not exists, then the path return, for Resources for instance, is "Resources/", instead of "/my/bin/dir/Resources/"

Well, if the search subdir does not exist, returning a valid string is not really a good idea I think.

src/Engine/RadiumEngine.cpp
```diff
@@ -43,7 +43,7 @@ RadiumEngine::~RadiumEngine() = default;
 
 void RadiumEngine::initialize() {
     LOG( logINFO ) << "*** Radium Engine ***";
-    m_resourcesRootDir    = {Core::Resources::getRadiumResourcesDir()};
+    m_resourcesRootDir    = {Core::Resources::getRadiumResourcesPath()};
```
MathiasPaulin, line 46: As the result of this is used quite several time (and not only at initialization), why not caching it instead of executing `search_path` each time ?

dlyr: Not the purpose of this PR.

src/Core/Resources/Resources.hpp
```diff
@@ -19,38 +19,32 @@ namespace Core {
...
+
+/// Get the path prefix to access the current executable.
+RA_CORE_API std::string getBasePath();
+
+/// Get the resource path of the current executable.
+RA_CORE_API std::string getBaseResourcesPath();
+
+/// Get a resource path that offsets the path to the dynamic library that contains
+/// the given symbol by the given offset.
+RA_CORE_API std::string getResourcesPath( void* symbol, const std::string& offset = "../" );
```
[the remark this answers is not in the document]
dlyr: Same as before, empty (i'll update the doc).

src/Core/Resources/Resources.cpp
```diff
@@ -3,59 +3,48 @@
 #include <Core/Utils/Log.hpp>
 #include <algorithm>
 #include <cpplocate/cpplocate.h>
+#include <filesystem>
```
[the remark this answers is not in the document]
dlyr: Well, I'm not sure it's what the doc says.
The variable is on the cmake side.

So we could either check that fs.Final is found, or do something like
add_compile_definition(FILESYSYTEM_EXPERIMENTAL_OR_FINAL  ...)

and 
```suggestion
#ifdef FILESYSTEM_FINAL
#include <filesystem>
#else
#include <experimental/filesystem>
#endif
```
But I'm not sure it's better.

By the way CI pass on the 3 arch with gcc-8 on linux.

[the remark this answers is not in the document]
MathiasPaulin: Your right, but then I prefer to add the following compile definition in the `CMakeFiles.txt`
`add_compile_definition(CXX_FILESYSTEM_HEADER=${CXX_FILESYSTEM_HEADER})` if the module is found and do not make the source heavier than needed by using `#include < CXX_FILESYSTEM_HEADER >` instead of `ifdef ... else ... endif`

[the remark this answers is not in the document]
dlyr: what about https://stackoverflow.com/questions/53365538/how-to-determine-whether-to-use-filesystem-or-experimental-filesystem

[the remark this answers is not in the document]
MathiasPaulin: Discussing in real ?

nmellado:
See detection in OpenGR
https://github.com/STORM-IRIT/OpenGR/blob/master/cmake/FindFilesystem.cmake

dlyr:
Concerning the return value in case of "not found", I do not change the current behavior.
But we might think of something more robust (for this PR or a next one ...)
  - std::optional to state if the path exists
  - a std::path so the caller can check easily if the path exists
  - a "default value" if the searched resource is not in the vicinity of the base path.

This behavior then will drive the tests.

nmellado:
> Concerning the return value in case of "not found", I do not change the current behavior.
> But we might think of something more robust (for this PR or a next one ...)
> 
> * std::optional to state if the path exists
> * a std::path so the caller can check easily if the path exists
> * a "default value" if the searched resource is not in the vicinity of the base path.
> 
> This behavior then will drive the tests.

I would vote for `optional`, for me it is exactly designed for this kind of cases.

src/Engine/Renderer/Renderer.cpp
```diff
@@ -54,7 +54,7 @@ Renderer::~Renderer() {
 
 void Renderer::initialize( uint width, uint height ) {
     /// For internal resources management in a filesystem
-    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesDir()};
+    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesPath()};
```
MathiasPaulin, line 57:
```suggestion
    auto resourcesRootDir {Core::Resources::getRadiumResourcesPath()};
```

src/Engine/RadiumEngine.cpp
MathiasPaulin, line 46: Not sure that `={}`is needed. Why not just an `=` or a `std::move()` ...
This is legal, and used several time in other files/functions, but I do not understand the need to cast explicitly and then force the assignment as the return value of `Core::Resources::getRadiumResourcesPath()` is of the same type as the attribute `m_resourcesRootDir`

src/Engine/Renderer/Renderers/ForwardRenderer.cpp
```diff
@@ -64,7 +64,7 @@ void ForwardRenderer::initializeInternal() {
 
 void ForwardRenderer::initShaders() {
     /// For internal resources management in a filesystem
-    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesDir()};
+    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesPath()};
```
MathiasPaulin, line 67:
```suggestion
    auto resourcesRootDir {Core::Resources::getRadiumResourcesPath()};
```

src/Engine/Renderer/Renderers/ForwardRenderer.cpp
```diff
@@ -610,7 +610,7 @@ bool ForwardRenderer::buildRenderTechnique( RenderObject* ro ) const {
             {
                 ShaderConfiguration config = rt->getConfiguration( pass );
                 config.addShader( ShaderType_GEOMETRY,
-                                  std::string( Core::Resources::getRadiumResourcesDir() ) +
+                                  std::string( Core::Resources::getRadiumResourcesPath() ) +
```
MathiasPaulin, line 613: Why casting a `std::string` to `std::string` ?

dlyr:
Fine with optional, but what should be the behavior, for instance in registerMaterial, if the resource dir is not found ? 
-  CORE_ASSERT (only in debug thow)
-  exit(0)
-  Error message then 
   - nothing ?
   - quit ?

src/GuiBase/BaseApplication.cpp
```diff
@@ -81,7 +81,7 @@ BaseApplication::BaseApplication( int& argc,
 
     m_targetFPS = 60; // Default
     // TODO at startup, only load "standard plugins". This must be extended.
-    std::string pluginsPath = std::string {Core::Resources::getRadiumPluginsDir()};
+    std::string pluginsPath = std::string {Core::Resources::getRadiumPluginsPath()};
```
MathiasPaulin, line 84:
```suggestion
    auto pluginsPath {Core::Resources::getRadiumPluginsPath()};
```

src/GuiBase/BaseApplication.cpp
```diff
@@ -374,7 +374,7 @@ void BaseApplication::framesCountForStatsChanged( uint count ) {
 void BaseApplication::addBasicShaders() {
     using namespace Ra::Engine;
     /// For internal resources management in a filesystem
-    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesDir()};
+    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesPath()};
```
MathiasPaulin, line 377:
```suggestion
    auto resourcesRootDir {Core::Resources::getRadiumResourcesPath()};
```

src/GuiBase/Utils/KeyMappingManager.cpp
```diff
@@ -10,7 +10,7 @@ namespace Ra::Gui {
 using namespace Core::Utils; // log
 
 KeyMappingManager::KeyMappingManager() :
-    m_defaultConfigFile( std::string( Core::Resources::getRadiumResourcesDir() ) +
+    m_defaultConfigFile( std::string( Core::Resources::getRadiumResourcesPath() ) +
```
MathiasPaulin, line 13:
```suggestion
    m_defaultConfigFile( Core::Resources::getRadiumResourcesPath()  +
```

tests/ExampleApps/CoreExampleApp/main.cpp
```diff
@@ -9,7 +9,7 @@ int main( int /*argc*/, char** /*argv*/ ) {
     auto box  = Geometry::makeSharpBox( {1., 1., 1.} );
     auto topo = Geometry::TopologicalMesh( box );
 
-    auto resourcesPaths = Resources::getBaseDir();
+    auto resourcesPaths = Resources::getBasePath();
```
MathiasPaulin, line 12:
```suggestion
    auto resourcesPaths {Resources::getBasePath()};
```

MathiasPaulin:
> Fine with optional, but what should be the behavior, for instance in registerMaterial, if the resource dir is not found ?
> 
> * CORE_ASSERT (only in debug thow)
> * exit(0)
> * Error message then
>   
>   * nothing ?
>   * quit ?

throw ?

Well, I'm going out :p

dlyr:
By doing so (using optional) I just notice that RadiumEngine singleton initialize RadiumResourcesRootDir, 
I think it's fine to changes all Core::Resources::getRadiumResourcesPath()  by a newly created
RadiumEngine::getRadiumResourcesDir which just return the string value, and existence is done only once at setup.

MathiasPaulin:
> By doing so (using optional) I just notice that RadiumEngine singleton initialize RadiumResourcesRootDir,
> I think it's fine to changes all Core::Resources::getRadiumResourcesPath() by a newly created
> RadiumEngine::getRadiumResourcesDir which just return the string value, and existence is done only once at setup.

👍

dlyr:
(but just in Engine, because it's not possible in GuiBase without adding dependencies and includes)

dlyr:
A proposal for "optional", I add some comments to continue discussion

src/GuiBase/BaseApplication.cpp
```diff
@@ -81,8 +81,8 @@ BaseApplication::BaseApplication( int& argc,
 
     m_targetFPS = 60; // Default
     // TODO at startup, only load "standard plugins". This must be extended.
-    std::string pluginsPath = std::string {Core::Resources::getRadiumPluginsDir()};
-
+    auto pluginsPath {*Core::Resources::getRadiumPluginsPath()};
+    ///\todo check optional
```
dlyr, line 85: how/what should we do here

src/GuiBase/Utils/KeyMappingManager.cpp
```diff
@@ -10,7 +10,8 @@ namespace Ra::Gui {
 using namespace Core::Utils; // log
 
 KeyMappingManager::KeyMappingManager() :
-    m_defaultConfigFile( std::string( Core::Resources::getRadiumResourcesDir() ) +
+    ///\todo how to check here ?
+    m_defaultConfigFile( *Core::Resources::getRadiumResourcesPath() +
```
dlyr, line 14: do we need to check here ? 
what behavior ? 
should add dependency to RadiumEngine (don't know if it's initialize before btw).

src/Core/Resources/Resources.cpp
```diff
@@ -4,58 +4,70 @@
...
 }
 
-ResourcesLocator::ResourcesLocator( void* symbol, const std::string& offset ) : m_basePath {""} {
-    m_basePath = searchPath( "/", "", symbol ) + offset;
+std::optional<std::string> getRadiumPluginsPath() {
+    auto basePath =
+        searchPath( "Plugins/lib", "", reinterpret_cast<void*>( getRadiumPluginsPath ) );
+    basePath = clean( basePath / "Plugins" / "lib" );
+    // second option
+    return basePath.empty() ? std::nullopt : std::optional<std::string>( basePath.string() );
```
dlyr, line 56: and the second... any pref ?

src/Engine/RadiumEngine.cpp
dlyr: Well, it seems it's not related to this PR, but yes, upcoming commit fix that
Feel free to fix all of them in the codebase :rofl:

src/Engine/RadiumEngine.hpp
```diff
@@ -289,6 +289,8 @@ class RA_ENGINE_API RadiumEngine
     uint getFrame() const;
     /// \}
 
+    std::string getRadiumResourcesDir() { return m_resourcesRootDir; }
```
MathiasPaulin, line 292: Two things : 
 - The Engine is a singleton and the Radium resources are not linked to the engine but to the whole Radium ecosystem. So, this method might be static, with m_resourcesRootDir being also static and initialized/checked when the singleton is created.
 - As this correspond to _internal_ radium resources, perhaps call the method `getRadiumInternalResourcesDir()` or something more clear about the _Radium Bundled resources_ semantic.

src/Engine/Renderer/Material/BlinnPhongMaterial.cpp
```diff
@@ -101,7 +99,7 @@ bool BlinnPhongMaterial::isTransparent() const {
 
 void BlinnPhongMaterial::registerMaterial() {
     // For resources access (glsl files) in a filesystem
-    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesDir()};
+    auto resourcesRootDir {RadiumEngine::getInstance()->getRadiumResourcesDir()};
```
MathiasPaulin, line 102: Simpler to write if the `getRadiumResourcesDir` is static : 
```suggestion
    auto resourcesRootDir {RadiumEngine::getRadiumResourcesDir()};
```

src/GuiBase/BaseApplication.cpp
MathiasPaulin:
```suggestion
    auto radiumPluginsPath {Core::Resources::getRadiumPluginsPath()};
    if (radiumPluginsPath) {
        // loads plugins from *radiumPluginsPath
        ...
    } else {
        // message telling that no Bundled plugins are available (or nothing) ....
    }
```

src/GuiBase/BaseApplication.cpp
```diff
@@ -374,7 +374,7 @@ void BaseApplication::framesCountForStatsChanged( uint count ) {
 void BaseApplication::addBasicShaders() {
     using namespace Ra::Engine;
     /// For internal resources management in a filesystem
-    std::string resourcesRootDir = {Core::Resources::getRadiumResourcesDir()};
+    auto resourcesRootDir {*Core::Resources::getRadiumResourcesPath()};
```
MathiasPaulin, line 377: from https://en.cppreference.com/w/cpp/utility/optional/operator*, `The behavior is undefined if *this does not contain a value.`.
So, the option must be checked first. As the requested shaders are mandatory for the app to function, exit if the path is not found.

src/GuiBase/Utils/KeyMappingManager.cpp
MathiasPaulin: from https://en.cppreference.com/w/cpp/utility/optional/operator*, `The behavior is undefined if *this does not contain a value.`.
Concl. Yes, we have to check. No keymapping if config file not found.
I'm quite against adding a dependency to the Engine.

src/Engine/RadiumEngine.cpp
MathiasPaulin: 🥵 Well, its better that everybody meeting one of these always fix it. Like this, the code will be asymptotically clear 🤣

src/Engine/RadiumEngine.hpp
dlyr: Static data member initialized by "dynamic" singletons ?
I do not get the point here.
And two notes:
I think of using `RadiumEngine->getInstance()->getResourcesDir()` is a simple and meaningful alternative (the resources of RadiumEngine)
I do not get the idea of "ecosystem" used several time in the code/doc.
This function is more like a cache on the Engine side.

src/Core/Resources/Resources.cpp
```diff
@@ -4,58 +4,70 @@
...
+    return clean( cpplocate::getModulePath() );
 }
 
-const std::string& ResourcesLocator::getBasePath() {
-    return m_basePath;
+std::optional<std::string> getResourcesPath( void* symbol, const std::string& pattern ) {
+    auto basePath = searchPath( pattern, "", symbol );
+    basePath      = clean( basePath / pattern );
+    if ( basePath.empty() ) return {};
+    return ( basePath ).string();
```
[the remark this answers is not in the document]
dlyr: are you sure implicit conversion from path to string is portable ?
the doc says "Paths are implicitly convertible to and from std::basic_strings, which makes it possible to use them with over files APIs, e.g. as an argument to std::ifstream::open"
and path::string()  [doc](https://en.cppreference.com/w/cpp/filesystem/path/string)

src/Core/Resources/Resources.cpp
dlyr: well, short answer: no
` error: operands to ‘?:’ have different types ‘const std::nullopt_t’ and ‘std::filesystem::__cxx11::path’
   56 |     return basePath.empty() ? std::nullopt :  basePath ;`

dlyr: (see https://en.cppreference.com/w/cpp/utility/optional examples)

dlyr:
Well, on second though (it's a global comment since it concerns several review comments) here are listed some observation:
 -  GuiBase already depends from Engine
 -  The resources seems to be something "Engine only", there is no need of Resources in Core.
 -  The getBasePath and getResroucesPath are straight forward calls to cpplocate::locatePath
 -  I really do not get the point with some static method/data in singleton. Our implementation of singleton use initialization, why introducing "double behavior" for some data ?

Maybe it can be cleaner to:
 -  move getRadiumResourcesPath and getRadiumPluginsPath to RadiumEngine's method (right now plugins are also dependent to Engine, but I can change my mind if there is any plugins without engine situation)
 - maybe give a wrapper to cpplocate, if we think client code prefer to use a radium specific wrapper rather than direct call to cpplocate, which may be more versatile. For instance what if I want to search for a file rather than a path ?

src/Core/Resources/Resources.cpp
MathiasPaulin: Well, sorry. so I propose 
```suggestion
    return basePath.empty() ? std::nullopt : std::optional<std::string>( basePath );
```

src/Engine/RadiumEngine.hpp
MathiasPaulin: OK, I'll replace all the "ecosystem" into the code/doc in a next PR.

MathiasPaulin:
I agree with most of the thoughs expressed here. But, to be clear : 

> * GuiBase already depends from Engine

Yes, but some resources are used only by GuiBase (e.g. Keymapping config). 

> * The resources seems to be something "Engine only", there is no need of Resources in Core.

No, Resources can be also GuiBase only (see above), and, for the core, I think that some functionalities might use resources (for, e.g. Geometry processing) but I have no example in mond for now. BTW, I'm on the way to integrate the SkeletonAnimationPlugin into the Core/Engine/GuiBase and it might need some resources for the core part of SkeletonAnimation. 

> * The getBasePath and getResroucesPath are straight forward calls to cpplocate::locatePath

Agree but I think abstracting this (wrapper to cpplocate) represents not really a big cost and allows further modifications and dependency removal over cpplocate.

> * I really do not get the point with some static method/data in singleton. Our implementation of singleton use initialization, why introducing "double behavior" for some data ?

Kind of different point of view over the Singleton pattern implementation/usage/need. But I agree, Radium relies on one way to implement this and there is no need to introduce "double behavior". Note that some "singleton" in the Engine have another implementation (e.g. MaterialConverters, TechniqueBuilders, ...) What to do with these ?

> 
> Maybe it can be cleaner to:
> 
> * move getRadiumResourcesPath and getRadiumPluginsPath to RadiumEngine's method (right now plugins are also dependent to Engine, but I can change my mind if there is any plugins without engine situation)

Must be discussed more. For now, we have an application (geometry and scene processing) that depends on Engine only because the Core does not propose sceneGraph (or assimilated) management.  It is planed to move some representation from Engine to Core, that will allow to manipulate scenes and scene data without initializing the Engine (as this was done for CoreMesh/EngineMesh). And, instead of relying only on the host filesystem to manage the resources files/hierarchy, it is planned to embed the resources into some binary internal representation. See, for this the feature request https://github.com/STORM-IRIT/Radium-Engine/issues/337

> * maybe give a wrapper to cpplocate, if we think client code prefer to use a radium specific wrapper rather than direct call to cpplocate, which may be more versatile. For instance what if I want to search for a file rather than a path ?

I prefer to have a wrapper over cpplocate, even if it does a direct call. For now, nothing prevent for searching for a file instead of a path as a `std::filesystem::path` could represent a regular file also.

dlyr:
Regarding @MathiasPaulin comments, I think the current state is complete.

In Engine, use RadiumEngine getResourcesDir cached at init
In other libs, calls (and cache if needed) Ra::Core::Resources::get's

MathiasPaulin:
> Regarding @MathiasPaulin comments, I think the current state is complete.
> 
> In Engine, use RadiumEngine getResourcesDir cached at init
> In other libs, calls (and cache if needed) Ra::Core::Resources::get's

OK, I'll test the PR with existing apps/lib/plugins and make my final review soon.

MathiasPaulin (approved):
Works fine on my tests (plugin, lib and app using both Engine Resources and own resources)

dlyr:
I have one commit to add, but could also be done in another PR since is not directly related with Resources (but resources management).

MathiasPaulin:
> I have one commit to add, but could also be done in another PR since is not directly related with Resources (but resources management).

Add it so that we can merge this PR. I found an issue related to resources sharing that I will describe now and fix as soon as this PR is merged.

nmellado (approved):

nmellado:
The CI can be fixed as described here: https://github.com/jurplel/install-qt-action/issues/78 (see example https://github.com/akiyosi/goneovim/commit/cb4b15d149375b5114d4c146f8c78bb053f8f8af)

MathiasPaulin (dismissed):

cmake/coverageFunc.cmake
```diff
@@ -0,0 +1,112 @@
...
+        set(RADIUM_ENABLE_TESTING "ON")
+        set(ENABLE_COVERAGE "ON")
+    else()
+        message( WARNING "[Tests] Coverage requested (RADIUM_ENABLE_COVERAGE), but not activated.")
+        message( WARNING "[Tests] Coverage only supported in DEBUG with GCC.")
+    endif()
+endmacro()
+
+macro(setup_coverage_targets)
+    add_custom_target(lcov-init
```
nmellado, line 92: Don't we need a guard here as well ?
```cmake
if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_BUILD_TYPE STREQUAL "Debug")
```

cmake/messageFunc.cmake
```diff
@@ -0,0 +1,78 @@
+
```
nmellado, line 1: Could you please add a one-line comment showing an example of call ?
Usage is not that obvious

src/Core/Resources/Resources.hpp
```diff
@@ -18,39 +19,27 @@ namespace Core {
  *   library containing an arbitrary symbol. (easy with cpplocate)
  */
 namespace Resources {
-/// Get the path prefix to access Radium resources in a filesystem.
-/// This base prefix is the parent directory of the directory containing libRadiumCore, appended
-/// with /Resources/ as all Radium resources will be installed, bundled or linked for the buildtree
-/// in such a directory
-RA_CORE_API std::string getRadiumResourcesDir();
+using namespace Ra::Core::Utils;
+/// Radium resources are locate in Resources directory, searched from Radium lib location.
```
nmellado, line 23:
```suggestion
/// Radium resources are located in the Resources directory, searched from Radium lib location.
```

src/Core/Resources/Resources.hpp
```diff
@@ -18,39 +19,27 @@ namespace Core {
...
+/// @return the path to access Radium Resources if found, otherwise !has_value
+/// @note the pattern searched is "Resrouces/Shaders" since it's the basic resources dir.
```
nmellado, line 25:
```suggestion
/// @note the pattern searched is "Resources/Shaders" since it's the basic resources dir.
```

src/Core/Resources/Resources.hpp
```diff
@@ -18,39 +19,27 @@ namespace Core {
...
+RA_CORE_API optional<std::string> getRadiumResourcesPath();
 
-/// Get the path prefix to access Radium plugins in a filesystem.
-/// A Radium plugin is a plugin installed in the Radium bundled and embedded in installed
-/// application bundle./// This base prefix is the parent directory of the directory containing
-/// libRadiumCore, appended with /Plugins/lib as all Radium plugins will be installed, bundled or
-/// linked for the buildtree in such a directory
-RA_CORE_API std::string getRadiumPluginsDir();
+/// Radium plugins are locate in Plugins directory, searched from Radium lib location.
```
nmellado, line 28:
```suggestion
/// Radium plugins are located in Plugins directory, searched from Radium lib location.
```

src/Core/Resources/Resources.hpp
```diff
@@ -18,39 +19,27 @@ namespace Core {
...
-  public:
-    /// Construct a resource locator that will offset the path to the dynamic library that contains
-    /// the given symbol by the given offset.
-    explicit ResourcesLocator( void* symbol, const std::string& offset = "../" );
-    /// Construct a resource locator that will search for the given pattern starting from the offset
-    /// the path to the dynamic library or executable that contains the given symbol.
-    explicit ResourcesLocator( void* symbol, std::string pattern, const std::string& offset );
-    /// Return the base path corresponding of the locator execution
-    const std::string& getBasePath();
+/// Search for an accesible Resources (or pattern if given) directory in the current executable (or
```
nmellado, line 36:
```suggestion
/// Search for an accessible Resources (or pattern if given) directory in the current executable (or
```

src/Engine/Renderer/Material/BlinnPhongMaterial.cpp
```diff
@@ -1,16 +1,13 @@
...
+#include <Engine/RadiumEngine.hpp>
 #include <Engine/Renderer/Material/BlinnPhongMaterial.hpp>
 #include <Engine/Renderer/Material/MaterialConverters.hpp>
 #include <Engine/Renderer/RenderTechnique/RenderTechnique.hpp>
 #include <Engine/Renderer/RenderTechnique/ShaderConfigFactory.hpp>
-#include <Engine/Renderer/RenderTechnique/ShaderProgram.hpp>
 #include <Engine/Renderer/RenderTechnique/ShaderProgramManager.hpp>
-
-#include <Engine/Renderer/Texture/Texture.hpp>
+//#include <Engine/Renderer/Texture/Texture.hpp>
```
nmellado, line 8:
```suggestion
```

src/Engine/Renderer/Renderer.cpp
```diff
@@ -54,15 +54,16 @@ Renderer::~Renderer() {
...
 
     m_width  = width;
     m_height = height;
 
     // Initialize managers
     m_shaderMgr = ShaderProgramManager::getInstance();
     m_roMgr     = RadiumEngine::getInstance()->getRenderObjectManager();
-    TextureManager::createInstance();
+
+    // TMR    TextureManager::createInstance();
```
nmellado, line 66:
```suggestion
```

nmellado (approved):

PR #589, opened 2020-07-09
 Add operation with wedge to ease the representation of the topology of core mesh, with wedges to handle attribs.

dlyr:
* **Please check if the PR fulfills these requirements**

First part of this PR has been moved to #595 (add quad meshes for `Core` and `Engine`).
Second part of this PR has been moved to #719  (add wedge collapse).

Next steps, sorted by decreasing priority:
- [ ] Topological test collapse on non triangular faces
- [ ] Topological mesh for polymesh: split
- [ ] Topological mesh add/del face/halfedge/edge/vertex which should update wedges ref count
- [ ] Add Catmull-Clark subdivision (from previous implementation, see in code todo's)
- [ ] Refactor UpdateGL to remove duplicated code
- [ ] Test on real data
- [ ] Docs have been added / updated (for bug fixes / features)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Check picking/wireframe to show only edges corresponding to the core geometry
- [ ] Cleanup history __NO FORCE PUSH IN THE MEANTIME__ (other than rebase)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Add operation with wedge to ease the representation of the topology of core mesh, with wedges to handle attribs.

PR #662, opened 2021-01-25
Fix topo init

dlyr:

Bug fix.

When a TopologicalMesh is create with the initWithWedge method, face indices are broken

* **Other information**:

This patch is a hot fix, adapted from some part of #589, but a bit differently. 
After merge, take care when rebasing #589 to have the new version (since the initWithWedge method move from cpp to inl, it will not be automatic).

### DIFF
--- a/src/Core/Geometry/TopologicalMesh.cpp
+++ b/src/Core/Geometry/TopologicalMesh.cpp
@@ -470,7 +470,7 @@ void TopologicalMesh::initWithWedge( const TriangleMesh& triMesh ) {
         w.setWedgeData( std::move( wd ) );
         // the newly added wedge is not referenced yet, will be done with `newReference` when
         // creating faces just below
-        m_wedges.m_data.emplace_back( w );
+        m_wedges.m_data.push_back( w );
     }
 
     LOG( logINFO ) << "TopologicalMesh: have  " << m_wedges.size() << " wedges ";
--- a/src/Core/Geometry/TopologicalMesh.cpp
+++ b/src/Core/Geometry/TopologicalMesh.cpp
@@ -454,6 +454,9 @@ void TopologicalMesh::initWithWedge( const TriangleMesh& triMesh ) {
 
     for ( size_t i = 0; i < triMesh.vertices().size(); ++i )
     {
+        // create an empty wedge, with 0 ref
+        Wedge w;
+
         WedgeData wd;
         wd.m_position = triMesh.vertices()[i];
         copyMeshToWedgeData( triMesh,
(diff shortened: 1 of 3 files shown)


src/Core/Geometry/TopologicalMesh.cpp
```diff
@@ -463,8 +466,11 @@ void TopologicalMesh::initWithWedge( const TriangleMesh& triMesh ) {
                              m_wedges.m_wedgeVector3AttribHandles,
                              m_wedges.m_wedgeVector4AttribHandles,
                              &wd );
-
-        m_wedges.m_data.emplace_back( wd );
+        // here ref is not incremented
+        w.setWedgeData( std::move( wd ) );
+        // the newly added wedge is not referenced yet, will be done with `newReference` when
+        // creating faces just below
+        m_wedges.m_data.emplace_back( w );
```
nmellado, line 473: Is it push or emplace here ?

src/Core/Geometry/TopologicalMesh.cpp
```diff
@@ -493,10 +498,10 @@ void TopologicalMesh::initWithWedge( const TriangleMesh& triMesh ) {
...
             { vh = vtr->second; }
 
-            face_vhandles[j]    = vh;
-            face_normals[j]     = n;
-            face_vertexIndex[j] = inMeshVertexIndex;
-            face_wedges[j]      = WedgeIndex {int( i )};
+            face_vhandles[j] = vh;
+            face_normals[j]  = n;
+            face_wedges[j] =
+                WedgeIndex {static_cast<WedgeIndex::Index::IntegerType>( inMeshVertexIndex )};
```
nmellado, line 504: Do you really need a static cast ? Can’t we call the constructor directly with the integer as input?

src/Core/Geometry/TopologicalMesh.cpp
```diff
@@ -463,8 +466,11 @@ void TopologicalMesh::initWithWedge( const TriangleMesh& triMesh ) {
                              m_wedges.m_wedgeVector3AttribHandles,
                              m_wedges.m_wedgeVector4AttribHandles,
                              &wd );
-
-        m_wedges.m_data.emplace_back( wd );
+        // here ref is not incremented
+        w.setWedgeData( std::move( wd ) );
```
nmellado, line 470: Move is already called in the method ˋsetWedgeData ˋ

src/Core/Geometry/TopologicalMesh.cpp
dlyr: it's an unsigned int, so
```
warning: narrowing conversion of ‘inMeshVertexIndex’ from ‘unsigned int’ to ‘Ra::Core::Utils::Index::IntegerType’ {aka ‘int’}
```

dlyr: We could also add an implicit cast from unsigned int to Index

nmellado: Yes, it would make sens to add this constructor to the Index class

src/Core/Geometry/TopologicalMesh.cpp
dlyr: right, it was emplace, but with the fix, it's push. Fixed.

src/Core/Geometry/TopologicalMesh.cpp
nmellado: Fine to keep it as it is.

src/Core/Geometry/TopologicalMesh.cpp
nmellado: Are you wiling to do it in this PR ?

dlyr: Nope ;) feel free to add it

nmellado (approved):

PR #665, opened 2021-01-26
Add compiler version verification at configure time.

### LINKED ISSUE #664: Use of filesystem::lexically_normal not available in experimental
Current version of `master` requires cxx_17 capabilities at this line: 
https://github.com/STORM-IRIT/Radium-Engine/blob/31523ca27aaa231751b9dff4e2c4343225ee2c00/src/Core/Resources/Resources.cpp#L31

Doc: https://en.cppreference.com/w/cpp/filesystem/path/lexically_normal

As such, some compilers (e.g., gcc-7) cannot compile radium.

MathiasPaulin:
closes #664
This PR add compiler version verification for Radium bundle at configure time. 
This verification is exported to any Radium based system and is automatically done when `find_package(Radium)` is executed.

Verify that the compiler used for compiling Radium or on of its client (plugin, lib, app) is conform to the requirements

No compiler version is verified

* 
Stop at configure time if the compiler version does not meet the requirements

* **Other information**:
Solution taken from there : https://stackoverflow.com/questions/14933172/how-can-i-add-a-minimum-compiler-version-requisite

do we have to implement more complex testing (i.e. feature-detection approach) ?


MathiasPaulin:
I found two concerns with the way I implement compiler version verification. 
 - It is done at two different places (general CMakeLists.txt and package configuration file)
 - The required version are encoded twice also.

nmellado:
In order to avoid duplication, could you move the verification code in a macro (accessible both at build time and in the package) ?
Also, the version number for VS is very very accurate, maybe we can skip the minor version check ?

nmellado:
> > In order to avoid duplication, could you move the verification code in a macro (accessible both at build time and in the package) ?
> 
> I prefer functions instead of macro but then, the function must be called explicitely and the included file installed with Radium as for `radium_setup_functions`
Don’t mind between macro and functions, but you get my point. And yes, this should be made available in the installed package.


> 
> > Also, the version number for VS is very very accurate, maybe we can skip the minor version check ?
> 
> need to test this. I took the version numbers from the doc.
Ok. Skipping minor is probably ok then, as supported features change with major compiler versions.

MathiasPaulin:
Supported compilers are defined only once without installing one more file. 
Using this solution, we can remove the need to install `radium_setup_functions.cmake` by embeding it into the `RadiumConfig.cmake`. This might prevent misusage of this include script.

nmellado (approved):

PR #674, opened 2021-01-31
Enable assimp by default.

nmellado:

Doc + cmake update

Assimp is disabled by default in cmake, and tested only on linux and macOS.

Assimp is enabled and tested on all platforms.

* **Other information**:

### DIFF
--- a/.github/workflows/pull-request-ci.yml
+++ b/.github/workflows/pull-request-ci.yml
@@ -21,9 +21,9 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          - { name: "Windows MSVC", os: windows-latest, cc: "cl.exe", cxx: "cl.exe", assimp: "OFF", icon: "Windows"}
-          - { name: "Ubuntu gcc", os: ubuntu-latest, cc: "gcc-8", cxx: "g++-8", assimp: "ON", icon: "Linux" }
-          - { name: "MacOS clang", os: macos-latest, cc: "clang", cxx: "clang++", assimp: "ON", icon: "Apple" }
+          - { name: "Windows MSVC", os: windows-latest, cc: "cl.exe", cxx: "cl.exe", icon: "Windows"}
+          - { name: "Ubuntu gcc", os: ubuntu-latest, cc: "gcc-8", cxx: "g++-8", icon: "Linux" }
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request-ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: external
-          key: ${{ runner.os }}-external-${{ hashFiles('src/Radium-Engine/external/**/CMakeLists.txt') }}
+          key: ${{ runner.os }}-external-v1-${{ hashFiles('src/Radium-Engine/external/**/CMakeLists.txt') }}
```
dlyr, line 56: Well, we don't need this, I think, but rather have something that depends on compiles options. maybe the whole config ?
For instance (not tested) 
```suggestion
          key: ${{ matrix.config }}-external-${{ hashFiles('src/Radium-Engine/external/**/CMakeLists.txt') }}
```
But I'm not sure the matrix.config will convert to a string suitable for key definition.
The hashFiles ensure that modification to the externals CMakeLists.txt create a new key,
we can also use all Radium's CMakeLists, but might be overkill. 

I'm also fine with the version, but it make a manual update needed here ...

.github/workflows/pull-request-ci.yml
```diff
@@ -53,19 +53,19 @@ jobs:
         uses: actions/cache@v2
         with:
           path: external
-          key: ${{ runner.os }}-external-${{ hashFiles('src/Radium-Engine/external/**/CMakeLists.txt') }}
+          key: ${{ matrix.config }}-external-${{ hashFiles('src/Radium-Engine/external/**/CMakeLists.txt') }}
```
dlyr, line 56: well this solution do not work, the key is 
`Object-external-c20c9af153db4d90af43dee80421f2b56ba8b04b59beb31bf701b2054d21790a`
so matrix.config becomes "Object" in the key.
Maybe it's simpler to use the "version hack" if I can name it like that.
By the way next cache retrieval with this key may mess up the different build due to key collision for all OS.

MathiasPaulin (approved):

MathiasPaulin (approved):

PR #681, opened 2021-02-05
Tag head commit when receiving dispatch event

nmellado:
see https://github.com/STORM-IRIT/Radium-Releases/pull/10

### DIFF
--- a/.github/workflows/create-tag-from-dispatch.yml
+++ b/.github/workflows/create-tag-from-dispatch.yml
@@ -1,0 +1,20 @@
+name: Create tag (notified from dispatch event)
+
+defaults:
+  run:
+    shell: bash
+
+
+on:
+  repository_dispatch:
+    types: [new-tag-event]
+
+jobs:
+  createTag:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Tag current head
+        uses: tvdias/github-tagger@v0.0.1
+        with:
+          repo-token: "${{ secrets.REPO_DISPATCH_TOKEN }}"
+          tag: "${{ github.event.client_payload.tagname }}"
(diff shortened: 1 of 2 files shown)


dlyr:
**CHECK LINEAR HISTORY FAIL**

Please rebase on current up to date master,
without merge commits (e.g. you can squash them).

.github/workflows/createTagFromDispatch.yml
```diff
@@ -0,0 +1,20 @@
+name: Create tag (notified from dispatch event)
```
dlyr, line 1: should the filename be 
create-tag-from-dispatch.yml
to be consistent with the other workflow file naming convention here ?
I see the convention is different on Radium-Release but the same on Radium-Apps ... it's not mandatory, but cleaner to have consistency.

dlyr, line 1:
```suggestion
name: Create tag (notified from Radium-Release dispatch event)
```

dlyr (changes requested):

.github/workflows/createTagFromDispatch.yml
nmellado: I didn't mention Radium-Release here, as we might have other trigger in the future (does not have to be from Radium-Release)

nmellado: Hum right, I've followed the naming convention of Radium-Releases, but true, it's better to stick to Radium-Engine convention.

nmellado: Done.

dlyr (dismissed):

nmellado:
Qt is down. Same on Radium-Release.

dlyr (approved):

PR #683, opened 2021-02-08
[cmake] Fix 682

### LINKED ISSUE #682: Missing cmake dependency between globject and Eigen
When compiling the externals, we do not explicitly create the dependency between globjects and Eigen. 
In some cases it might lead to compilation issues: https://github.com/STORM-IRIT/Radium-Apps/runs/1856309287?check_suite_focus=true#step:9:4606

nmellado:
Bug fix

Fixes #682.

Create target-level dependency between globject and eigen when compiling externals

No change on the client side

* **Other information**:


dlyr (commented):
We should have the same dep between globjects and glbinding right ?

dlyr:
Actually it's done, I propose [this](https://github.com/STORM-IRIT/Radium-Engine/pull/683/commits/a054399709bf372e6631f26d60e94711c9791279) alternative, but I'll not vote against yours ;)

nmellado:
I prefer yours, clearly.
NO need for gl bindings due to transitivity

dlyr (approved):

PR #686, opened 2021-02-10
Set CMAKE_OSX_DEPLOYMENT_TARGET to 10.15

nmellado:
Bugfix

Radium cannot be compiled on OSX<10.15, because `std::filesystem` is not yet available.
https://developer.apple.com/documentation/xcode-release-notes/xcode-11-release-notes

Require OSX 10.15 buildchain (also available on older versions of MACOSX).

No

* **Other information**:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,9 @@ set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
 # Enable helpfull warnings and C++17 for all files
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+# Require OSX 10.15 to support std::filesystem (https://developer.apple.com/documentation/xcode-release-notes/xcode-11-release-notes)
+set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X deployment version")
+
 
 # Project setup, versioning stuff here, change when changing the version
 # Note: keep the project name lower case only for easy linux packaging support


dlyr (approved):

nmellado:
(Waiting for user feedback on OSX 10.14)

MathiasPaulin (changes requested):
Not needed. The compiler version requested is 12.0 and this compiler support filesystem. No nee to add more constraints than needed.

nmellado:
Apparently these are two different things.
Adding this option allow to use appleclang11 on osx 10.14 (I’ll just change the minimum version for apple Lang)

so we don’t add constraints, but rather support more platforms.

MathiasPaulin:
> (Waiting for user feedback on OSX 10.14)

as said in my negative review, AppleClang 12.0 is requested to compile Radium (see `cmake/compilerVersion.cmake:4` And this version is only available since MacOs 10.15. 
It is not necessary to have both belt and suspenders, just one is enough. And it is easier to maintain one file that defines compiler/os constraints than several. 

And if you want absolutely to set this constraints, set it so that even client applications/plugins/lib have the same deployement constraints. Thats why the compiler version verification is check when configuring Radium but is also embed in the `RadiumConfig.cmake` installed configuration script.

And Radium can be deployed on MacOsX 10.14 (and even less) when compiled with GNU gcc (using the macport or Homebrew installed version of GNU gcc.

MathiasPaulin:
> Adding this option allow to use appleclang11 on osx 10.14(I’ll just change the minimum version for apple Lang)

Not sure to understand ... if you limit the deployment to 10.15, what is the advantage to compile on a system you will not be able to deploy ? I've not seen the compiler version changed ... Are you sure AppleClang11 support filesystem. In this case, update also the doc. I'm on 10.14 (with Clang 10) and can't compile Radium. I'll give a try.
But I've also to verify that RAdium will stil be compiled and deploy with 10.14 SDK if compiled with GNU g++

And what about the client applications ? This file is not part of the `RadiumConfig.cmake` used when a `FindPackage(Radium)` is made ... Client should have also this automatically checked at configure time.

PatatalouiS:
Hello, I am the student who sent the email to @nmellado  before this pull request. 

It turns out that after updating my old gcc v9.x to gcc v10.2, I can compile Radium without problems. 

But the latest version of CLang and stdlib available on OSX 10.4 (Mojave) does not offer it, the std :: filesystem module is written but unavailable and marked "experimental". I tried to compile a program containing calls to the namespace concerned to understand the why. 

Capture d’écran 2021-02-10 à 15 40 42

It appears that, (and I have no clear explanation), by passing the flag --mmacosx-version-min = 10.15 to CLang, the std :: filesystem module compiles perfectly and without warning, until the program is executed. 

Capture d’écran 2021-02-10 à 15 44 51

If building Radium goes perfectly well with gcc, the problem is less annoying. If CLang 11 support is too complicated, it might be a good idea to report it to user when OSX <10.15 (Catalina) is detected and prompt the user to install gcc. Many MAC users swear by CLang, as it comes directly with XCode, by default. In addition, Apple misleads many novice users, many people think they are using clang on their OSX machine while they are using CLang, the 'gcc' and 'g ++' commands are alias directly to the default CLang. .

Obviously, pleaser consider i’m speaking as a student, I guess someone more experienced than me would have spent a lot less time getting worked up at their machine to build the engine.
(Sorry for my bad English)

nmellado:
@MathiasPaulin from my understanding of @PatatalouiS feedback, this PR allows to compile even on macOS 10.14.
@PatatalouiS can you confirm you are then able to run the compiled application ?

MathiasPaulin:
I compile Radium and all its client apps/plugins/lib without problem using GNU g++ 9.3 on MacOs Mojave 10.14.6 with macOS sdk 10,14.
As I have only XCode 10.3 installed, I can't compile using AppleClang 10 which comes with xcode 10, does not support std::filesystem.
As soon as i got the required bandwidth to install XCode 11 (7.5 Gb to download, my network will take 12 hours :( ) and so AppleClang 11, I'll give a try.

BTW, beside waiting to test, this PR is incomplete. The doc must be updated if we go down to AppleClang 11, the compiler version must be updated in the cmake/compilerVersion.cmake and if this SDK requirement is valid, it must be set in such a way that client of Radium libs will have the same configuration/requirements.

PatatalouiS:
@nmellado @MathiasPaulin  as i said by email, with this PR i can't build Radium-Engine with CLang v11.0 because of :

Capture d’écran 2021-02-10 à 17 29 20

Which is seems logical with the cmake check.

After a successful compilation with gcc v10.0, I went to test lauch one of the examples in ./tests/ExamplesApps and all give me the same result :

Capture d’écran 2021-02-11 à 13 58 47

Or by launching via the 'open' cmd :

Capture d’écran 2021-02-11 à 13 59 03

After trying to remove the added line in CMakeLists.txt, The problem is the logically the same on CLang v11.0, but everythink is OK on gcc v.10.

What I find aberrant about this story is that, I repeat, a simple non-Radium program that uses std :: filesystem can compile and run smoothly with this obscure parameter: -mmacosx-version-min = 10.5

This problem did not exist before the use of std :: filesystem and in the version of radium that I cloned at the beginning of January, It was while updating to be able to work in a group that I had the problem .

nmellado:
> @nmellado @MathiasPaulin as i said by email, with this PR i can't build Radium-Engine with CLang v11.0 because of :

I understood you disabled the check and were able to compile, my bad.

So, as far as I understand, we drop this change, and refer to doc for supported compilers (https://storm-irit.github.io/Radium-Engine/basicsCompileRadium.html): gcc8 and clang 12 on macox.

MathiasPaulin:
@PatatalouiS If I understand : 
you have removed the line
`set(CMAKE_OSX_DEPLOYMENT_TARGET "10.15" CACHE STRING "Minimum OS X deployment version")` from the main `CMakeLists.txt` and you can compile and execute using GNU gcc (9 might be enough, I have this on my macbook) But, as Radium check for AppleClang12, you can't compile with AppleClang11.
With this line, you can't execute anything compiled using gcc as you have MacOs 10.14 and not MacOs 10.15.
For me, there is nothing aberrant with this.

Radium ask for AppleClang 12.0 but if you succeed to compile and execute with AppleClang 11.0, we can change this. But, As I said, we can't ask for SDK 10.15 that is really to restrictive for deployement as we can deploy on MacOs 10.14 (and even less).

So, @PatatalouiS, can you modify the file `cmake/compilerVersion.cmake` to change the requested AppleClang version from `12.0` to `11.0` (on line 4 of this file), then reconfigure/rebuild Radium and its application. if std::filesystem is supported by AppleClang11, this will be enough and we'll change or requisite.

PatatalouiS:
What I found out of place is that I can compile and run a small test program that uses std :: filesystem on MacOS 10.4 with this CLang compiler setting. By changing the cmake / compilerVersion.cmake file the error on filesystem is the same.

What is certain is that I should have updated gcc before warning @nmellado. What motivated me to take contact is the fact that 1 month later, what worked perfectly with my environment no longer works today. There must have been another update in the meantime that I haven't, because I was never warned by the prompt, on my current version, that my compiler was not supported, I was only warned of the std::filesystem error.

So I will now explain to the other MacOS user in my group that we can no longer use CLang unless we update our machines or install gcc.

Sorry for the lost time, it was complicated for us to follow the updates of Radium, The last update having been particularly difficult with the changes of organization of the headers and namespaces. we have been trying for 2 days to get the application we have, to work on again ;-( 

Thank you for your time !

PR #689, opened 2021-02-11
Check if normals exists before accessing them (fix #688)

### LINKED ISSUE #688: Topological Mesh requires normal vectors, but they are optional
A mesh can exists even if the normal vectors are not set.
However, the constructor of `TopologicalMesh` expect (without tasking or requiring it) and read this attribute:
https://github.com/STORM-IRIT/Radium-Engine/blob/a50a3b45a694855e2fbf2a657656a5e98605af34/src/Core/Geometry/TopologicalMesh.inl#L119
This cause a segfault at runtime.

A way to fix this would be to test if the normal buffer is empty, and skip the normal attribute setting in that case.

nmellado:
Bugfix

See #688 

Check if the mesh has normal attributes before copying them

 --

### DIFF
--- a/src/Core/Geometry/TopologicalMesh.cpp
+++ b/src/Core/Geometry/TopologicalMesh.cpp
@@ -166,7 +166,7 @@ struct DefaultNonManifoldFaceCommand {
     inline void initialize( const TriangleMesh& /*triMesh*/ ) {}
     /// \brief Process non-manifold face
     inline void process( const std::vector<TopologicalMesh::VertexHandle>& /*face_vhandles*/ ) {
-        LOG( logWARNING ) << "Invalid face handle returned : face not added (1)";
+        LOG( logWARNING ) << "Invalid face handle returned : face not added";
         /// TODO memorize invalid faces for post processing ...
         ///  see
         ///  https://www.graphics.rwth-aachen.de/media/openflipper_static/Daily-Builds/Doc/Free/Developer/OBJImporter_8cc_source.html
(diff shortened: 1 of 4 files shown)


dlyr:
After testing :
"bow tie" vertex is handled by open mesh and has `TopologicalMesh::isManifold(vertexHandle)` returning true.
"double pyramid" as the example here generate invalid faces and could be handled with the `NonManifoldFaceCommand`
I have added some tests to demonstrate this.
OpenMesh doc is misleading on this point, I'll maybe report this upstream, if I found time.

src/Core/Geometry/TopologicalMesh.cpp
```diff
@@ -162,11 +166,13 @@ void copyAttribToCore( TriangleMesh& triMesh, const HandleAndValueVector<T>& dat
 
 //! [Default command implementation]
 struct DefaultNonManifoldFaceCommand {
+    /// \brief details string is printed along with the message
+    DefaultNonManifoldFaceCommand( std::string details = {} ) : m_details {details} {}
```
nmellado, line 170: I would prefer that we pass a `struct TopologicalMesh::NonManifoldFaceCommandContext` instead of a `std::string`, which will allow us to add more attributes if needed, and without changing the API.

nmellado (commented):
Nice, thanks for the additional change.
Just one suggestion to discuss, and I guess we are good to go.

dlyr:
Well. It’s an internal class. There is no api here  

++ david

nmellado (commented):
Ok my bad, I misread your code.
All good for me.

dlyr:
I suspect a bug in fromWedge conversion with normal ctor, even if we need to get rid of this normal ctor and replace it by the fromWedge version, I prefer to investigate before accepting the PR.

dlyr:
Done.

dlyr (approved):

dlyr (approved):

PR #697, opened 2021-02-21
Add other interger ctor to Index. Fix #666

### LINKED ISSUE #666: Missing constructor Index(unsigned int)
As described below, it would make sense to add an implicit constructor for the Index class, from unsigned integers. That would prevent to generate narrowing conversion issues.

_Originally posted by @nmellado in https://github.com/STORM-IRIT/Radium-Engine/pull/662#discussion_r563943674_

dlyr:

Feature: add cast constructor to Index, to avoid static cast

Index takes only int (aka Index::IntegerType) other type may need static cast
Fix #666

Now one can construct an index from unsigned int, long int, and unsigned long int without casts.
If the value is more than int limit, an invalid index is built.

### DIFF
--- a/src/Core/Utils/Index.hpp
+++ b/src/Core/Utils/Index.hpp
@@ -18,7 +18,7 @@ class RA_CORE_API Index
     constexpr Index( const Index& i );
 
     /// Templated constructor to convert any interger type to index.
-    /// Integer template parameter has to be std::is_integral (static_assert'ed)
+    /// \tparam Integer template parameter has to be std::is_integral (static_assert'ed)
     template <typename Integer>
     explicit constexpr Index( Integer i );
 
--- a/src/Core/Utils/Index.hpp
+++ b/src/Core/Utils/Index.hpp
@@ -16,6 +16,9 @@ class RA_CORE_API Index
     /// Default constructor that allow implicit conversion from integer to Index
     constexpr Index( IntegerType i = s_invalid );
     constexpr Index( const Index& i );
+    template <typename Integer>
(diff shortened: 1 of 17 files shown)


nmellado (approved):

src/Core/Utils/Index.hpp
```diff
@@ -16,7 +16,8 @@ class RA_CORE_API Index
     /// Default constructor that allow implicit conversion from integer to Index
     constexpr Index( IntegerType i = s_invalid );
     constexpr Index( const Index& i );
-    constexpr Index& operator=( long int i );
+    template <typename Integer>
```
nmellado, line 19: Please add static_assert to check if the type is integral, and mention this in the documentation.
Doc: https://en.cppreference.com/w/cpp/types/is_integral

nmellado (changes requested):

src/Core/Utils/Index.hpp
```diff
@@ -16,6 +16,9 @@ class RA_CORE_API Index
     /// Default constructor that allow implicit conversion from integer to Index
     constexpr Index( IntegerType i = s_invalid );
     constexpr Index( const Index& i );
+
+    /// Templated constructor to convert any interger type to index.
+    /// Integer template parameter has to be std::is_integral (static_assert'ed)
```
nmellado, line 21:
```suggestion
    /// \tparam Integer template parameter has to be std::is_integral (static_assert'ed)
```

nmellado (approved):

PR #642, opened 2020-12-28
Engine reorganize directory structure

dlyr:
First tentative to reorganize source files in src/Engine
Most of the changes are scripted, so upcomming comments are "easy" to account for.
When ready, I'll rewrite history, and release the script for include renaming (for third party applications and plugins).

PR #700, opened 2021-02-25
Fix engine namespace

MathiasPaulin:
Following PR #642 and when updating an external application, I found two concerns : 
  The class `ShaderProgram` was predeclared in the wrong namespace in the file `Mesh.hpp` 
    --> Fix the predeclaration
  The class `ViewingParameters` is declared in the namespace `Data` but the file `ViewingParameters.hpp` is in the directory `Rendering` 
    --> The file is moved in this PR

The 2 commits of this PR correspond to the given fixes.


dlyr:
I have added more fixes (since I was wierd the program compiles with this predeclaration, there might be a unneeded include, I saw also an header include at the end of another ... don't know what happens, either my script or clion ...)

dlyr (approved):

dlyr:
When merged, I'll update the automation script.

MathiasPaulin:
> I have added more fixes (since I was wierd the program compiles with this predeclaration, there might be a unneeded include, I saw also an header include at the end of another ... don't know what happens, either my script or clion ...)

You're welcome. This was detected by clang-tidy which tell about ambiguities. I found also a lot of unneeded includes but wants to keep the PR small. I'll do another one that removes all these unneeded includes in a few days (and will run cpp-check also).

dlyr:
perfect, I'm fine with merge.

PR #706, opened 2021-03-05
Small fixes

dlyr:
Here is two small fixes

The first one is on the cmake side
the modification add quotes around one variable where I had some issue when there is two much files in the shader list (I got an error at config time wrong number of parameter)
Other variables deserve quotes, but it's another history.
The second one is in ForwardRenderer, the hack about point clouds adds a geomShader even if the geometry is not a point cloud.
It creates problems with any faceless renderered geometry.

### DIFF
--- a/cmake/radium_setup_functions.cmake
+++ b/cmake/radium_setup_functions.cmake
@@ -431,7 +431,7 @@ function(installTargetResources)
     if (ARGS_FILES)
         set_target_properties(${ARGS_TARGET}
             PROPERTIES
-            RADIUM_TARGET_RESOURCES_FILES ${ARGS_FILES}
+            RADIUM_TARGET_RESOURCES_FILES "${ARGS_FILES}"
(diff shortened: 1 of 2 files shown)


cmake/radium_setup_functions.cmake
```diff
@@ -429,9 +429,10 @@ function(installTargetResources)
         set(ARGS_PREFIX "")
     endif ()
     if (ARGS_FILES)
+        message("[${ARGS_FILES}]")
```
nmellado, line 432: Do we really need this peint ? Or was it for debugging ?

dlyr: I really don't know what you're talking about, like I was debugging by printing messsage ? :smile:

nmellado: :D

nmellado (approved):

PR #707, opened 2021-03-05
Wireframe

dlyr:
Can (should) be merged before #589.

But there is design flaws we need to discuss.
 - Where to store the wireframe meshes.
 - How to control wireframe appearance
 - Do we need better/adapted polygon offset ?

### DIFF
--- a/Shaders/Lines/Wireframe.frag.glsl
+++ b/Shaders/Lines/Wireframe.frag.glsl
@@ -1,14 +1,20 @@
 in vec4 gColor;
 out vec4 fragColor;
 in float pixelWidthDiv2;
-in float slopeY2;
 
-// some sort of https://www.geeksforgeeks.org/anti-aliased-line-xiaolin-wus-algorithm/
-// by reconstructing a rectangle triangle
+const float PI = 3.1415926535897932384626433832795;
+
 // AA is one pixel wide after pixelWidth of fully filled.
-void main() {
-    float d = max( abs( gColor.a ) - pixelWidthDiv2 + 0.5, 0. );
-    float a = clamp( 1 - d * sqrt( 1. + slopeY2 ), 0., 1. );
 
-    fragColor = vec4( vec3( 1., 0., 0. ), a );
+float aa( in float dist ) {
+    float R = sqrt( 2. * .5 * .5 );
+    float s = sign( dist );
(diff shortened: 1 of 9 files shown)


dlyr:
For comparison
wireframe-old
wireframe-new

dlyr:
The way wireframe meshes are generated and stored is not perfect.
The are stored in a map in the renderer, while it's really easy to do, it's prone to memory leaks (well there are a lot on the OpenGL side, since most of GL buffers are not deleted).
Also, the DrawPrimitive app need to be updated.

LineMesh indicies computation could also be done using topological mesh (for triangle now, for poly in a near futur), it's better in term of memory footprint (only one line per edge, guaranteed) while now we may have one line drawn multiple times.

nmellado:
@dlyr you marked this review as ready to review, but it still have the `WIP` label. Shall we review now ?

dlyr:
Yes, in other term, I need help and regarding previous comment.
Ready for review, not for merge ;)

MathiasPaulin:
> The way wireframe meshes are generated and stored is not perfect.

Even if not perfect, it helps a lot and might be easily replaced later (e.g. when multiviewindex will be available). 

> The are stored in a map in the renderer, while it's really easy to do, it's prone to memory leaks (well there are a lot on the OpenGL side, since most of GL buffers are not deleted).

Even if globject lifetime management is hard to follow in the code, and after verification, all OpenGL objects constructed by globjects are automatically deleted when their owing objects got out of scope. For buffers, the documentation about the destructor clearly state that : 
```
/** Automatically deletes the associated OpenGL buffer unless the object was created with an external id.
        \see https://www.opengl.org/sdk/docs/man4/xhtml/gl::glDeleteBuffers.xml
    */
```

And no buffers in Radium is created from an already existing OpenGL buffer.

I verified using OpenGL  profiler and VAO/VBO are deleted as expected (at least on MacOs)

> Also, the DrawPrimitive app need to be updated.

Which kind of update ?

dlyr:
Well DrawPrimitive is a bit out of sync with current RadiumApp init, or maybe I misunderstand recent changes.
Nevertheless I have pointed out an invalid read, and I got some segfault time to time...

dlyr:
Could be done in another PR, as well as the mesh factory thing.

nmellado (approved):
Ready to merge if @MathiasPaulin approves.

PR #708, opened 2021-03-09
Fix topo complex

dlyr:

Bug fix

When converting to Topological Meshes, some non manifold faces are detected, some raise an assert (on purpose according to openmesh doc, see commit messages).

We pre process the index list to skip badly formed faces.
For the current version, it removes the faces (since inputs are only triangles, bad formed faces drops down to lines which are not handled)
When #589 will be merged, it can add faces that "loose" some vertices, but are still relevant.

nope

* **Other information**:

### DIFF
--- a/doc/developer/mesh.md
+++ b/doc/developer/mesh.md
@@ -109,6 +109,17 @@ During manipulation, if two wedges become the same (i.e. for the example on the 
 To merge wedges an explicit call to TopologicalMesh::mergeEqualWedges is needed.
 When a wedge is not referenced anymore (because referencing halfedges have been deleted), it is marked for deletation. When one call Ra::Core::Geometry::TopologicalMesh::garbage_collection(), the marked wedges are eventually deleted and halfedges index are updated accordingly.
 
+During conversion from Core::Geometry::*Mesh to Core::Geometry::TopologicalMesh, non manifold mesh faces are not added but given to an optionally provided user defined functor (see Ra::Core::Geometry::TopologicalMesh::TopologicalMesh and Ra::Core::Geometry::DefaultNonManifoldFaceCommand).
+Some meshes have multiple times the **same** vertices with **different** values for attributes inside a **single** face.
+This kind of situation can either raise degenerated faces, (in case of triangle mesh) or valid faces for some cases for polymesh.
+
+\todo will be supported when TopologicalMesh will support PolyMesh.
+
+In the latter case, the face is cleaned from the duplicated position and added to the TopologicalMesh, using the first encountered wedge value.
+The following figure illustrate this.
+![Mesh to Topo conversion for faces with multiples times the same 3D position](wedge-degen.svg)
+
+
 TopologicalMesh methods and types related to wedges: 
 
 - Ra::Core::Geometry::TopologicalMesh::WedgeData the actual wedge data, with one vector array for each of the supported types (float, Vector2, Vector3, Vector4). The order in these arrays follow the names found in getXXXAttribNames referenced below.
--- a/src/Core/Geometry/TopologicalMesh.inl
+++ b/src/Core/Geometry/TopologicalMesh.inl
@@ -111,9 +111,69 @@ inline TopologicalMesh::TopologicalMesh( const TriangleMesh& triMesh,
             face_wedges[j] = m_wedges.add( wd );
         }
 
-        // Add the face, then add attribs to vh
-        auto fh = add_face( face_vhandles );
-        // In case of topological inconsistancy, face will be invalid ...
+        // take care of degen, see below in method initWithWedge for comments
+        // copied from initFromWedges EXECPT THE TWO LINES WITH COMMENT
+        // x-----------------------------------------------------------------------------------x
+        {
+            auto begin = face_vhandles.begin();
+            if ( face_vhandles.size() > 2 )
+            {
+                auto end       = face_vhandles.end() - 1;
+                auto wedgeEnd  = face_wedges.end() - 1;
+                auto normalEnd = face_normals.end() - 1;
+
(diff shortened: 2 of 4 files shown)


nmellado (commented):
I don't have time for an in-depth review, but my first feeling is that your code is not documented enough: I have seen you refer to another documentation, but in current state it is a bit difficult to follow the code

src/Core/Geometry/TopologicalMesh.inl
```diff
@@ -111,9 +111,72 @@ inline TopologicalMesh::TopologicalMesh( const TriangleMesh& triMesh,
...
+                    end--;
+                    // explicit del wedge (not needed in initWithWedges)
+                    m_wedges.del( *wedgeEnd );
+                    wedgeEnd--;
+                    normalEnd--;
+                }
+                face_vhandles.erase( end + 1, face_vhandles.end() );
+                face_wedges.erase( wedgeEnd + 1, face_wedges.end() );
+                face_normals.erase( normalEnd + 1, face_normals.end() );
+            }
```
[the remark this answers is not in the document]
dlyr: Well, it's line 293 just below
The idea is to remove consecutive references to the same topological vertex
for instance if your face vhandles is 1 2 2, it is not valid on the OpenMesh side.
So I remove duplication in vhandles and update other vectors (normal and wedges) accordingly.
First I remove front / back equal vhandles, for instance in the case 1 2 1
Then I remove consecutive for instance 1 2 2 
This code is already adapted to poly mesh
so 1 1 1 2 2 3 3 3 1 1 becomes 1 2 3

dlyr:
The proposed algo will also work on TopologicalMesh for PolyMesh, in this case the first vertex wedge is used.
wedge-degen

nmellado (approved):

nmellado:
Good for me, once updated wrt `master`

dlyr:
Let me write some documentation and rebase.

doc/developer/mesh.md
```diff
@@ -109,6 +109,14 @@ During manipulation, if two wedges become the same (i.e. for the example on the
 To merge wedges an explicit call to TopologicalMesh::mergeEqualWedges is needed.
 When a wedge is not referenced anymore (because referencing halfedges have been deleted), it is marked for deletation. When one call Ra::Core::Geometry::TopologicalMesh::garbage_collection(), the marked wedges are eventually deleted and halfedges index are updated accordingly.
 
+During convertion from Core::Geometry::*Mesh to Core::Geometry::TopologicalMesh, non manifold mesh faces are not added but given to a optionally provided user defined functor (\see Ra::Core::Geometry::TopologicalMesh::TopologicalMesh).
```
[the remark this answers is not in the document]
MathiasPaulin: Even if documented in the source, saying here what is the main purpose of such a functor might be helpful.

doc/developer/mesh.md
```diff
@@ -109,6 +109,14 @@ During manipulation, if two wedges become the same (i.e. for the example on the
...
+Some meshes have multiple times the **same** vertices with **different** values for attributes inside a single face.
```
MathiasPaulin, line 113: And ? What is the bad effect of such a case? Given a cube, vertices have the same position but with 3 attributes (one par square face). And this situation is not problematic as wedges are built for that, right ?

doc/developer/mesh.md
```diff
@@ -109,6 +109,14 @@ During manipulation, if two wedges become the same (i.e. for the example on the
...
+This kind of situation can either raise degerated faces, (in case of triangle mesh) or valid faces for some cases for polymesh (\todo will be supported when TopologicalMesh will support PolyMesh).
```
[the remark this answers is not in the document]
MathiasPaulin: is it only "one vertex, several attributes" described above that gives this "kind of situation" ?

doc/developer/mesh.md
```diff
@@ -109,6 +109,14 @@ During manipulation, if two wedges become the same (i.e. for the example on the
...
+In the latter case, the face is cleaned from the duplicated position and added to the TopologicalMesh, using the first encountered wedge value.
```
MathiasPaulin, line 115: Why the first. Is there a way, for polymesh, to specify which wedge to take (e.g. The one with the maximum cotangent weight) ?

doc/developer/mesh.md
dlyr: inside a single face

doc/developer/mesh.md
dlyr: arbitrarly ... their is no cotan weight here, the position are the same .... 
the other solution is to drop them to the functor for non manifold faces.

doc/developer/mesh.md
dlyr:
```suggestion
Some meshes have multiple times the **same** vertices with **different** values for attributes inside a **single** face.
```

doc/developer/mesh.md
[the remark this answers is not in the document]
dlyr: Their is no purpose, it's user defined.

doc/developer/mesh.md
dlyr: Also there is a figure to show the problem, it's really related to bad topology, for instance if the top face of the cube is a pentagon with two time the same position with different normals. The rendering will change dramatically depending on triangulation, and it's bad ;)

doc/developer/mesh.md
dlyr: see figure (here in the comment, or [updated in the doc](https://github.com/STORM-IRIT/Radium-Engine/blob/f7e57ca4e7c5ca01cc33e61433a12678f1aa72b6/doc/images/wedge-degen.svg))

MathiasPaulin (approved):

PR #717, opened 2021-03-16
Make scene display setting more clear and usable - update drawPrimitive Demo

MathiasPaulin:
In all the demo applications and Radium apps (PR associated), the method `MainWindowInterface::postLoadFile()` was called each time a scene is set. As the role of this method was not so clear when a scene is built from scratch (without loading a file), This PR rename this function as `MainWindowInterface::prepareDisplay()` and clarifies its reference documentation.
To be more convenient, a new method was added to the `Gui::Viewer` class to prepare the viewer to display a scene and this method is called by `MainWindowInterface::prepareDisplay()` for `Gui::BaseApplication` derived apps.
For other apps, i.e. DrawPrimitiveDemo, this method could be called directly to allow scene creation in a renderer independant way.

Updated packages : 
  - Gui : `MainWindowInterface` and `Viewer`
  - test : all applications now use `MainWindowInterface::prepareDisplay()`
  - test : DrawPrimitiveDemo was update to the correct way of defining custom `RenderObjects` with materials and let the `Viewer` (and hence the active renderer) to manage the `RenderTechnique` that is a renderer-dependent properties of the RenderObjects created by the renderer when calling `Viewer::prepareDisplay()`.

### DIFF
--- a/src/Gui/RadiumWindow/SimpleWindow.cpp
+++ b/src/Gui/RadiumWindow/SimpleWindow.cpp
@@ -6,6 +6,8 @@
 #include <Gui/Viewer/CameraManipulator.hpp>
 #include <Gui/Viewer/Viewer.hpp>
 
+#include <QStatusBar>
+
 namespace Ra {
 using namespace Gui;
 using namespace Engine;
@@ -60,12 +62,9 @@ void SimpleWindow::addRenderer( const std::string&,
     m_viewer->addRenderer( e );
 }
 
-void SimpleWindow::prepareDisplay( const std::string& ) {
+void SimpleWindow::prepareDisplay() {
     m_selectionManager->clear();
-
-    if ( m_viewer->prepareDisplay() ) {
-        emit frameUpdate();
-    }
+    if ( m_viewer->prepareDisplay() ) { emit frameUpdate(); }
 }
 
 void SimpleWindow::cleanup() {
--- a/src/Gui/RadiumWindow/SimpleWindow.cpp
+++ b/src/Gui/RadiumWindow/SimpleWindow.cpp
@@ -60,16 +60,12 @@ void SimpleWindow::addRenderer( const std::string&,
     m_viewer->addRenderer( e );
 }
 
-void SimpleWindow::postLoadFile( const std::string& ) {
-    m_viewer->makeCurrent();
-    m_viewer->getRenderer()->buildAllRenderTechniques();
-    m_viewer->doneCurrent();
+void SimpleWindow::prepareDisplay( const std::string& ) {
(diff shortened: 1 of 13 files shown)


src/Gui/RadiumWindow/SimpleWindow.hpp
```diff
@@ -50,7 +50,7 @@ class RA_GUI_API SimpleWindow : public Ra::Gui::MainWindowInterface
 
   public slots:
     /// Call after loading a new file to let the window resetview for instance.
-    virtual void postLoadFile( const std::string& filename ) override;
+    virtual void prepareDisplay( const std::string& filename ) override;
```
dlyr, line 53: can we get rid of this filename here ?

MathiasPaulin: Sure.

dlyr: Well, why this parameter is needed ? 
If it's on the App side, then overload in app, but lib do not use it ... simply delete ?

MathiasPaulin: Some apps used it to name the root entity. Others use it to generate some informations about scene ... Well, it is now always used and you might forget it if you dont care about scene naming.

src/Gui/RadiumWindow/SimpleWindow.cpp
```diff
@@ -60,9 +62,9 @@ void SimpleWindow::addRenderer( const std::string&,
     m_viewer->addRenderer( e );
 }
 
-void SimpleWindow::prepareDisplay( const std::string& ) {
+void SimpleWindow::prepareDisplay( const std::string& sceneName ) {
+    if ( !sceneName.empty() ) { statusBar()->showMessage( sceneName.c_str() ); }
```
dlyr, line 66: I still don't get the point here ... why simple window display this message ? should not be the code as simple as possible without any superflous stuff ?

dlyr (changes requested):

src/Gui/RadiumWindow/SimpleWindow.cpp
MathiasPaulin: I recall the objective of this PR that is to allow users to build scene without worry about the renderer that will display it.
So, I will definitively not refactor in this PR how the apps interact with the scene after loading a file or creating a scene from scratch. So, the sceneName parameter will remain and is not used in SimpleWindow as before. If this is problematic for you, fell free to either contribute to this PR or propose another PR later.

dlyr (approved):

PR #719, opened 2021-03-17
TopologicalMesh Wedge's round 2.

dlyr:
Second part of the #589, to add some functionalities, the rest is left for futur work.
This PR is to fix bugs only, new update has to be added to #589 .

This PR includes:
- [x] Conversion (backward and forward) between `Core::PolyMesh` and `Core::TopologicalMesh`.
- [x] Remove props management in topological mesh and use only wedges.
- [x] Topological collapse, wedge aware, only tested on triangular faces

Move old TopologicalMesh in deprecated subfolder, is someone wants to have props its helper function, it is still possible until everything is working with wedges.

Topological have both props on halfedges and wedges.

props (renamed as attribs) are stored on wedge only.
One can temporary copy some attribs to openmesh properties to do some computation and update the wedge attribs afterwards.


MathiasPaulin (approved):
I'm fine with this PR.

MathiasPaulin (approved):

PR #721, opened 2021-03-26
Io add custom vertexattribs

nmellado:

Allow to load vertex attributes from a ply file

The ply loader only loads "standard" vertex attributes : position, normal and color

The ply loader also loads all the vertex attributes defined in the ply file

No

* **Other information**:


nmellado:
Example of file to be loaded:  [goursat_0100000APSS0.ply.zip](https://github.com/STORM-IRIT/Radium-Engine/files/6234703/goursat_0100000APSS0.ply.zip)

Next step: use https://github.com/nmwsharp/happly

MathiasPaulin:
> Example of file to be loaded: [goursat_0100000APSS0.ply.zip](https://github.com/STORM-IRIT/Radium-Engine/files/6234703/goursat_0100000APSS0.ply.zip)

The crash is fixed and this file loads well currently.


> Next step: use https://github.com/nmwsharp/happly
Is this really needed now ?

nmellado (commented):
Fine for me.
@dlyr can you please review this PR, we need it for a project.
Thanks

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -74,6 +75,66 @@ struct memory_stream : virtual memory_buffer, public std::istream {
         memory_buffer( first_elem, size_ ), std::istream( static_cast<std::streambuf*>( this ) ) {}
 };
 
+template <typename DataType, typename ContainerType>
+void copyArrayToRadiumcontainer( const uint8_t* buffer, ContainerType& container, size_t count ) {
```
dlyr, line 79:
```suggestion
void copyArrayToContainer( const uint8_t* buffer, ContainerType& container, size_t count ) {
```
I don't get why the container is "Radium" here

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -74,6 +75,66 @@ struct memory_stream : virtual memory_buffer, public std::istream {
...
+    CORE_UNUSED( buffer );
```
dlyr, line 80: Could we avoid these and simple comment or remove args names ?

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -74,6 +75,66 @@ struct memory_stream : virtual memory_buffer, public std::istream {
...
+    CORE_UNUSED( container );
+    CORE_UNUSED( count );
+    LOG( logERROR )
```
dlyr, line 83: Should this error be raised at compile time ? I think so (so use a static_assert for instance)

src/IO/TinyPlyLoader/TinyPlyFileLoader.cpp
```diff
@@ -174,29 +221,49 @@ FileData* TinyPlyFileLoader::loadFile( const std::string& filename ) {
...
-            container.reserve( buffer->count );
-            for ( size_t i = 0; i < buffer->count; ++i )
+            for ( const auto& p : e.properties )
             {
-                container.emplace_back(
-                    floatBuffer[i * 3 + 0], floatBuffer[i * 3 + 1], floatBuffer[i * 3 + 2] );
+                bool exists = usedAttributes.find( p.name ) != usedAttributes.end();
+                if ( !exists )
+                {
+#ifdef OUTPUT_CUSTOM_ATTRIBS
```
dlyr, line 252: could we get rid of this define ?

dlyr (commented):
First round of review. 
I do not get at a first read what is not working yet on master, and how it is handled in this PR.

A test file that fail on current master and now is handled correclty is welcom, or a bit of explanation of what is fixed.
My first guess is that you want to load custom attributes from .ply file to core mesh attribs.

dlyr (approved):

PR #722, opened 2021-03-26
[apps] Fix minimal radium when not using assimp

nmellado:
Bugfix

MinimalApp does not compile without assimp

Protect assimp code: Radium logo is disabled when compiling without assimp.

nop

* **Other information**:


tests/ExampleApps/DrawPrimitivesApp/minimalradium.cpp
```diff
@@ -649,14 +653,18 @@ void MinimalComponent::initialize() {
...
     if ( ENABLE_LOGO )
     {
+
         updateCellCorner( cellCorner, cellSize, nCellX, nCellY );
         updateCellCorner( cellCorner, cellSize, nCellX, nCellY );
 
-        Asset::FileData* data;
+        Asset::FileData* data {nullptr};
+
+#ifdef IO_USE_ASSIMP
```
dlyr, line 662: this test is sufficient right ? we don't need the ifdef for the const bool ...

dlyr (approved):

PR #727, opened 2021-04-01
[github] update gcc according to last ubuntu image

dlyr:
CI hot fix.

Another solution is to stick to 18.04


nmellado:
I would prefer that we stick to the previous version, as we claim compatibility with gcc8, and I guess that it is good to keep it.

dlyr:
After double checking gcc 8 isn't include in any of the virtual environnement.

nmellado:
ok so let switch to to newer compiler then.

nmellado (approved):

Issue #38, opened 2015-08-20
addComponentToEntity system method


Zouch:
Since it's the third "plugin" I'm writing now, I'm a bit tired of always writing the same stuff in my `PluginSystem::addComponentToEntity()` methods :

```
Ra::Engine::Component* MyPluginSystem::addComponentToEntity( Ra::Engine::Entity* entity ) 
{
  uint componentId = entity->getComponentsMap().size();

  std::string componentName = "MyPluginComponent_" = entity->getName() + std::to_string( componentId );
  MyPluginComponent* component = new MyPluginComponent( componentName );

  component->setEntity( entity );
  component->setRenderObjectManager( m_engine->getRenderObjectManager() );

  entity->addComponent( component );
  this->addComponent( component );

  component->initialize();
  return component;
}
```

The only lines that are not common in all the plugins I've written are 
```
std::string componentName = "MyPluginComponent_" = entity->getName() + std::to_string( componentId );
MyPluginComponent* component = new MyPluginComponent( componentName );
```

Maybe we should think about a bit of factorisation with something like
```
class System {
  Component* addComponentToEntity( Entity* entity ) final {
    uint id = entity->getComponentsMap().size(); // -> getComponentsCount() method ?
    Component* component = addComponentToEntityInternal( entity, id );
    // All the set stuff
    return component;
  }

  virtual Component* addComponentToEntityInternal( Entity* entity, uint id ) = 0;
};

class MyPluginSystem : public System
{
  virtual Component* addComponentToEntityInternal( Entity* entity, uint id ) override {
    std::string name = "WhateverMyNameIs" + std::to_string( id );
    MyPlyginComponent* component = new MyPluginComponent( name /*, other stuff */ );
    return component;
  }
};
```

---
(Fun fact : Changing the code would have taken the same time as writing this issue)

louen:
(like)

Zouch:
Done

PR #735, opened 2021-04-06
Merge Skeleton-Based Animation Plugin into Radium-Engine

hoshiryu:

This PR merges the Skeleton-based Animation Plugin into Radium-Engine.
It imports:
- The SkeletonComponent, which manages the animation skeleton (animations, manipulation)
- The SkinningComponent, which manages the skinning process
- The SkeletonBasedAnimationSystem, which manages both Components
- The SkeletonBasedAnimationUI, which allows for user interaction with the Skeleton-Based animation process.

* **Other information**:

There is a joint PR in Radium-Apps that adds the SkeletonBasedAnimationUI to the Sandbox.


MathiasPaulin:
@hoshiryu thanks for this PR we are all waiting for. I'll do a review and test ASAP. 
Nonetheless, I do not see the associated PR on radium apps
> There is a joint PR in Radium-Apps that adds the SkeletonBasedAnimationUI to the Sandbox.

The only one that seems to match is "Add Skeleton-Based animation UI to the SandBox #38" but it is out of date wrt Sandbox status. Can you verify and update your PR if needed ?

hoshiryu:
Effectively, I did a PR against the wrong branch, my bad.
The right PR is present now ;)

src/Core/Animation/DualQuaternionSkinning.cpp
```diff
@@ -1,12 +1,14 @@
 #include <Core/Animation/DualQuaternionSkinning.hpp>
 
+#include <Core/Animation/SkinningData.hpp>
+
 namespace Ra {
 namespace Core {
 namespace Animation {
-void computeDQ( const Pose& pose, const Sparse& weight, DQList& DQ ) {
+
+DQList computeDQ( const Pose& pose, const Sparse& weight ) {
```
nmellado, line 9: What is the benefit of returning local object from the function ? Does this introduce any overhead ?

src/Core/Animation/DualQuaternionSkinning.hpp
```diff
@@ -9,29 +10,55 @@ namespace Ra {
...
+ * \brief Default non-optimized, non-parallel implementation of computeDQ.
+ */
+DQList RA_CORE_API computeDQ_naive( const Pose& pose, const WeightMatrix& weight );
 
-/*
- * DualQuaternionSkinning applies a set of dual quaternions to a given input set of vertices and
- * returns the resulting transformed vertices.
+/**
+ * \brief Applies the given Dual-Quaternions to the given vertices.
+ */
```
nmellado, line 39:
```suggestion
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
 */
```

src/Core/Animation/DualQuaternionSkinning.hpp
```diff
@@ -9,29 +10,55 @@ namespace Ra {
...
+ * \f$\mathbf{p}_i^t = \mathbf{Q}_i(\mathbf{p}_i^0)\f$
+ * , where \f$\mathbf{Q}_i = \frac{\sum_{s\in S} \omega_{is}\mathbf{Q}_s}
+ *                                {\|\sum_{s\in S} \omega_{is}\mathbf{Q}_s\|}\f$
+ *
+ * The skinning of the normal, tangent and bitangent vectors is approximated as:
+ * \f$\mathbf{v}_i^t = \mathbf{Q}_i(\mathbf{v}_i^0)\f$
  *
- * WARNING : in Debug the function will assert if input and DQ size mismatch. In Release will simply
- * crash.
+ * \note Assumes frameData is well sized.
```
nmellado, line 54:
```suggestion
 * \note Assumes frameData is well sized.
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
```

src/Core/Animation/LinearBlendSkinning.hpp
```diff
@@ -1,18 +1,36 @@
...
+ * \brief Applies Linear-Blend skinning to the current frame.
+ *
+ * Linear-Blend skinning deforms the mesh geometry for frame \f$t\f$ as follows:
+ * \f$\mathbf{p}_i^t = \sum_{s\in S}\omega_{is}\mathbf{M}_s\mathbf{p}_i^0\f$
+ *
+ * The skinning of the normal, tangent and bitangent vectors is approximated as:
+ * \f$\mathbf{v}_i^t = \sum_{s\in S}\omega_{is}\mathbf{R}_s\mathbf{v}_i^0\f$
+ *
+ * \note Assumes frameData is well sized.
+ */
```
nmellado, line 27:
```suggestion
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
 */
```

src/Core/Animation/RotationCenterSkinning.hpp
```diff
@@ -2,32 +2,73 @@
...
+ *
+ * The optimal center of rotations are computed as:
+ * \f$\mathbf{p}_i^* = \frac{\sum_{t}\varsigma(\mathbf{w}_i, \mathbf{w}_t)\mathcal{A}_t\mathbf{v}_t}
+ *                          {\sum_{t}\varsigma(\mathbf{w}_i, \mathbf{w}_t)\mathcal{A}_t}\f$
+ * , where \f$\mathbf{w}_i=\begin{bmatrix}\omega_{i0}&\omega_{i1}&\cdots&\omega_{in}\end{bmatrix}\f$
+ * , \f$\mathbf{w}_t = \frac{1}{3}(\mathbf{w}_{t_0}+\mathbf{w}_{t_1}+\mathbf{w}_{t_2})\f$
+ * and \f$\mathbf{v}_t = \frac{1}{3}(\mathbf{p}_{t_0}+\mathbf{p}_{t_1}+\mathbf{p}_{t_2})\f$
+ * , \f$t_j\f$ being the \f$j\f$-th vertex of triangle \f$t\f$ and \f$\mathcal{A}_t\f$ its area.
+ *
+ */
```
nmellado, line 34:
```suggestion
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
 */
```

src/Core/Animation/RotationCenterSkinning.hpp
```diff
@@ -2,32 +2,73 @@
...
+ * , \f$\mathbf{w}_t = \frac{1}{3}(\mathbf{w}_{t_0}+\mathbf{w}_{t_1}+\mathbf{w}_{t_2})\f$
+ * and \f$\mathbf{v}_t = \frac{1}{3}(\mathbf{p}_{t_0}+\mathbf{p}_{t_1}+\mathbf{p}_{t_2})\f$
+ * , \f$t_j\f$ being the \f$j\f$-th vertex of triangle \f$t\f$ and \f$\mathcal{A}_t\f$ its area.
+ *
+ * The skinning of the normal, tangent and bitangent vectors is approximated as:
+ * \f$\mathbf{v}_i^t = \mathbf{R}\mathbf{v}_i^0\f$
+ *
+ *
+ * \note Considers frameData is well sized.
+ */
```
nmellado, line 65:
```suggestion
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
 */
```

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
```diff
@@ -2,49 +2,109 @@
...
+ *                                 {\|\mathbf{s_b}-\mathbf{s_a}\|} - 1)
+ *                           (\mathbf{s_b}-\mathbf{s_a})\f$
+ * , \f$\mathbf{s_*}'=\mathbf{M}_s\mathbf{s_*}\f$ and \f$\mathbf{K}_{is}\f$ is the
+ * twisting rotation, function of the STBS weight \f$e_{is}\f$.
+ *
+ * The skinning of the normal, tangent and bitangent vectors is approximated as:
+ * \f$\mathbf{v}_i^t = \sum_{s\in S}\omega_{is}\mathbf{R}_s\mathbf{K}_{is}\mathbf{v}_i^0\f$
+ *
+ * \note Assumes frameData is well sized.
+ */
```
nmellado, line 54:
```suggestion
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
 */
```

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
```diff
@@ -2,49 +2,109 @@
...
+ * \f$\mathbf{Q}_i = \frac{\sum_{s\in S} \omega_{is}\mathbf{Q}_{is}}
+ *                        {\|\sum_{s\in S} \omega_{is}\mathbf{Q}_{is}\|}\f$
+ * , where \f$\mathbf{Q}_{is}\f$ is the STBS dual quaternion w.r.t. bone \f$s\f$,
+ * computed from the STBS translation and rotation:
+ * \f$\mathcal{T}_{is} = \mathbf{s_a}'+
+ *                       \mathbf{R}_s\mathbf{K}_{is}(e_{is}\mathbf{e}_s-\mathbf{s_a})\f$
+ * and \f$\mathcal{R}_{is} = \mathbf{R}_s\mathbf{K}_{is}\f$
+ *
+ * \note The current pose must be given relatively to the reference pose.
+ * \see R::Core::Animation::relativePose.
```
nmellado, line 75:
```suggestion
 * \see R::Core::Animation::relativePose.
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
```

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
```diff
@@ -2,49 +2,109 @@
...
+ * , where \f$\mathbf{Q}_{is}\f$ is the STBS dual quaternion w.r.t. bone \f$s\f$,
+ * computed from the STBS translation and rotation:
+ * \f$\mathcal{T}_{is} = \mathbf{s_a}'+
+ *                       \mathbf{R}_s\mathbf{K}_{is}(e_{is}\mathbf{e}_s-\mathbf{s_a})\f$
+ * and \f$\mathcal{R}_{is} = \mathbf{R}_s\mathbf{K}_{is}\f$
+ *
+ * The skinning of the normal, tangent and bitangent vectors is approximated as:
+ * \f$\mathbf{v}_i^t = \mathbf{Q}_i(\mathbf{v}_i^0)\f$
+ *
+ * \note Assumes frameData is well sized.
```
nmellado, line 101:
```suggestion
 * \note Assumes frameData is well sized.
 * \warning Parallelized loop inside (using openmp), openmp cannot be used by the caller.
```

src/Engine/Scene/SkeletonComponent.hpp
```diff
@@ -0,0 +1,242 @@
...
+
+    /// Return true is animation ping-pong is on, false otherwise.
+    bool isPingPong() const;
+    /// \}
+
+    /// \name Skeleton display
+    /// \{
+
+    /// Turns xray display on/off for the skeleton bones.
+    void setXray( bool on ) const;
```
nmellado, line 149: I guess the Xray concept is somehow outdated and will be removed later ? It does not seems to fit the actual design of Radium.

src/Gui/BaseApplication.cpp
```diff
@@ -332,10 +335,26 @@ void BaseApplication::engineBaseInitialization() {
...
+    // skeletons and skinning data
+    m_engine->registerSystem( "SkeletonBasedAnimationSystem",
+                              new Ra::Engine::Scene::SkeletonBasedAnimationSystem,
+                              defaultSystemPriority );
 }
+
 void BaseApplication::engineOpenGLInitialize() {
     // initialize here the OpenGL part of the engine used by the application
     m_engine->initializeGL();
+
```
nmellado, line 348: I find the following changes very intrusive.
Can we add a state `m_skinning_enabled` in the Engine, and protect all skinning-related stuff by this state ?
This state could be set to false to disable these things (e.g. in command line applications, or any other project not related to animation).

nmellado (commented):
First round of review, I didn't see anything critical, except the modification of the 'Engine' (see comment)

src/Core/Animation/DualQuaternionSkinning.cpp
MathiasPaulin: no overhead due to RVO (or copy elision). see https://en.cppreference.com/w/cpp/language/return#Notes

src/Core/Animation/DualQuaternionSkinning.hpp
MathiasPaulin: Well, support for parallel nested loop is available on OMP but is not active by default. According to the spec of OpenMP, on nested parallel loop, only the englobing loop is parallelized except if the (deprecated in openmp 5.0) OMP_NESTED  https://www.openmp.org/spec-html/5.1/openmpse66.html#x332-5070006.9 variable is set. So here, the doc can be informational but no need to set a warning as thereis no warning. If the caller is inside a parallel lopp, the inner loop is not parallelized and this change nothing for the caller state.

src/Core/Animation/DualQuaternionSkinning.hpp
MathiasPaulin: I propose to replace thes \warning by a simple note `\note Parallelized loop inside (using openmp).` as the the caller can used openmp.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
```diff
@@ -0,0 +1,142 @@
...
+    m_xrayOn( false ) {}
+
+// System Interface
+
+void SkeletonBasedAnimationSystem::generateTasks( Ra::Core::TaskQueue* taskQueue,
+                                                  const Ra::Engine::FrameInfo& frameInfo ) {
+    for ( auto compEntry : m_components )
+    {
+        // deal with AnimationComponents
+        if ( compEntry.second->getName().compare( 0, 3, "AC_" ) == 0 )
```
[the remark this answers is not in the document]
MathiasPaulin: Not in the documentation, and only something internal we can ignor until we hit a problem with this, and historical hypothesis that should be removed. This results from the old (and deprecated/removed?) directive "No RTTI nor dynamic_cast in the code". I can do that without looking for name and we will consider how to design the skeleton-related components interface for a next PR ...

src/Engine/Scene/SkeletonComponent.hpp
MathiasPaulin: XRay concept is quite outdated but is still managed by the `Renderer` interface. Modifying the way bones are rendered is out of scope of this PR and can be addressed later. I add a todo for this.

src/Gui/BaseApplication.cpp
MathiasPaulin: Base application is a QtApplication and is not used in CLI based on Radium Core only(do you have some usecases of CLI that used BaseApplication ? ). CLI that uses Engine neither use `BaseApplication` and add explicitely the needed systems. So, no intrusion of this modification other than in graphical `BaseApplication` derived apps (Sandbox ?)

For Graphical app that do not bother about animation, this add no overhead, except a very small at load time. Only when loading animated scenes AND activating the play state of the engine there generate something.
But we can envision to modify the interface of `BaseApplication::initialize` by adding a "systems" parameter so that the user can ask for the system they want. But this is out of scope of this PR.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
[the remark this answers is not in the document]
nmellado: Ok fine, just wanted to be sure.

src/Core/Animation/DualQuaternionSkinning.cpp
nmellado: Hum ok, fine.

src/Core/Animation/DualQuaternionSkinning.hpp
nmellado: Does this work if the nested loop is in a library, and the caller dynamically links with this library ?

src/Gui/BaseApplication.cpp
nmellado: Still, BaseApplication and Engine are supposed to be general, and I don’t like the idea to force loading ressources for something MyApllication will never do.
If you don’t like the idea of having a state, why not initialize the texture only when the skinning is called the first time ?

src/Core/Animation/DualQuaternionSkinning.hpp
MathiasPaulin: I think so. do you have a test for that ?

nmellado: Not really.
It’s just that I remember having problems with this long time ago, I guess openmp changed in the meantime. On windows maybe.

src/Gui/BaseApplication.cpp
MathiasPaulin: You're partly wrong. BaseApplication is clearly not general. It defines an application with a Gui (even minimalistic) and that own a viewer (an opengl windows to render the scene). So it is very specific to Gui/openGL application.
For CLI application, even those that rely on Engine, the BaseApplication class is not used.

I like the idea of asking only for needed components. But what I said is that it is **out of scope of this PR** as it will incur changes in several interfaces not related to SkeletonBasedAnimation. 

Here, the skeletonanimationsystem is only added for applications derived from BaseApplication, i.e. Graphical applications with OpenGL windows ... and there is no overhead in your application if you do not use the animationsystem. 

So yes, definitely, we need to be able to asks for the systems to be used when creating a graphical interface. But again, it is out of scope of this PR and you can open an issue/feature request on that, so that we'll remember and can solve this in another PR.

And I don't understand why you speak about texture, it is generated only after openGL initialization so that the attached systems can be used and the openGL representation of the texture is generated only at the first use. Here, we just register the texture to the texture manager. And, if you think it is not related to BaseApplication, you're right but for now, it works like this and we'll put elsewhere in the PR that manage the state.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
```diff
@@ -0,0 +1,149 @@
...
+SkeletonBasedAnimationSystem::SkeletonBasedAnimationSystem() : System(), m_xrayOn( false ) {}
+
+// System Interface
+
+void SkeletonBasedAnimationSystem::generateTasks( Core::TaskQueue* taskQueue,
+                                                  const FrameInfo& frameInfo ) {
+    for ( auto compEntry : m_components )
+    {
+        // deal with AnimationComponents
+        if ( auto animComp = dynamic_cast<SkeletonComponent*>( compEntry.second ) )
```
hoshiryu, line 32: So if I understand well, dynamic_cast are "kind of" authorized.
Since the goal here is only to separate `SkeletonComponent` from `SkinningComponent`, would it be nicer to have `m_components` registering one kind, and another `Component` list registering the other kind?

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Skeleton display
+
+void SkeletonBasedAnimationSystem::setXray( bool on ) {
+    m_xrayOn = on;
+    for ( const auto& comp : m_components )
+    {
+        if ( comp.second->getName().compare( 0, 3, "AC_" ) == 0 )
+        { static_cast<SkeletonComponent*>( comp.second )->setXray( on ); }
```
hoshiryu, line 131: b.t.w. here again we distinguish the `Component` kind through its name.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+bool SkeletonBasedAnimationSystem::isXrayOn() {
+    return m_xrayOn;
+}
+
+void SkeletonBasedAnimationSystem::toggleSkeleton( const bool status ) {
+    for ( const auto& comp : m_components )
+    {
+        if ( comp.second->getName().compare( 0, 3, "AC_" ) == 0 )
+        { static_cast<SkeletonComponent*>( comp.second )->toggleSkeleton( status ); }
```
hoshiryu, line 143: b.t.w. here again we distinguish the Component kind through its name.

src/Engine/Scene/SkeletonComponent.cpp
```diff
@@ -0,0 +1,456 @@
...
+        const auto& times = boneAnim.getTimes();
+        startTime         = std::min( startTime, *times.begin() );
+        endTime           = std::max( endTime, *times.rbegin() );
+    }
+    return endTime - startTime;
+}
+
+std::pair<Scalar, Scalar> SkeletonComponent::getAnimationTimeInterval() const {
+
+    if ( m_animations.empty() ) { return {0_ra, 0_ra}; }
```
hoshiryu, line 222:
```suggestion
std::pair<Scalar, Scalar> SkeletonComponent::getAnimationTimeInterval() const {
    if ( m_animations.empty() ) { return {0_ra, 0_ra}; }
```

src/Engine/Scene/SkinningComponent.hpp
```diff
@@ -0,0 +1,246 @@
...
+    inline SkinningType getSkinningType() const { return m_skinningType; }
+
+    /// Sets the method to use to skin the normal, tangent and binormal vectors.
+    void setNormalSkinning( NormalSkinning normalSkinning );
+
+    /// Returns the current method used to skin the normal, tangent and binormal vectors.
+    inline NormalSkinning getNormalSkinning() const { return m_normalSkinning; }
+
+    /// Toggles use of smart stretch.
+    void setSmartStretch( bool on );
```
hoshiryu, line 106: Shall we switch the name from `smart stretch` to `pseudo IK` ?

src/Gui/BaseApplication.cpp
hoshiryu: Could it be simpler to just add an OpenGL initialization method in the Systems?

src/Gui/SkeletonBasedAnimation/SkeletonBasedAnimationUI.cpp
```diff
@@ -0,0 +1,475 @@
...
+            if ( animComp->getName().compare( 0, 3, "AC_" ) == 0 )
+            {
+                auto skel   = static_cast<Ra::Engine::Scene::SkeletonComponent*>( animComp.get() );
+                auto [s, e] = skel->getAnimationTimeInterval();
+                startTime   = std::min( startTime, s );
+                endTime     = std::max( endTime, e );
+            }
+        }
+        m_timeline->onChangeStart( startTime );
+        m_timeline->onChangeEnd( endTime );
```
hoshiryu, line 205: Isn't this redundant with what is done in `SkeletonBasedAnimationSystem::handleAssetLoading`? (I would use only the one here)

src/Engine/Scene/SkeletonComponent.hpp
```diff
@@ -0,0 +1,243 @@
...
+    /// \{
+
+    /// Updates the skeleton pose as the pose corresponding to time \p time.
+    void update( Scalar time );
+
+    /// Returns the current time of animation.
+    Scalar getAnimationTime() const;
+
+    /// Returns the duration of the current animation.
+    Scalar getAnimationDuration() const;
```
hoshiryu, line 121: Since there is `getAnimationTimeInterval`, I think we should get rid of this one (which is not used anymore b.t.w.)

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
```diff
@@ -2,49 +2,112 @@
...
+ * \f$\mathbf{Q}_i = \frac{\sum_{s\in S} \omega_{is}\mathbf{Q}_{is}}
+ *                        {\|\sum_{s\in S} \omega_{is}\mathbf{Q}_{is}\|}\f$
+ * , where \f$\mathbf{Q}_{is}\f$ is the STBS dual quaternion w.r.t. bone \f$s\f$,
+ * computed from the STBS translation and rotation:
+ * \f$\mathcal{T}_{is} = \mathbf{s_a}'+
+ *                       \mathbf{R}_s\mathbf{K}_{is}(e_{is}\mathbf{e}_s-\mathbf{s_a})\f$
+ * and \f$\mathcal{R}_{is} = \mathbf{R}_s\mathbf{K}_{is}\f$
+ *
+ * \note The current pose must be given relatively to the reference pose.
+ * \see R::Core::Animation::relativePose.
```
hoshiryu, line 76: Shall we internalize the use of `Ra::Core::Animation::relativePose`, and remove the relative pose from the inputs?

src/Engine/Scene/SkinningComponent.cpp
```diff
@@ -0,0 +1,505 @@
...
+            const auto& tH      = m_refData.m_referenceMesh.getAttribHandle<Vector3>( tangentName );
+            const auto& tangents = m_refData.m_referenceMesh.getAttrib( tH ).data();
+            Vector3Array bitangents( normals.size() );
+#pragma omp parallel for
+            for ( int i = 0; i < int( normals.size() ); ++i )
+            {
+                bitangents[i] = normals[i].cross( tangents[i] );
+            }
+            m_refData.m_referenceMesh.addAttrib( bitangentName, std::move( bitangents ) );
+        }
```
hoshiryu, line 137: Shall we have tangent and bitangent vectors computation in `Core` ?
This might be useful for other purposes than just skinning.
Also, there might be better ways to compute them, like consistent and continuous ways using curvature or something else?

src/Engine/Scene/SkinningComponent.cpp
```diff
@@ -0,0 +1,505 @@
...
+    if ( !areEqual( currentPose, prevPose ) || m_forceUpdate )
+    {
+        m_forceUpdate            = false;
+        m_frameData.m_doSkinning = true;
+        m_frameData.m_frameCounter++;
+
+        const auto tH = m_refData.m_referenceMesh.getAttribHandle<Vector3>( tangentName );
+        const Vector3Array& tangents = m_refData.m_referenceMesh.getAttrib( tH ).data();
+        const auto bH = m_refData.m_referenceMesh.getAttribHandle<Vector3>( bitangentName );
+        const Vector3Array& bitangents = m_refData.m_referenceMesh.getAttrib( bH ).data();
```
hoshiryu, line 223: Is there a nice way to reach the tangent and bitangent vectors while in `Core` (i.e. without needing `Engine::Data::Mesh::getAttribName( VERTEX_TANGENT )`?
This would simplify all skinning functions' parameters set, and might allow to move tangent and bitangent vectors computation into `Core`.

hoshiryu (commented):
I still have some questions

src/Gui/BaseApplication.cpp
MathiasPaulin: This is what I plan to do in a next PR, alongside the selective instanciation of systems (this might modify a lot of things)

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
MathiasPaulin: yes, `dynamic_cast` can be used in a few situations and there is nothing bad about it if it is not too frequent (in which case, there is a design problem)
Your proposal is an option, yes, and deserve to be examine. But `m_component` is inherited from `System` class, this might have an impact on the way components managed by a system can be accessed, mainly for "uniformity" access to the components.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
MathiasPaulin: oups ... will be corrected in the next push, and might be removed in a near future when XRay state willl be handle differently.

src/Engine/Scene/SkeletonBasedAnimationSystem.cpp
hoshiryu: Yes you are right.
Maybe having two additional lists, one per kind of component? But that's redundant data (only pointers), made to avoid a `dynamic_cast`. Not sure it's better.

src/Engine/Scene/SkeletonComponent.hpp
MathiasPaulin: We can keep it for now as there is no impact, no cost, and it can be used in several uses cases.

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
MathiasPaulin: Can you develop ?

src/Engine/Scene/SkinningComponent.cpp
MathiasPaulin: As this operation is done also on some fileloaders, yes, it is planned to bring such algorithms to the core. This will be done in a next PR as soon as we found a good design for this.

src/Engine/Scene/SkeletonComponent.hpp
hoshiryu: well, one just have to to the diff between the time interval bounds...
Plus it's a remnant from the old way to manage time in the animation system

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
hoshiryu: I mean, since we can fetch the poses from the skeletons, the call to `relativePose` could be made internally by the function, and it will clean the parameter list a bit.
I think it will be cleaner, since the caller would not have to deal with it.

src/Engine/Scene/SkinningComponent.cpp
MathiasPaulin: all attributes are managed the same way, except, position that is mandatory, and normal that is almost mandatory (and even tis one might be manage as others). So, we will always need the name to access an attribute array.

hoshiryu: Yes I understand that, but why isn't the name for some "standard" geometric attributes defined in `Core`, allowing the `Core` lib to create/manipulate them directly? why is there only the normal available?

dlyr: To complete, each attributes are managed the same way on the Core side,
there is a reminiscence of old fashioned way to handle attributes for normals.
Then we need to define some semantics with a naming convention, this semantic depends on the usage you want to do, and can even use "translators" from core names to engine names.

I haven't read everything in the PR, but we can imagine moving the "standard names" from Engine to Core

src/Gui/BaseApplication.cpp
MathiasPaulin: I tried to fix this and it is not so direct. So, This will be solved in another PR.

src/Engine/Scene/SkinningComponent.hpp
MathiasPaulin: todo added on the code. Will be done in a next PR

src/Gui/SkeletonBasedAnimation/SkeletonBasedAnimationUI.cpp
MathiasPaulin: This is not redundant. In `SkeletonBasedAnimationSystem::handleAssetLoading` the engine is configured. here, the timeline is updated. As the engine as no information about timeline, this must be done like this.

src/Core/Animation/StretchableTwistableBoneSkinning.hpp
hoshiryu: @MathiasPaulin could you pleasu also remove the `see relativePose` comment line? thanks

MathiasPaulin (changes requested):
During test of this PR, I found the following problems that should be fixed before merging into master :
- Center of Rotation Skinning is broken (assert in topomesh raised when computing CoR)
- Unable to remove a skeleton from the scene tree (segfault)

hoshiryu:
Regarding the segfault due to the suppression of the `SkeletonComponent`, it should also happen when suppressing the `SurfaceMeshComponent`.
That must be due to the fact that the `SkinningComponent` relies on them to register some getters and setters in its initialization,
which will be used during task execution.
Thus, whenever one of them is deleted, those getters and setters are not valid anymore, which makes the app crash.

What would be needed is a listener on the deletion of these components, which would either:
- delete also the `SkinningComponent`, considering that if the components we rely on are deleted, so must be the ones that rely on them;
-  invalide the getters and setters on these components, with the idea that one may want to replace them with another (e.g. skeleton change, mesh change for higher/lower resolution, ...);

That mechanism would be highly valuable for the management of the scene.

doc/developer/animation.md
```diff
@@ -21,8 +21,8 @@ There are two ways data can be animated using `Ra::Core::Animation::KeyFramedVal
     struct MyStruct {
```
dlyr, line 21: Theses snippets could be moved as a unit tests and referenced here.

hoshiryu:
Regarding the segfault in the CoR computation, it happens because the mesh is not closed.
To compute the CoRs, we need to apply edge_split operations on the mesh in order to reduce the skinning weight difference between 2 neighbour vertices (cf paper: https://studios.disneyresearch.com/wp-content/uploads/2020/04/Real-time-Skeletal-Skinning-with-Optimized-Centers-of-Rotation.pdf).
However, the edge_split operation, as implemented now in `TopologicalMesh`, does not work when the edge is a boundary edge, i.e. has only one face attached to it, as it is mentioned in a documentation note.
@dlyr Is there a reason to prevent this operation to be performed on such an edge? If so, we should have another implementation targetting these edges; if not, we must make it work.

doc/developer/animation.md
hoshiryu: This is out of the scope of this PR, but sure that would be great.

dlyr:
Concerning the splitEdge for boundary, there is no particular reason (a part from time to do it, along with corresponding tests ...)
I move this to the top of the todo list

doc/developer/animation.md
dlyr: Well, I'll do it, because the current code in the doc do not compile (mistyped param name, struct could is a reserved keyword, .hpp missing in headers ... I have started but still some work to do)

dlyr: done.

dlyr:
@hoshiryu splitCopy should work with boundary, I have done a "visual" test with DrawPrimitiveApp, but no unit tests yet.

hoshiryu:
@ dlyr @MathiasPaulin I tested the topomesh with the Astroboy model, the CoR computation runs fine.

MathiasPaulin:
>  I tested the topomesh with the Astroboy model, the CoR computation runs fine.
Seems to work also with the model I just send to you. When loading  with the dedicated plugin as assimp do not load the animations and build a wrong skeleton for gltf 2.0 scenes ...

MathiasPaulin (commented):
Seems to work. Quite slow on some scenes, smart_stretch is quite problematic (do not work for all skeletons) but can be activated by the user if needed. Still the problem of sene editing that make the lib to crash but can be fixed in another PR.
Another review is welcome (I'll do one more) before merging.

dlyr:
And after these review, we need to cleanup history (I can do that when appropriate).

MathiasPaulin (dismissed):

nmellado:
@dlyr do you want to clean the commit history ?
I am ok to keep it as is (don't have time to clean myself).

MathiasPaulin:
> @dlyr do you want to clean the commit history ?
> I am ok to keep it as is (don't have time to clean myself).

We ca either clean the history or squash and merge ... what do you prefer ?

MathiasPaulin (dismissed):

dlyr (changes requested):
At least for Core structure, new code should be covered by unit tests.

MathiasPaulin (changes requested):
The texture file Influence0.png, used to visualize the weights is managed in such a way that the Skeleton System generates an error message when used on apps not using Qt. This texture should be added as a Radium Resource of the libEngine.

hoshiryu:
I thought about the addition of the keyframes when manipulating the skeleton while in Pseudo-IK mode, and it is not working well.

For example, using the Capsule model, I checked what happens when moving from the refPose to the "up" pose in Forward mode, then going back to the rest pose, and finally to the "up" pose in Pseudo-IK mode:
setup

When registering the `KeyframedValueController` into the `Timeline` in the `SkeletonBasedAnimationUI`, the `insertionCallback` is:
```
[&anim, j, skel]( const Scalar& t ) {
    anim[j].insertKeyFrame( t, skel->getPose( HandleArray::SpaceType::LOCAL )[j] );
};
```
which produces this effects during animation, the forward part is going fine, but the Pseudo-IK part is not:
old

This is due to the fact that the Pseudo-IK scheme also modifies the parent's transform, which is not captured when inserting the keyframe.

Thus I tried the following `insertionCallback`, taking into account the fact that the parent's pose might be needed:
```
[&anim, j, skel]( const Scalar& t ) {
    int parent = skel->m_graph.parents()[j];
    if ( parent != -1 )
        anim[parent].insertKeyFrame( t, skel->getPose( HandleArray::SpaceType::LOCAL )[parent] );
    anim[j].insertKeyFrame( t, skel->getPose( HandleArray::SpaceType::LOCAL )[j] );
}
```
which produces this effects during animation, the forward part is still fine (not shown), but the Pseudo-IK part, while going to the desired keyframe, is badly interpolated (middle image):
new

What I mean by "badly interpolated" is that the edited movement for the bone is only a translation upwards, but when animating it describes an arc of a circle due to the mix of the interpolation of the parent's rotation and the bone's translation.

MathiasPaulin:
> Shall I update the `insertionCalback`?

Thanks for this detailed discussion. 
Do not update the `insertionCalback` as it does not solve the problem.
"pseudo_IK" is still an experimental mode and not the default one. 
So, I propose that, as soon as this PR is merged, you open an issue about "pseudo_IK" with the above discussion. 
Then, once the issue is opened, we will remind that there is a problem with  "pseudo_IK" and you will have time to experiment and propose a solution that correctly fix the problem.

MathiasPaulin (approved):

nmellado:
@dlyr are you still requiring changes on this PR ?

dlyr:
Testing in progress.

dlyr (approved):

nmellado:
Thanks for the hard work.
@hoshiryu please don't forget to write an issue from https://github.com/STORM-IRIT/Radium-Engine/pull/735#issuecomment-830639263

PR #740, opened 2021-04-14
Camera near far, split data/compononent

### LINKED ISSUE #181: Camera management : automatic and optimal znear/zfar
I've found a little bug when playing around with the Radium's MinimalApp.

When you replace the shader used by the MinimalComponent (BlinnPhong) by any other shader (Default for example), the spinning cube is blinking dark.

bug

### LINKED ISSUE #661: Camera split Data and Scene part
Referring discussion in #642 
Camera should be refactored to a Data part (in Core namespace) and a component part (in Scene namespace)

Data part should handle data storage,
Scene Component part, handle system communication.

grandch:

Feature fix

Constant zNear and zFar. Compute aabb for each frame.
Fix #181 
Fix #661

Compute optimal zNear and zFar. Cached aabb.

zRange isn't constant.

* **Other information**:


dlyr:
Test do not pass on macos, let's postprone this issue after the PR is ready for merge.
First of all it could be great to have feedback on the new fonctionality and it's implementation.

src/Core/Asset/Camera.cpp
```diff
@@ -1,79 +1,66 @@
...
+    if ( c.isApprox( Core::Vector3::Zero() ) && d < 0.0 )
+    { T.rotate( Core::AngleAxis( Core::Math::PiDiv2, Core::Vector3::UnitY() ) ); }
+    else
+    { T.rotate( Core::Quaternion::FromTwoVectors( d0, d1 ) ); }
+    applyTransform( T );
 }
 
-void Camera::applyTransform( const Core::Transform& T ) {
-    CORE_ASSERT( m_RO, "Camera's render object must be initialize with Camera::intialize()" );
+void Camera::resize( Scalar width, Scalar height ) {
```
MathiasPaulin, line 50:
```suggestion
void Camera::resizeViewport( Scalar width, Scalar height ) {
```

src/Core/Asset/Camera.cpp
```diff
@@ -164,14 +133,17 @@ void Camera::fitZRange( const Core::Aabb& aabb ) {
     adaptRange( maxAabb[0], minAabb[1], maxAabb[2] );
     adaptRange( maxAabb[0], minAabb[1], minAabb[2] );
 
-    // ensure a minimum depth range
-    Scalar range = ( m_zFar - m_zNear ) / 100_ra;
-    m_zNear      = std::max( range, m_zNear - range );
-    m_zFar       = std::max( 2_ra * range, m_zFar + range );
+    if ( m_zNear < m_minZNear ) { m_zNear = m_minZNear; }
+    if ( ( m_zFar - m_zNear ) < m_minZRange ) { m_zFar += m_minZRange - ( m_zFar - m_zNear ); }
```
MathiasPaulin, line 137:
```suggestion
    if ( ( m_zFar - m_zNear ) < m_minZRange ) { m_zFar =  m_zNear + m_minZRange; }
```

src/Core/Geometry/TriangleMesh.cpp
```diff
@@ -36,6 +40,11 @@ void AttribArrayGeometry::colorize( const Utils::Color& color ) {
     getAttrib( colorAttribHandle ).setData( Vector4Array( vertices().size(), color ) );
 }
 
+void AttribArrayGeometry::invalidateAabb() {
```
MathiasPaulin, line 43: Why is this method not declared virtual in `AbstractGeometry` that has the `computeAabb` and `getAabbObservable`method declaration ?

src/Engine/RadiumEngine.cpp
```diff
@@ -57,6 +58,12 @@ void RadiumEngine::initialize() {
 
     m_loadedFile.reset();
     Scene::ComponentMessenger::createInstance();
+
+    auto cameraManager = new Scene::DefaultCameraManager();
+    cameraManager->initialize();
+    Ra::Engine::RadiumEngine::getInstance()->registerSystem( "DefaultCameraManager",
```
MathiasPaulin, line 64:
```suggestion
    Ra::Engine::RadiumEngine::getInstance()->registerSystem( "Camera Manager",
```

MathiasPaulin: This recall me that there is no `unregisterSystem` in the Engine class. If an application wants to change this camera manager and replace by its own ? How can this be done ?

src/Engine/Rendering/ForwardRenderer.cpp
```diff
@@ -51,9 +53,11 @@ void ForwardRenderer::initializeInternal() {
...
     initBuffers();
 
-    auto cameraManager = new Scene::DefaultCameraManager();
-    Ra::Engine::RadiumEngine::getInstance()->registerSystem( "DefaultCameraManager",
-                                                             cameraManager );
+    auto cameraManager = static_cast<Ra::Engine::Scene::CameraManager*>(
+        Engine::RadiumEngine::getInstance()->getSystem( "DefaultCameraManager" ) );
+
+    auto comp = cameraManager->getCamera( 0 );
+    comp->getCamera()->resize( m_height, m_width );
```
MathiasPaulin, line 60: Well, it is strange to do that here. Why to get the camera manager only to resize the camera number 0 ? If this is really needed, why to not doing this in the `resizeInternal` method ?

src/Engine/Rendering/RenderObject.cpp
```diff
@@ -73,10 +76,12 @@ const std::string& RenderObject::getName() const {
 
 void RenderObject::setVisible( bool visible ) {
     m_visible = visible;
+    invalidateAabb();
 }
 
 void RenderObject::toggleVisible() {
     m_visible = !m_visible;
+    invalidateAabb();
```
MathiasPaulin, line 84: Not sure about his. The object is not visible but still belong to the scene/engine. So, its aabb might be used to compute the scene aabb no?

src/Engine/Scene/CameraComponent.cpp
```diff
@@ -0,0 +1,116 @@
...
+    auto m = std::make_shared<Data::LineMesh>( m_name + "_mesh" );
+    m->loadGeometry( std::move( triMesh ) );
+    // Create the RO
+    auto mat              = Core::make_shared<PlainMaterial>( m_name + "_Material" );
+    mat->m_color          = {1_ra, .5_ra, 0_ra, 1_ra};
+    mat->m_perVertexColor = false;
+    Rendering::RenderTechnique rt;
+    auto cfg = Data::ShaderConfigurationFactory::getConfiguration( "Plain" );
+    rt.setConfiguration( *cfg );
+    rt.setParametersProvider( mat );
```
MathiasPaulin, line 59: When building a component, one must be agnostic to the renderer. RenderTechniques are specific to the renderer. Just define a Plain Material here and the camera should be visualized with all the renderer (or define it as a Debug or UI object). See GeometryComponent to build a RenderObject usable by any renderer.

src/Gui/BaseApplication.cpp
```diff
@@ -312,9 +313,16 @@ void BaseApplication::initialize( const WindowFactory& factory ) {
...
+            /// \todo fix this, I don't get this code (d.),
+            /// updated using cameraManager.
             auto entity = *( m_engine->getEntityManager()->getEntities().rbegin() );
-            auto camera = static_cast<Engine::Scene::Camera*>( entity->getComponents()[0].get() );
-            m_viewer->setCamera( camera );
+            auto camera =
+                static_cast<Engine::Scene::CameraComponent*>( entity->getComponents()[0].get() );
+            auto cameraManager = static_cast<Ra::Engine::Scene::CameraManager*>(
+                Engine::RadiumEngine::getInstance()->getSystem( "DefaultCameraManager" ) );
+
```
MathiasPaulin, line 324: This code set the first loaded camera as active. This seems not needed now as CameraManager could do that. right ?

MathiasPaulin (commented):
Firs (uncomplete) review. More to come (after some discussions)

src/Core/Asset/Camera.cpp
dlyr: right, we fixed this one on another branch, I'll do a cherry pick for this PR.,

src/Engine/Rendering/ForwardRenderer.cpp
dlyr: Well, the init pipeline is not very clear. 
The idea is that camera 0 is the default rendering camera, the only one that need to have a viewport size corresponding to the actual viewport. 
Other camera's can keep their loaded size. 
When switching camera's, only data is copied from a given camera to camera 0.

src/Gui/BaseApplication.cpp
dlyr: right,

dlyr: The question is how do we now that getComponents()[0] is a camera ?
Does it rely on the file given that contains only a camera ?

src/Engine/RadiumEngine.cpp
dlyr: Good question, should we do this in the current PR ?

src/Engine/Rendering/ForwardRenderer.cpp
MathiasPaulin: OK, so, it should be better to do that on the base class Renderer, and only into the `resize` methods.
`initialize` is called only once when a renderer is added to the viewer (or any object that manages the rendering) and explicitly call the `resize` method with the correct size. 
The base `resize` method from the class Renderer, who call the concrete renderer `resizeInternal` method, is called each time the viewport change. So this is the place where Camera should be resized.
ForwardRenderer is just a renderer among others and should not do things other than just managing its rendering sequence and renderTechnique. All things that should be common to any renderer should be done in the base RendererClass.

dlyr: Also in the current version, setViewport (the new resize name ;)) is called from the Viewer.
Do you think it's better to let the renderer control camera viewport size ?
On my side I think yes, but it recognize the renderer to be in charge of the camera management.

src/Gui/BaseApplication.cpp
MathiasPaulin: Well, it's more of a hack but it's actually very clear :p.
File associated to the "camera" parameter are expected to contains only cameras.
So the entity access with `rbegin` that is the last entity added to the Engine is the entity containing the loaded cameras. Then, we are sure that the requested component is a camera.
Nevertheless, I think this kind of hypothesis is at the base of several nightmares one can have when trying to understand  (and fix) some bugs that appear due to such undocumented, non robust and risky operations.

src/Engine/Rendering/ForwardRenderer.cpp
MathiasPaulin: I prefer that the renderer just use the given camera and is not in charge of its management. If you can put any camera management method outside of the renderer (even the base class) and just make it to render using the camera (or the associated matrices as done actually), it should be better.

MathiasPaulin: But we have to think about something. Do we pass the camera to the "render" method or do we still pass some "ViewingParameters" that are disconected from the camera and contains only the associated matrices (viewMatrix and projMatrix) ?

dlyr: The last point is clearly out of scope of this PR, 

The question is : when/who do resizee the camera ?
The Viewer ?
The Renderer (I mean Ra::Engine::Renderer) ?

MathiasPaulin: The viewer

src/Engine/RadiumEngine.cpp
dlyr: I do not see any convention for system naming.
In current source code of Radium Lib, and plugins I found:
`
"GeometrySystem"
"DefaultLightManager"
"DefaultCameraManager"
"Minimal system"
"AnimationSystem"
"FakeToonSystem"
"DummySystem"
"MeshPaintSystem"
"LaplacianSmoothingSystem"
`
I propose to keep the current name for the moment until a convention is defined.

dlyr:
After some checks, I've found issues in frustum/perspective function, e.g. incoherences between fov and fovy usage.
Let me fix this before merging, the rest of the PR is ready for review though.

src/Core/Asset/Camera.hpp
```diff
@@ -2,31 +2,27 @@
...
+namespace Core {
+namespace Asset {
 
-namespace Scene {
-class Entity;
-/// A Camera is an Engine Component storing a Camera object.
-class RA_ENGINE_API Camera : public Scene::Component
+/// Camera class, define by a frame (assume to be orthonormal)
+/// The view direction is -z in camera space.
+/// Non orthormal frame are supported but leads to bad computation in fitZRange (bug suspected).
```
nmellado, line 12:
```suggestion
/// \brief Camera class storing the Camera frame and the projection properties
/// The view direction is -z in camera space.
/// \note Both orthonormal and non-orthonormal frames are supported. However, in the current implementation suspect a bug in fitZRange which does not behave as expected for non-orthogonal frames..
```

src/Core/Asset/Camera.hpp
```diff
@@ -141,41 +128,56 @@ class RA_ENGINE_API Camera : public Scene::Component
...
+    /// https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#cameras
+    /// equivalent to scale = tan(m_fov*.5)*m_zNear
+    /// frustum(-scale/m_aspect,+scale/m_aspect, -scale, scale, m_zNear, m_zFar);
+    /// \param a : aspect ratio as height/width, it's 1_ra/getAspect()
+    /// \param y : fov in the vertical direction, it's getFOV/getAspect()
+    /// \param n : z near
+    /// \param f : z far
+    static Core::Matrix4 perspective( Scalar a, Scalar y, Scalar n, Scalar f );
+
+    /// Return a projection matrix
```
nmellado, line 152:
```suggestion
```

src/Core/Asset/Camera.hpp
```diff
@@ -141,41 +128,56 @@ class RA_ENGINE_API Camera : public Scene::Component
...
+    /// Compute projection matrix as describe here
+    /// https://www.scratchapixel.com/lessons/3d-basic-rendering/perspective-and-orthographic-projection-matrix/opengl-perspective-projection-matrix
+    /// \param b : bottom
+    /// \param t : top
+    /// \param l : left
+    /// \param r : right
+    /// \param n : z near
+    /// \param f : z far
+    static Core::Matrix4 frustum( Scalar b, Scalar t, Scalar l, Scalar r, Scalar n, Scalar f );
+    static Core::Matrix4 ortho( Scalar b, Scalar t, Scalar l, Scalar r, Scalar n, Scalar f );
```
nmellado, line 162: Documentation is missing

src/Core/Asset/Camera.hpp
```diff
@@ -141,41 +128,56 @@ class RA_ENGINE_API Camera : public Scene::Component
...
     /// Fit the Z-range of the camera to the scene's aabb. This will maximize z-buffer precision
-    virtual void fitZRange( const Core::Aabb& aabb );
-
-  protected:
+    void fitZRange( const Core::Aabb& aabb );
+
+    const Scalar m_minZNear {0.01_ra};
+    const Scalar m_minZRange {0.01_ra};
+
+    /// Compute a project projection matrix as describe here
```
nmellado, line 142:
```suggestion
    /// \brief Build a perspective projection matrix from camera parameters (ratio, for, near and far)
    /// \see frustum()
    /// Implements the algorithm described here
```

src/Core/Asset/Camera.hpp
```diff
@@ -141,41 +128,56 @@ class RA_ENGINE_API Camera : public Scene::Component
...
+    /// equivalent to scale = tan(m_fov*.5)*m_zNear
+    /// frustum(-scale/m_aspect,+scale/m_aspect, -scale, scale, m_zNear, m_zFar);
+    /// \param a : aspect ratio as height/width, it's 1_ra/getAspect()
+    /// \param y : fov in the vertical direction, it's getFOV/getAspect()
+    /// \param n : z near
+    /// \param f : z far
+    static Core::Matrix4 perspective( Scalar a, Scalar y, Scalar n, Scalar f );
+
+    /// Return a projection matrix
+    /// Compute projection matrix as describe here
```
nmellado, line 153:
```suggestion
    /// \brief Build a projection matrix from the parameters of the view volume
    /// \see perspective()
    /// Implements the algorithm described here
```

src/Core/Asset/Camera.inl
```diff
@@ -124,34 +99,46 @@ inline void Camera::setType( const ProjType& projectionType ) {
...
+
 inline Core::Vector2 Camera::project( const Core::Vector3& p ) const {
     Core::Vector4 point = Core::Vector4::Ones();
     point.head<3>()     = p;
     auto vpPoint        = getProjMatrix() * getViewMatrix() * point;
 
-    return Core::Vector2( m_width * Scalar( 0.5 ) * ( vpPoint.x() + Scalar( 1 ) ),
-                          m_height * Scalar( 0.5 ) * ( vpPoint.y() + Scalar( -1 ) ) );
+    return Core::Vector2( getWidth() * Scalar( 0.5 ) * ( vpPoint.x() + Scalar( 1 ) ),
+                          getHeight() * Scalar( 0.5 ) * ( vpPoint.y() + Scalar( -1 ) ) );
```
nmellado, line 127:
```suggestion
    return Core::Vector2( getWidth() * 0.5_ra * ( vpPoint.x() + 1_ra ),
                          getHeight() * 0.5_ra * ( vpPoint.y() + -1_ra ) );
```

src/Core/Asset/Camera.inl
```diff
@@ -124,34 +99,46 @@ inline void Camera::setType( const ProjType& projectionType ) {
...
 }
 
 inline Core::Vector3 Camera::unProject( const Core::Vector2& pix ) const {
-    const Scalar localX = ( Scalar( 2 ) * pix.x() ) / m_width - Scalar( 1 );
+    const Scalar localX = ( Scalar( 2 ) * pix.x() ) / getWidth() - Scalar( 1 );
```
nmellado, line 131:
```suggestion
    const Scalar localX = ( 2_ra * pix.x() ) / getWidth() - 1_ra;
```

src/Core/Asset/Camera.inl
```diff
@@ -124,34 +99,46 @@ inline void Camera::setType( const ProjType& projectionType ) {
...
     // Y is "inverted" (goes downwards)
-    const Scalar localY = -( Scalar( 2 ) * pix.y() ) / m_height + Scalar( 1 );
+    const Scalar localY = -( Scalar( 2 ) * pix.y() ) / getHeight() + Scalar( 1 );
```
nmellado, line 133: Use `_ra` postfix.

src/Core/Asset/Camera.inl
```diff
@@ -124,34 +99,46 @@ inline void Camera::setType( const ProjType& projectionType ) {
...
 
     // Multiply the point in screen space by the inverted projection matrix
     // and then by the inverted view matrix ( = m_frame) to get it in world space.
     // NB : localPoint needs to be a vec4 to be multiplied by the proj matrix.
-    const Core::Vector4 localPoint( localX, localY, -m_zNear, Scalar( 1 ) );
+    const Core::Vector4 localPoint( localX, localY, -getZNear(), Scalar( 1 ) );
```
nmellado, line 138: Use `_ra` postfix.

src/Core/Geometry/AbstractGeometry.hpp
```diff
@@ -23,16 +24,44 @@ struct RA_CORE_API AbstractGeometry {
      * have to be declared as defaulted.
      * See http://en.cppreference.com/w/cpp/language/rule_of_three
      */
```
nmellado, line 26:
```suggestion
     */
```

src/Core/Geometry/TriangleMesh.hpp
```diff
@@ -164,7 +164,7 @@ class RA_CORE_API AttribArrayGeometry : public AbstractGeometry
     /// \warning The original handles are not valid for the AttribArrayGeometry copy.
     inline bool copyAllAttributes( const AttribArrayGeometry& input );
 
-    inline Aabb computeAabb() const override;
+    inline Aabb computeAabb() override;
```
nmellado, line 167: Why discarding the const operator ?
The bbox can be a `mutable` field.

src/Core/Asset/Camera.cpp
```diff
@@ -0,0 +1,193 @@
...
+Camera& Camera::operator=( const Camera& rhs ) {
+    m_frame      = rhs.getFrame();
+    m_projMatrix = rhs.getProjMatrix();
+    m_projType   = rhs.getType();
+    m_zoomFactor = rhs.getZoomFactor();
+    m_fov        = rhs.getFOV();
+    m_zNear      = rhs.getZNear();
+    m_zFar       = rhs.getZFar();
+    setViewport( rhs.getWidth(), rhs.getHeight() );
+    return *this;
```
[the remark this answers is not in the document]
dlyr: sure, we will finalize unit test before final review.

dlyr (approved):

PR #754, opened 2021-05-06
[IO] Fix assimp loading : do not load empty skeleton

MathiasPaulin:
Do not load empty skeleton in assimp loader.


dlyr:
The fix seems "correct", but we way need to refactor animation loading a bit.
I'm not an expert in assimp functionalities, maybe check some tutorial like  http://ogldev.org/www/tutorial38/tutorial38.html

nmellado (approved):

PR #764, opened 2021-05-22
Resources has data path

MathiasPaulin:

This PR brings the capabilities of defining a data path that can be used for all artefacts generated by Radium.
The Base application now have an option `-d | -data | -export` that allows user to set the default path for artefacts.
This defaults path is stored in the application settings so that it will be used for future application execution without giving the option in the command line. Giving another option will change the data path. 
All applications derived from Base Application (and only them) benefit from this.

Radium based softwares (applications, plugin, libraries, ...) should also set and use the data path through fonctions in `Ra::Core::Utils::Resources` namespace. It is of the responsibility of the application, plugin, library, ... to use or not this possibility.

Artefacts of the BaseApplication derived apps (and several others Radium-based plugins, library, ...) are written in the working directory of the application. This result in data duplication, difficulties to find the artefact, filesystem pollution, ... when executing from the build tree (different for Debug, Release, ...), from a bundled application, from an installed tree without write access ... 

This PR brings a simple way to output artifact in one location only, with the ability for the application to manage output directory in a stack-based state management way.

* **Other information**:


src/Core/Resources/Resources.cpp
```diff
@@ -55,6 +56,29 @@ optional<std::string> getResourcesPath( void* symbol, const std::string& pattern
     return p.string();
 }
 
+namespace DataPath {
+static std::stack<std::string> s_dataPaths;
+}
+
+std::string getDataPath() {
+    if ( DataPath::s_dataPaths.empty() )
+    { DataPath::s_dataPaths.emplace( fs::current_path().string() ); }
```
dlyr, line 65: Do that mean that if one call pop before get, there is a core assert, and if pop is called after get, everything is fine? 
```suggestion
    { return fs::current_path().string(); }
```

MathiasPaulin: Yes, calling get never generates an error, even if the stack is empty. In this case (first call to get), the default path is pushed and get return this default pushed path and the stack is not empty after that. Your suggestion is also fine. If the stack is empty, return the default.
We can also change the pop method so that if the stack is empty, nothing happen.

dlyr: both behavior are fine for me. I prefer not to have the current version with side effect.

MathiasPaulin: OK, I'll change that

src/Core/Resources/Resources.hpp
```diff
@@ -12,11 +12,11 @@ namespace Core {
  *    - default shaders
  *    - ...
  *
- *   @todo ensure that resources are installed or linked in the parent directory of the
- * libRadiumCore dynamic library.
- *
- *   @todo Generalize the locator such that one can find an offsetted directory from a dynamic
- *   library containing an arbitrary symbol. (easy with cpplocate)
+ *  This namespace also contains functions allowing to set and access paths to store data
+ *  generated by Radium applications, plugins or library in a portable way.
```
nmellado, line 16:
```suggestion
 *  generated by Radium applications, plugins or libraries in a portable way.
```

src/Core/Resources/Resources.hpp
```diff
@@ -12,11 +12,11 @@ namespace Core {
...
+ *  If not set, the default data path is the working directory of the application.
+ *  For Gui::BaseApplication derived applications, and for plugins, the default data path can be set
+ *  using command line parameter and is stored in the application settings.
```
nmellado, line 19: Shall we have a snippet to list in the documentation all the options of the CLI ? This snippet could be added in the documentation `Gui::BaseApplication`.

For users of the present class, it is not easy to know where to search where are the options, so we could also add a link here to the aforementioned CLI documentation.

src/Core/Resources/Resources.hpp
```diff
@@ -40,6 +40,26 @@ RA_CORE_API optional<std::string> getBasePath();
 RA_CORE_API optional<std::string> getResourcesPath( void* symbol               = nullptr,
                                                     const std::string& pattern = "Resources" );
 
+/// Data path functions.
```
nmellado, line 43:
```suggestion
/// \brief Data path functions.
```

src/Core/Resources/Resources.hpp
```diff
@@ -40,6 +40,26 @@ RA_CORE_API optional<std::string> getBasePath();
...
+/// These functions manage a stack of paths so that application, library or plugins could manage
+/// paths to store their data in a state stack-based approach.
+///@{
+/// Get the current data path
```
nmellado, line 47:
```suggestion
/// \brief the current data path
```

src/Core/Resources/Resources.hpp
```diff
@@ -40,6 +40,26 @@ RA_CORE_API optional<std::string> getBasePath();
...
+/// If an application's data path was not pushed before the call of this function, the default
+/// value, corresponding to the current working directory is returned. Otherwise, the last path
+/// pushed using a call to pushDataPath() is returned.
+RA_CORE_API std::string getDataPath();
+
+/// Push a data path that will be returned by getDataPath until it is popped.
+/// @param datapath the data path to set as current
```
nmellado, line 54:
```suggestion
/// \brief Push a data path that will be returned by getDataPath until it is popped.
/// \param datapath the data path to set as current
```

src/Core/Resources/Resources.hpp
```diff
@@ -40,6 +40,26 @@ RA_CORE_API optional<std::string> getBasePath();
...
+RA_CORE_API void pushDataPath( std::string datapath );
+
+/// Pop the current data path.
+/// The returned data path will not be available anymore for getDataPath().
+/// @return the popped data path.
```
nmellado, line 59:
```suggestion
/// \brief Pop the current data path.
/// The returned data path will not be available anymore for getDataPath().
/// \return the popped data path.
```

MathiasPaulin:
@nmellado Review comments fixed. PR Ready to review/merge.

dlyr (approved):

PR #765, opened 2021-05-22
Camera manipulator

dlyr:

Add a new camera manipulator to have "rotate around" behavior.
It differs from trackball since the target and pivot point can be different, and all rotation are made as we roll the scene like a little ball.

There is a design issue/question : RotateAround need to read screen depth, in the proposed version it is done during handleMouseEvent, which needs to make viewer's context active.
It's not an ideal situation I think, we may think the viewer's read depth value, and optionally send it to the manipulator.


src/Engine/Rendering/Renderer.cpp
```diff
@@ -714,6 +714,11 @@ int Renderer::buildAllRenderTechniques() const {
     return 0;
 }
 
+Scalar Renderer::getDepth( int, int ) {
+    CORE_ASSERT( false, "Renderer::getDepth() not implemented, must be overriden" );
```
MathiasPaulin, line 718: as the (abstract) class `Renderer` own a z-buffer (the member `m_depthTexture`) This method could be implemented using this depth buffer. But you should add a note in the documentation saying that, if a concrete renderer do not use the base depth-buffer, this method should be overridden. it is the case, e.g. for Forward Renderer that use its own depth-buffer for scene rendering.

Another solution, instead of an error message, is tu make this method pure virtual so that any renderer must implement it.

src/Engine/Rendering/Renderer.hpp
```diff
@@ -338,6 +338,9 @@ class RA_ENGINE_API Renderer
      */
     virtual std::unique_ptr<uchar[]> grabFrame( size_t& w, size_t& h ) const;
 
+    /// return the value of the depth buffer under pixel (x,y)
+    ///
```
MathiasPaulin, line 342: It should be of importance to be more clear in which space the depth value is expected : screen space, NDC, camera space ?

src/Gui/Configs/default.xml
```diff
@@ -18,6 +18,9 @@
             action="TRACKBALLCAMERA_MOVE_FORWARD"/>
     <keymap context="CameraContext" key="Key_T" modifiers="" buttons="" action="TRACKBALLCAMERA_PROJ_MODE"/>
 
+    <keymap context="CameraContext" key="Key_B" modifiers="" buttons="" action="ROTATEAROUND_ALIGN_WITH_CLOSEST_AXIS"/>
+    <keymap context="CameraContext" key="Key_V" modifiers="" buttons="" action="ROTATEAROUND_SET_PIVOT"/>
```
MathiasPaulin, line 22: Can you add also the same entry in other keymapping config (macos, alternative) ?

MathiasPaulin (commented):
This is a first round of review. Some questions are in the comments.
For the design question from the PR description, I think the viewer has to first read the depth value and give it to the manipulator. This might minimize the interactions between manipulators and renderer. For this, perhaps the renderer could rely on picking to have this value.

src/Engine/Rendering/Renderer.cpp
dlyr: Done, but reading a texture (without fbo) needs to read back the whole z buffer.
Moreover ForwardRenderer do not use m_depthTexture.
I'm not sure what is the better.

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
```diff
@@ -68,7 +68,7 @@ void AssimpCameraDataLoader::loadCameraData( const aiScene* scene,
     Core::Matrix4 rootMatrix;
     rootMatrix = Core::Matrix4::Identity();
     /// \todo tmp test    Core::Matrix4 frame  = loadCameraFrame( scene, rootMatrix, data );
```
MathiasPaulin, line 70: This todo should be removed. It was for testing purpose only and is not really informative about what to do (nor why).

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
```diff
@@ -97,9 +97,8 @@ void AssimpCameraDataLoader::loadCameraData( const aiScene* scene,
 }
 
 Core::Matrix4 AssimpCameraDataLoader::loadCameraFrame( const aiScene* scene,
-                                                       const Core::Matrix4& parentFrame,
-                                                       const aiCamera& cameraNode,
-                                                       Camera& data ) const {
+                                                       const Core::Matrix4& /*parentFrame*/,
+                                                       const aiCamera& cameraNode ) const {
     // old version
```
MathiasPaulin, line 102: If this code is no more needed, it should be deleted.

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
```diff
@@ -130,7 +129,7 @@ Core::Matrix4 AssimpCameraDataLoader::loadCameraFrame( const aiScene* scene,
     return frame;
 }
 
-void AssimpCameraDataLoader::fetchName( const aiCamera& camera, Camera& data ) const {
+void AssimpCameraDataLoader::fetchName( const aiCamera& camera, Camera& ) const {
```
MathiasPaulin, line 132: I'm not sure to understand what is the intent of this method.

MathiasPaulin (commented):
Second round of review. Perhaps we should discuss IRL.

src/Gui/Configs/alternative.xml
```diff
@@ -1,26 +0,0 @@
-<?xml version="1.0" encoding="ISO-8859-1" ?>
```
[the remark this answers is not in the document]
dlyr: After discussion afk:
- we will keep only _one_ config example, compatible with desktop, laptop and mac.
- it's meant to be adapted to each user needs, but radium lib is not the place to maintain such configuration.
- upcoming commits will update default.xml with default binding, without middle button which can be absent from desktop or mac.

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
dlyr: neither do I, it was done before this PR, here is just a "fix warning ...
todo in another PR ? I'll do a quick check if this name is meaningful and easy to propagate to component ...

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
dlyr: actually, keep the old version for the record, if someone as a issue with camera loading, it's here to investigate easilly.

src/IO/AssimpLoader/AssimpCameraDataLoader.cpp
dlyr: after a quick look, need a refactor of camera loading to feed camera manager component names.

dlyr:
There is some un init value according to valgrind, I'll check and fix quickly.

src/Engine/Rendering/Renderer.cpp
```diff
@@ -720,6 +722,11 @@ int Renderer::buildAllRenderTechniques() const {
     return 0;
 }
 
+Scalar Renderer::getDepth( int, int ) {
+    LOG( logERROR ) << "call to not implemented Renderer::getDepth (return -1) , please override";
+    return -1_ra;
```
MathiasPaulin, line 727: This does not match the documentation of the function that says return value is in [0, 1].
As this function must never be called we have four possibilities : 
 - make it pure virtual (force to implement it in concrete renderers)
 - make it return a "valid value", either 1 to rotate around the background, either 0.5 to rotate around the center of the scene, or 0 to rotate around the near plane.
 - Add a note in the documentation saying that the return value is {-1} U [0, 1] with the value -1 indicating that the service is not implemented and that the value should not be used.
 - We discussed to augment the picking result with the picked depth and to use it in Viewer instead of getDepth. I think it might be a better approach to do that but is it in the scope of this PR ?

src/Gui/Viewer/RotateAroundCameraManipulator.cpp
```diff
@@ -0,0 +1,230 @@
...
+    rotateAroundPoint( m_camera, m_target, rot, pivot );
+}
+
+Scalar RotateAroundCameraManipulator::projectOnBall( Scalar x, Scalar y ) {
+    const Scalar size       = 1.0;
+    const Scalar size2      = size * size;
+    const Scalar size_limit = size2 * 0.5;
+
+    const Scalar d = x * x + y * y;
+    return d < size_limit ? sqrt( size2 - d ) : size_limit / sqrt( d );
```
MathiasPaulin, line 178:
```suggestion
    return d < size_limit ? std::sqrt( size2 - d ) : size_limit / std::sqrt( d );
```

src/Engine/Rendering/Renderer.cpp
dlyr: fine with picking version, it solves lot's of question. 
I'll test it now.

dlyr: Some sort of resolved, but preparePicking is still in a not really finished state.

dlyr:
There is room for improvement in picking code (render only what is needed, some factorization, maybe use camera instead of rendering parameter to ease this ...)
I'm not sure it's in the scope of this PR, at least the current version do not introduce a "hacky" getDepth ...
I'm fine taking time to cleanup picking, but it can be in serveral days (or weeks).

MathiasPaulin:
> There is room for improvement in picking code (render only what is needed, some factorization, maybe use camera instead of rendering parameter to ease this ...)
> I'm not sure it's in the scope of this PR, at least the current version do not introduce a "hacky" getDepth ...

I agree, picking needs to be deeply refactored and optimized/clean for unneeded code. It is one of the topic of the discussion we had about Rendering in RadiumV2. But I also think this is out of scope for this PR.
What we can do for now is just modify picking to have also depth as you did and open an issue/feature request about picking refactoring.
BTW, the modification you did (comment out preparePicking at line 212 and replace by the code of the commented call) might lead to some inconsistencies as there are two methods that do the picking. Th one you modify (doPickingNow) and another, at line 418 : doPicking. The depth should be updated from both  the method is'nt it ?

dlyr:
Yes, depth could also be read in doPicking, but we then need to have a depth vector for each picked pixel.
In the current use case, we only need picking to set depth value for the viewer under mouse.
I'm open to any improvements.

MathiasPaulin (approved):

PR #766, opened 2021-05-23
Scalar double

### LINKED ISSUE #390: Clean Radium Options
When Compiling Radium with non-default options, there are some small issues (not detected by the CI since it runs only with default option values):
 - `RADIUM_WITH_DOUBLE_PRECISION`: in the code, there are some real constants that are not cast into the `Scalar` type, some variables are typed `float` instead of `Scalar`, and the same goes for some function/method parameters.
 - `RADIUM_WITH_OMP`: msvc does not with OpenMP 3.0, thus forbidding the use of `unsigned int`s as for-loop parameters.

I didn't test all options and some issues might still be hiding.
All this should be addressed before the v1 release.

dlyr:
This PR is made from STORM-IRIT to let the new action run correctly, need to fix workflow hook before merging, currently all options are triggered on PR before merge.

a cmake option to use double as Scalar type.
OpenGL data are kept float and on the fly conversion is added.
Also CI with different options (only when PR merged to master, current CI on PR before merge)

Fixes #390
Scalar work only with float

Scalar can be double, unittests pass, and rendering is fine.

Almost nothing for Scalar=float, on the fly conversion for GPU data if Scalar=double

* **Other information**:


dlyr:
**CHECK LINEAR HISTORY FAIL**

Please rebase on current up to date master,
without merge commits (e.g. you can squash them).

nmellado (approved):

PR #776, opened 2021-06-04
[test] Add Custom KeyEvent demo

MathiasPaulin:
All is in the title. Use `<F1>`  or `<F2>` keys to toggle colors.

### DIFF
--- a/src/Gui/Utils/KeyMappingManager.hpp
+++ b/src/Gui/Utils/KeyMappingManager.hpp
@@ -57,7 +57,7 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
     /// \brief Add a given action within a possibly non existing context (also created in this case)
     /// to the mapping system.
     /// This allow to define default behavior when some KeyMappingManageable object is not
-    /// parameterized in the application config file. The action is added to the current config file
+    /// parameterized in the application config file. The action can be added to the current config file
     /// so that it will remain for subsequent usage.
     /// \param context the context of the action
     /// \param keyString  represents the key that needs to be pressed to trigger the event
--- a/src/Gui/Utils/KeyMappingManager.hpp
+++ b/src/Gui/Utils/KeyMappingManager.hpp
@@ -59,7 +59,7 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
     /// This allow to define default behavior when some KeyMappingManageable object is not
     /// parameterized in the application config file. The action can be added to the current config
     /// file so that it will remain for subsequent usage. \param context the context of the action
-    /// \param keyString  represents the key that needs to be pressed to trigger the event
+    /// \param keyString represents the key that needs to be pressed to trigger the event
     /// (ie Key_Z, for example), "" or "-1" corresponds to no key needed.
     /// \param modifiersString represents the modifier used along with key or mouse button `
     /// (needs to be a Qt::Modifier enum value) to trigger the action. Multiples modifiers can be
--- a/src/Gui/Utils/KeyMappingManager.hpp
+++ b/src/Gui/Utils/KeyMappingManager.hpp
@@ -70,37 +70,13 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
     /// \param actionString represents the KeyMappingAction enum's value you want to
     /// trigger.
     /// \param saveToConfigFile set it to true (the default) to save the action on the config file.
-
-    void addAction( const std::string& context,
-                    const std::string& keyString,
-                    const std::string& modifiersString,
-                    const std::string& buttonsString,
-                    const std::string& wheelString,
-                    const std::string& actionString,
-                    bool saveToConfigFile = true );
-
-    /// Add a given action to an existing context in the mapping system.
-    /// This allow to define custom actions without the need of a KeyMappingManageable object.
-    /// This method do not add the action in the config file
-    /// @param context the context of the action.
-    /// @param actionName represents the KeyMappingAction enum's value you want to
-    /// trigger.
-    /// @param keyString  represents the key that needs to be pressed to trigger the event
-    /// (ie Key_Z, for example), "" or "-1" corresponds to no key needed.
-    /// @param modifiersString represents the modifier used along with key or mouse button `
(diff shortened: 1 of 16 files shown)


src/Gui/Utils/KeyMappingManager.hpp
```diff
@@ -58,23 +58,55 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
...
+    /// @param keyString  represents the key that needs to be pressed to trigger the event
+    /// (ie Key_Z, for example), "" or "-1" corresponds to no key needed.
+    /// @param modifiersString represents the modifier used along with key or mouse button `
+    /// (needs to be a Qt::Modifier enum value) to trigger the action. Multiples modifiers can be
+    /// specified, separated by commas as in "ControlModifier,ShiftModifier".
+    /// @param buttonsString represents the button to trigger the event (e.g. LeftButton).
+    /// @param wheelString if true, it's a wheel event !
+    /// @return an invalid action if context is not valid, or if actionName has  not been created.
+    // (i,e action.isInvalid())
+    KeyMappingAction addAction( const Context& context,
```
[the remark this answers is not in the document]
MathiasPaulin: The new functions allow to be more concise when creating a new (non existing) custom context with some actions and follow the principle : 
 1. Create the context and check for validity (valid if no context with the same name exists)
 2. if the context is valid, Add actions to the existing context given its id

The existing ones need to be more verbose as they create both the context and the action in the same time without returning their id and requires to do the following : 
 1. Check for existance of the context name
 2. If the context do not exist, create actions given the context name and action name (this will create the context while creating the first action)
 3. Get the context id
 4. For each action, get the action id.

In the demo, creating a context with 2 actions using the added API takes 4 lines while it takes 8 with the existing API. 
In this PR, I also add the addContext method which allows to do the concise version above. I keep the old version to be backward compatible with existing code. 
The verbose version, however, is relevant when reading configs from a xml file while the concise one are relevant when adding custom context and actions and even allows to do that dynamically.

I agree that the doc should be updated but we can also discuss on refactoring the KeyMappingManager API to allow more user friendly usage of the functionalities.

src/Gui/Viewer/CameraManipulator.hpp
```diff
@@ -69,7 +69,8 @@ class RA_GUI_API CameraManipulator : public QObject
     virtual bool handleKeyPressEvent( QKeyEvent* event,
                                       const KeyMappingManager::KeyMappingAction& action ) = 0;
```
dlyr, line 70: could we provide a default implementation "return false" so that it's not necessary to make derived class handle this ?
(it can be considered out of scope of this PR, but it would reduce code duplication).

dlyr (commented):
I have several small concerns about the code duplication, I'll do some proposal as a review (with the day)

src/Gui/Viewer/CameraManipulator.hpp
MathiasPaulin: sure, for both `handleKeyPressEvent` and `handleKeyReleaseEvent`.

MathiasPaulin:
> I have several small concerns about the code duplication, I'll do some proposal as a review (with the day)

I agree about the code duplication in this PR. BTW, integrate directly your proposal in a commit rather that in review comments.

src/Gui/Utils/KeyMappingManager.hpp
```diff
@@ -54,27 +54,36 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
                                                    int key,
                                                    bool wheel = false );
 
-    /// Add a given action to the mapping system.
+    /// \brief Add a given action within a possibly non existing context (also created in this case)
+    /// to the mapping system.
     /// This allow to define default behavior when some KeyMappingManageable object is not
     /// parameterized in the application config file. The action is added to the current config file
```
nmellado, line 60:
```suggestion
    /// parameterized in the application config file. The action can be added to the current config file
```

src/Gui/Utils/KeyMappingManager.hpp
```diff
@@ -54,27 +54,36 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
...
     /// so that it will remain for subsequent usage.
-    /// @todo write the configuration in the configFile to be later reused or modified ?
-    /// @param context the context of the action
-    /// @param keyString  represents the key that needs to be pressed to trigger the event
+    /// \param context the context of the action
+    /// \param keyString  represents the key that needs to be pressed to trigger the event
```
nmellado, line 63:
```suggestion
    /// \param keyString represents the key that needs to be pressed to trigger the event
```

src/Gui/Utils/KeyMappingManager.hpp
```diff
@@ -54,27 +54,36 @@ class RA_GUI_API KeyMappingManager : public Ra::Core::Utils::ObservableVoid
...
+    /// \param wheelString if true, it's a wheel event !
+    /// \param actionString represents the KeyMappingAction enum's value you want to
     /// trigger.
-    void addAction( const std::string& context,
-                    const std::string& keyString,
-                    const std::string& modifiersString,
-                    const std::string& buttonsString,
-                    const std::string& wheelString,
-                    const std::string& actionString );
+    /// \param saveToConfigFile set it to true (the default) to save the action on the config file.
```
nmellado, line 72:
```suggestion
    /// \param saveToConfigFile request to save the action on the config file (true by default).
```

src/Gui/Viewer/CameraManipulator.cpp
```diff
@@ -71,6 +71,42 @@ void CameraManipulator::attachLight( Engine::Scene::Light* light ) {
 KeyMappingManager::Context Gui::CameraManipulator::mappingContext() {
     return Gui::KeyMappingManager::Context();
 }
+/// @return true if the event has been taken into account, false otherwise
```
nmellado, line 74: Why do we repeat doxygen comments is the cpp file ? (same for other functions)

nmellado (commented):
Looks nice !

src/Gui/Viewer/CameraManipulator.cpp
dlyr: because of quick copy paste.

dlyr (approved):

PR #782, opened 2021-06-15
[gui] Mark event VIEWER_TOGGLE_WIREFRAME caught to generate redraw event

### LINKED ISSUE #781: Missing redraw event when toggling wireframe mode on Viewer
In any Viewer based application (including the small demos), when toggling wireframe mode using the corresponding key-mapping event, the event is not marked as caught and no redraw event is generated.
This make the user to move the mouse over the Viewer (or force redraw) to have wireframe activated/deactivated.

MathiasPaulin:
closes #781


MathiasPaulin:
Wrong destination branch

PR #784, opened 2021-06-16
[core] detach observers on MultiIndexedGeometry before deletion.

### LINKED ISSUE #783: Application crash after toggling wireframe
In example applications (HelloRadium, DrawPrimitiveDemos, ...) as well as in Sandbox from Radium-Apps,
when the user has switched to wireframe in its interactive session, the application crashed at exit.
The application also crash, each time wireframe mode was used,  when clearing the scene or selecting an object after removing another one.
To reproduce : 
 - run the HelloRadium exemple app 
 - in the 5s interval, toggle wireframe using the corresponding key-mapping (`<Ctrl-Z>` on Linux or windows, `<Cmd-W>` on macos)
 - At exit, application crashes and, in debug mode, print the message : 
 > .....Radium-Engine/src/Core/Geometry/IndexedGeometry.cpp:182: Assertion `!p.first` failed : try to get already locked layer

MathiasPaulin:
Closes #783

### DIFF
--- a/src/Core/Geometry/IndexedGeometry.inl
+++ b/src/Core/Geometry/IndexedGeometry.inl
@@ -80,6 +80,7 @@ inline GeometryIndexLayerBase* GeometryIndexLayer<T>::clone() {
 
 // MultiIndexedGeometry
 inline MultiIndexedGeometry::~MultiIndexedGeometry() {
+    detachAll();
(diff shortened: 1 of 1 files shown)


MathiasPaulin:
Wrong destination branch

PR #786, opened 2021-06-16
[core] detach observers on MultiIndexedGeometry before deletion.

### LINKED ISSUE #783: Application crash after toggling wireframe

MathiasPaulin:
Closes #783

### DIFF
--- a/src/Core/Geometry/IndexedGeometry.inl
+++ b/src/Core/Geometry/IndexedGeometry.inl
@@ -80,6 +80,7 @@ inline GeometryIndexLayerBase* GeometryIndexLayer<T>::clone() {
 
 // MultiIndexedGeometry
 inline MultiIndexedGeometry::~MultiIndexedGeometry() {
+    detachAll();
     clear();
 }
 


nmellado:
Good point. 
But, shouldn’t we fix that in the destructor of Observable directly ? This can happen with other types of objects.

MathiasPaulin:
> Good point.
> But, shouldn’t we fix that in the destructor of Observable directly ? This can happen with other types of objects.

Not really. If we do that, as the destructor of `Observable` will be called after the destructor of `MultiIndexedGeometry` (inheritance property), the observers will still be notified and the crash will persist. Same for others types deriving from observable.

nmellado:
Hum true...
Could we add something like `delete_later` in `Observable` ? I am afraid that this bug appears again and again...

MathiasPaulin:
> Could we add something like `delete_later` in `Observable` ? I am afraid that this bug appears again and again...

I think that this problem comes from 2 concerns about the use of observer/observable pattern. For me : 
 1. It is the responsibility of the observer or of the one who attached an observer to an observable to ensure that this connection has a lifetime shorter than the one of the observable. The lifetime of an object being defined from the last instruction of the constructor to the start of the destructor (https://en.cppreference.com/w/cpp/language/lifetime) and for a observer/observable connection from the attach to the detach.
 2. Due to the previous, calling notify at the end of the destructor in not really a good idea. 

In my opinion, observers must be detached from an observable before the destructor of the observable start. 
The proposed fix force this detach by calling detachAll as the first destructor instruction to ensure this is done but is only a workaround for lifetime management problem on objects and connections between objects.
And I recall that inside the Engine, ComponentMessenger allows to do that easily for Entity/Component/RenderObjects. Outside of Engine, it is the responsibility of the programmer to ensure this.

dlyr:
Maybe it's because the dtor should not call clear, but only    
AttribArrayGeometry::clear();
    deepClear();

MathiasPaulin:
> Maybe it's because the dtor should not call clear, but only
> AttribArrayGeometry::clear();
> deepClear();

Sure, this matches the point 2. of my previous answer to not call notify in a destructor.

nmellado:
Ok so what do we do ? Do we keep the present fix, or the one proposed by @dlyr  ?
I don't mind.

MathiasPaulin:
Whatever you want. But the same concept than what I proposed (detach the observer(s)  before deleting) is used in `RenderObject::~RenderObject()` that detach the aabb observer as the first operation of the destructor. Thats why there is no crash into the call to `invalidateAabb` in the `clear` method.

dlyr:
The case of RenderObject is slightly different, the observer detach himself before deletion.
We could (in an upcomming PR) add a notification to every observer that they are detached ... but the current lightweight design do not need this I think.

On issue arise from the fact we provide a ref accessor to the core geometry, and the observer store this ref.
I think a better solution is to switch to weak ptr so that observer can check if its observable is still valid.

So my point of view is:
- observer detach themselves on dtor
- observable dtor should call notify, but observer must check observable validity (hence storing a weak ptr)

nmellado (approved):

MathiasPaulin:
> * observer detach themselves on dtor

I agree, this is done for several observers (e.g. in `RenderObject` and when using `SignalManager`). But for the use case that generates the bug, due to the way LineMeshes are handled for now, its quite hard to implement as the observer is not stored outside of the observable. Even if this will change in a near future by the full use of layers (no more "siamese" meshes but only several layers on one mesh), the same problem can arise for other use cases. And how to do that if the observer is a lambda ?
So, my proposal is to have either observer to detach themselves or "owner" of the attachment (the one who attach the oserver) to ensure lifetime of the attachement is valid.

> * observable dtor should call notify, but observer must check observable validity (hence storing a weak ptr)

We had a discussion about this with @nmellado and it seems to us that we can't enforce observers to test for observable validity. And How can we do that if observer is a lambda (we had several such cases, with a good exemple on how to handle this in Timeline, using Signal manager.

dlyr:
I'm fine with this PR.

Regarding observable validity, it's only if the observer store a ref/ptr to the observable. In the case of a lambda, their is no such things (it will only have the args value from notify).

Let's move to a discussion for the upcoming comments.

dlyr (approved):

PR #791, opened 2021-07-02
Volume loader and demo

MathiasPaulin:
This PR adds a volume loaders, handling pvm volume format from The Volume Library (http://schorsch.efi.fh-nuernberg.de/data/volume/) and add a demo about volume rendering.


nmellado:
It seems that the volume files are not copied in the build directory, and so I cannot run the tests because the files are not found.

nmellado:
@MathiasPaulin FYI you have compiling errors on this PR,

nmellado:
Nice proposal. 
Would it make sense to have the VolumeDemo loading one of the volumetric file ? It would be more fancy.

nmellado (approved):
Ready to merge on my side.
@dlyr do you want to have a look ?

src/Core/Geometry/Volume.hpp
```diff
@@ -186,6 +186,10 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
 
     using Container = std::vector<ValueType>;
 
+    // For now, assume that AbstractDiscreteVolume::ValueType is Scalar
+    // vector4 : gradient and density
+    using Gradientcontainer = std::vector<Vector4>;
+
```
dlyr, line 192: Not sure to understand the comment (i.e. for now)
Should it state that ValueType is expected to be scalar, so the gradient is Ra::Vector<ValueType> or something like that ?

tests/ExampleApps/VolumeDemoApp/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+#include <QTimer>
+
+int main( int argc, char* argv[] ) {
+    //! [Creating the application]
+    Ra::Gui::BaseApplication app( argc, argv );
+    app.initialize( Ra::Gui::SimpleWindowFactory {} );
+    //! [Creating the application]
+
+#ifndef DEMO_FILE
```
dlyr, line 19: Is this define planned to be kept as this ?

tests/ExampleApps/VolumeDemoApp/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+        for ( int j = 0; j < sy; ++j ) {
+            for ( int k = 0; k < sz; ++k ) {
+                data[i + sx * ( j + sy * k )] = densityField( i, j, k, 50 );
+            }
+        }
+    }
+    auto volume    = new Ra::Core::Asset::VolumeData( "DemoVolume" );
+    volume->volume = density;
+
+    // from measured media at https://github.com/mmp/pbrt-v3/blob/master/src/core/medium.cpp
```
dlyr, line 49: Should be either removed or added as option (with keyboard key to switch material for instance)

src/Core/Geometry/Volume.hpp
MathiasPaulin: For  now (the current implementation), the gradient storing and computation is only compatible with ValueType=Scalar. This is similar to the general note about volume stating that (file Volume.hpp, line 21): 
```
\note Right now the encoded function \f$f\f$ is dimension 1, ie. \f$f:\Re^3->\Re\f$. If needed,
 *          the class can be templated so the built-in typedef #ValueType is replaced by a
 * parameter.
```
This comment is just here to recall that if we change `AbstractDiscreteVolume::ValueType`, then we need to integrate this change here. May be reworded into
```suggestion
    // Right now, assume that AbstractDiscreteVolume::ValueType is Scalar
    // In this implementation, the type Vector4 store the gradient (x, y, z) and density (w)
    using Gradientcontainer = std::vector<Vector4>;

```

tests/ExampleApps/VolumeDemoApp/main.cpp
MathiasPaulin: Yes. The symbol DEMO_FILE is defined in the CMakeLists.txt so that we are sure the path to the loaded file is correctly set. If this symbol is not define, the demo default to a simpler, synthetic scene.

tests/ExampleApps/VolumeDemoApp/main.cpp
MathiasPaulin: If the symbol DEMO_FILE is not defined, the synthetic spherical volume is used. So, there is an option at build time to choose between loading a volume file or building a synthetic volume.

MathiasPaulin: > (with keyboard key to switch material for instance)
You're welcome to commit and push any suggestion.

src/Core/Geometry/Volume.hpp
```diff
@@ -186,6 +186,10 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
 
     using Container = std::vector<ValueType>;
 
+    // For now, assume that AbstractDiscreteVolume::ValueType is Scalar
+    // vector4 : gradient and density
```
dlyr, line 190:
```suggestion
    // Since AbstractDiscreteVolume::ValueType is scalar (see class documentation),
    // hence gradient is a Vector3<ValueType>. 
    // GradientContainer also store density as forth component
```

src/Core/Geometry/Volume.hpp
```diff
@@ -186,6 +186,10 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
...
+    using Gradientcontainer = std::vector<Vector4>;
```
dlyr, line 191:
```suggestion
    using GradientContainer = std::vector<Eigen::Matrix<AbstractDiscreteVolume::ValueType, 4, 1>;>;
```

src/Core/Geometry/Volume.hpp
```diff
@@ -208,6 +212,17 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
...
     }
 
+    /// Test if gradients are defined
+    bool hasGradients() const { return m_data.size() == m_gradient.size(); }
+
+    /// Generate gradients from data
+    void computeGradients();
+
+    /// Direct access to the managed gradients
+    inline const Gradientcontainer& gradient() const { return m_gradient; }
```
dlyr, line 222:
```suggestion
    inline const GradientContainer& gradient() const { return m_gradient; }
```

src/Core/Geometry/Volume.hpp
```diff
@@ -208,6 +212,17 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
...
+    /// Direct access, with modification allowed to the managed gradients
+    inline Gradientcontainer& gradient() { return m_gradient; }
```
dlyr, line 224:
```suggestion
    inline GradientContainer& gradient() { return m_gradient; }
```

src/Core/Geometry/Volume.hpp
```diff
@@ -227,8 +242,12 @@ class RA_CORE_API VolumeGrid : public AbstractDiscreteVolume
     }
 
   private:
+    /// sample the volume at (i, j, k) with "clamp to border" behavior
+    ValueType sample( const IndexType& i );
+
     ValueType m_defaultValue;
     Container m_data;
+    Gradientcontainer m_gradient;
```
dlyr, line 250:
```suggestion
    GradientContainer m_gradient;
```

tests/ExampleApps/VolumeDemoApp/main.cpp
dlyr, line 19:
```suggestion
// set during configure time by cmake
#ifndef DEMO_FILE
```

tests/ExampleApps/VolumeDemoApp/main.cpp
dlyr, line 49:
```suggestion

```

tests/ExampleApps/VolumeDemoApp/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+    volume->sigma_s = Ra::Core::Utils::Color{ 1.7982e-05_ra, 1.3758e-05_ra, 1.2023e-05_ra };
+    volume->sigma_a = Ra::Core::Utils::Color{ 0.010782_ra, 0.011855_ra, 0.023997_ra };
+    */
+    /*
+        // As the scene unit is meter, multiply by 1000
+        volume->sigma_s *= 1000;
+        volume->sigma_a *= 1000;
+    */
+
+    // keep default absorption/diffusion  coeffficients : default to Air
```
dlyr, line 77:
```suggestion
```

tests/ExampleApps/VolumeDemoApp/main.cpp
nmellado: This is here because I asked if we could have the demo loading a file instead of displaying a sphere, it is more fancy.
But on the other hand we do not have a proper resource management for demos, so here comes the hack.

dlyr:
I'm fine for accepting this PR, @MathiasPaulin if it's ok for you could you rebase and remove WIP tag ?
I could also do this, but I prefer to be sure there is no pending WIP before merge.

MathiasPaulin:
> I'm fine for accepting this PR, @MathiasPaulin if it's ok for you could you rebase and remove WIP tag ?
> I could also do this, but I prefer to be sure there is no pending WIP before merge.

Rebase done, WIP tag removed, this PR is ready for merge.

PR #794, opened 2021-07-10
Check compiler version in external/cmake

nmellado:
Bugfix (kind of)

It is currently possible to compile externals with an unsupported compiler.

Check compiler version in 'external/cmake'

No change required downstream

* **Other information**:


MathiasPaulin (approved):
I'm not sure that defining more constraints than the ones expressed by the externals libs themselves is mandatory but this cost only a few and can help the dizzy.

dlyr (approved):

PR #803, opened 2021-07-19
[io] Remove deprecated attribute color in GeometryData

nmellado:

Remove deprecated code

Loaders use deprecated attribute m_color in `Asset::GeometryData`.

Remove this deprecated attribute and update Tinyply loader to set attributes directly

Other loaders using this attribute will have to be updated to use the attribmanager.

* **Other information**:
Other deprecated attributes will have to be cleaned the same way, but this can be done in another PR.

### DIFF
--- a/src/Core/Asset/GeometryData.cpp
+++ b/src/Core/Asset/GeometryData.cpp
@@ -18,7 +18,6 @@ GeometryData::GeometryData( const std::string& name, const GeometryType& type ) 
     m_tangent(),
     m_bitangent(),
     m_texCoord(),
-    m_color(),
     m_material() {}
 
 GeometryData::~GeometryData() {}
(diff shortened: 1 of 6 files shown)


dlyr (approved):

dlyr:
It could be a great improvement to add an integration tests that checks the colors are loaded correclty.
This test should be tested before this PR changes, and after, to check there isn't any regression.
But even if it's added only after, it will be a valuable improvement.

nmellado:
You mean the one you already wrote ?
https://github.com/STORM-IRIT/Radium-Engine/blob/master/tests/integration/TinyPlyFileLoader/main.cpp#L21

dlyr:
Not exactly, this tests only check loading stdout output, not what is actually stored in GeometryData nor GeometryComponent after loading.
Maybe what we need is more a unit tests that checks loaded data.

nmellado:
I added an issue for better integration tests, if it is fine for you I guess we can merge this PR to `release-candidate` ?
@MathiasPaulin could you please have a look at some point ? It might affect your loaders.

MathiasPaulin (approved):

Issue #808, opened 2021-07-27
Seg Fault when loading a camera with DefaultCameraManager


MatheoDumont:
I tried to use the `DefaultCameraManager, so i instanced it in a derived class of BaseApplication :
```cpp
 void MainApplication::engineBaseInitialization()
    {
        ...
        m_engine->registerSystem(SYS_CAMERA_NAME, new Ra::Engine::Scene::DefaultCameraManager);
        ...
    }

```

Then created a CameraComponent and attached it to an entity and the system :
```cpp
...

Ra::Engine::Scene::DefaultCameraManager *cam_sys = 
                (Ra::Engine::Scene::DefaultCameraManager *)engine->getSystem(SYS_CAMERA_NAME);
CameraComponent *camera = new CameraComponent(entity, COMP_CAMERA_NAME);
camera->getCamera()->setPosition(Vec3(0, 0, 1));
cam_sys->addComponent(entity, camera);
cam_sys->activate(cam_sys->getCameraIndex(camera)); 

...
```

But i get a SegFault, am i doing something wrong ?

MathiasPaulin:
is the segfault on the line `camera->getCamera()->setPosition(Vec3(0, 0, 1));` ?
what is the value of the pointer returned by `camera->getCamera()` ?
Which revision of Radium do you use (branch name and last commit date except yours) ? 
There have been some camera management modifications in the past few weeks, this precision might help us to find and explain why you have a segfault.

MatheoDumont:
When `camera->getCamera()` is printed out, i get an address.
Last Commit :
commit cfcf1fabfeb35790c03291a370a0c7dcfc6d2e03 (origin/master, origin/HEAD)
Merge: 43c185efa 62e5b3604
Author: Nicolas Mellado <nmellado0@gmail.com>
Date:   Wed May 26 15:40:17 2021 +0200

And here's the stack trace of Valgrind:
```
==996== Invalid read of size 1
==996==    at 0x52726E0: Ra::Engine::Rendering::RenderObject::isVisible() const (in /home/simulateurrsm/Library/Radium-Engine-Release/Bundle-GNU/lib/libEngine.so.1.0.0)
[...]

PR #811, opened 2021-07-30
Camera frame issues solved

### LINKED ISSUE #810: Camera issues
There are several issues with camera frame : 
  - Assimp loader generate wrong camera frames compared to references
  - The camera frame, when used for rendering, does not take into account the transformation of the entity where the camera is attached. This is not coherent with the way frames are computed for geometric object. If both a mesh and a camera is attached to an animated entity, only the mesh and the camera representation (a mesh) follow the animation. The camera frame does not.
 
Due to these, it is very hard to animate a camera from an animation of its entity or to have the same view of a scene when loading scenes with assimp loader.

MathiasPaulin:
Closes #810 and #808

Bug fix

See issue #810 : camera frame are wrong when loaded by assimp loader and camera frame does not follow the entity transformation

Camera frame are correct when loaded using assimp loader
The active camera frame, managed the CameraManager takes into account the entity transformation : 
 - when requesting a camera from the camera manager, the frame of the camera is computed to take into account the transformation of the entity
 - if an entity is animated, the frame of the active camera is automatically updated so that the active camera is also animated.
 - The viewer now observe the Camera manager so that when the active camera is modified (either by switching the camera or due to the entity animation), it could update its CameraManipulator internal state.

No.

* **Other information**:

This PR also add a demonstration on how to animate an entity. In this demonstration, a system animate an entity and any component attached to this entity (geometry, Camera)  is then animated also.
The demo show the benefit of this PR as only the entity must be animated. 
 - Geometry components follow the transformation of the entity (this was already the case in the previous Radium version)
 - Camera component follow the animation also and when the camera is active, the viewer render the scene from the animated camera (this was not the case before this PR)
 -

### DIFF
--- a/src/Engine/Scene/Entity.hpp
+++ b/src/Engine/Scene/Entity.hpp
@@ -44,7 +44,7 @@ class RA_ENGINE_API Entity : public Core::Utils::IndexedObject
     void swapTransformBuffers();
 
     /// get a ref to transformation observers to add/remove an observer
-    inline Core::Utils::Observable<const Entity*>& transformationObservers();
+    inline Core::Utils::Observable<const Entity*>& transformationObservers() const;
 
     // Components
     /// Add a component to the given entity. Component ownership is transfered to the entity.
@@ -82,7 +82,8 @@ class RA_ENGINE_API Entity : public Core::Utils::IndexedObject
     Core::Aabb m_aabb;
 
     /// Listeners on Entity Transformation changes
-    Core::Utils::Observable<const Entity*> m_transformationObservers;
+    mutable Core::Utils::Observable<const Entity*> m_transformationObservers;
+
     ///\todo add an index;
 };
 
--- a/src/Engine/Scene/Entity.hpp
+++ b/src/Engine/Scene/Entity.hpp
@@ -9,6 +9,7 @@
 
 #include <Core/Types.hpp>
 #include <Core/Utils/IndexedObject.hpp>
+#include <Core/Utils/Observable.hpp>
 
 namespace Ra {
 namespace Engine {
@@ -41,6 +42,9 @@ class RA_ENGINE_API Entity : public Core::Utils::IndexedObject
     const Core::Matrix4& getTransformAsMatrix() const;
 
     void swapTransformBuffers();
+
+    /// get a ref to transformation observers to add/remove an observer
+    inline Core::Utils::Observable<const Entity*>& transformationObservers();
 
     // Components
     /// Add a component to the given entity. Component ownership is transfered to the entity.
@@ -77,6 +81,8 @@ class RA_ENGINE_API Entity : public Core::Utils::IndexedObject
     bool m_isAabbValid {false};
     Core::Aabb m_aabb;
 
+    /// Listeners on Entity Transformation changes
+    Core::Utils::Observable<const Entity*> m_transformationObservers;
     ///\todo add an index;
 };
 
--- a/tests/ExampleApps/EntityAnimationDemo/main.cpp
+++ b/tests/ExampleApps/EntityAnimationDemo/main.cpp
@@ -14,12 +14,10 @@
 // include the task animation interface
 #include <Core/Tasks/Task.hpp>
 
-// include the Camera and default camera manager interface
-#include <Engine/Scene/CameraComponent.hpp>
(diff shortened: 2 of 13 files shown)


src/Engine/Scene/Entity.hpp
```diff
@@ -77,6 +81,8 @@ class RA_ENGINE_API Entity : public Core::Utils::IndexedObject
     bool m_isAabbValid {false};
     Core::Aabb m_aabb;
 
+    /// Listeners on Entity Transformation changes
+    Core::Utils::Observable<const Entity*> m_transformationObservers;
```
nmellado, line 85: Another option would be to mark `m_transformationObservers` as `mutable`, so the method `transformationObservers` could be kept as `const`. This would simplify user code, don't you think ?

tests/ExampleApps/EntityAnimationDemo/main.cpp
```diff
@@ -0,0 +1,231 @@
...
+    explicit DemoWindow( uint w = 800, uint h = 640, QWidget* parent = nullptr ) :
+        Ra::Gui::SimpleWindow::SimpleWindow( w, h, parent ) {}
+
+    //! [Configure Custom KeyEvent for the window]
+    void configure() override {
+        // <alt>+F will switch to default camera
+        // <alt>+C will toggle on cameras
+        // these customs events are only here for demonstration purposse.
+        // events expressed for an apple french keyboard
+        getViewer()->addKeyPressEventAction(
```
nmellado, line 55: These events look nice. Shall we add them to `SimpleWindows` as default ?

src/Engine/Scene/Entity.hpp
MathiasPaulin: I though about that, yes. It could be simpler. I'll test it.

tests/ExampleApps/EntityAnimationDemo/main.cpp
MathiasPaulin: The viewer, and so the `SimpleWindow`, already has an event to switch cameras in turn : `<ALT>+c` Key event. 
So, there is no need for these that are redundant with viewer switching.
If an application want a direct access to the camera, it is easy to add custom event, even on SimpleWindow, without defining a Custom Window as in this demo ... I'll simplify more the demo to do that.

tests/ExampleApps/EntityAnimationDemo/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+        //! [Creating the cube]
+        auto cube = Geometry::makeSharpBox( {0.5f, 0.5f, 0.5f}, Utils::Color::Green() );
+        //! [Creating the Cube]
+
+        //! [Create a geometry component with the cube]
+        auto c = new Scene::TriangleMeshComponent(
+            "Fixed cube geometry", e, std::move( cube ), nullptr );
+        //! [Create a geometry component with the cube]
+    }
+    //! [Create the demo fixed component]
```
nmellado, line 115:
```suggestion
```

tests/ExampleApps/EntityAnimationDemo/main.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    //! [Cache the camera manager]
+    auto cameraManager = static_cast<Scene::DefaultCameraManager*>(
+        app.m_engine->getSystem( "DefaultCameraManager" ) );
+    //! [Cache the camera manager]
+
+    //! [Add usefull custom key events]
+    app.m_mainWindow->getViewer()->addKeyPressEventAction(
+        "switchCam0", "Key_0", "ShiftModifier", "", "false", [cameraManager]( QKeyEvent* event ) {
+            cameraManager->activate( event->key() - '0' );
```
nmellado:
```suggestion
            // Convert ascii code to camera index: https://stackoverflow.com/a/30727561
            cameraManager->activate( event->key() - '0' );
```

tests/ExampleApps/EntityAnimationDemo/main.cpp
MathiasPaulin: this is the end of a nested snippet ...

nmellado (approved):

nmellado:
Ready to merge when up to date with `release_candidate`.

PR #839, opened 2021-12-16
Fix #838

### LINKED ISSUE #838: Render pass properties (#define) depending on the pass parameter provider are not set for pass other than default
When building render technique for different passes, one can add to the shader configuration per-pass properties using `RenderTechnique::addPassProperties`.
The parameter providers could also add specific properties to the pass they are attached (in the method `RenderTechnique::setParametersProvider`). But this is not correct in the actual implementation as Providers only set properties to the default pass.
This prevents to build complex render graph were each pass as its own parameter provider that needs to add properties to the shader configuration.

MathiasPaulin:
Fix issue #838 

This PR solve the bug described in the issue #838.
It also add a better tangent computation for textured objects in the default implementation of VertexAttribInterface.frag.glsl which improve tangent space normal mapping robustness.

see description in the issue 

pass properties depending on parameter provider are set for the correct pass.

No

* **Other information**:

### DIFF
--- a/Shaders/Lights/DefaultLight.glsl
+++ b/Shaders/Lights/DefaultLight.glsl
@@ -48,7 +48,7 @@ vec3 spotLightDirection( Light light, vec3 position ) {
 }
 
 float directionalLightAttenuation( Light light ) {
-    return 1.0f;
+    return 1.f;
 }
 
 float pointLightAttenuation( Light light, vec3 position ) {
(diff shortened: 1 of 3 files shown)


nmellado (approved):
Sounds good to me, but I didn't try to compile and see the impact on rendering.
Btw I changed the base branch from `master` to `release-candidate`.

nmellado:
@MathiasPaulin the CI didn't start because you pushed the PR against the `master` branch. I have fixed it afterward, but apparently the CI didn't start and I cannot launch it.
Can you try to push something minor to trigger the CI ?

MathiasPaulin:
> Sounds good to me, but I didn't try to compile and see the impact on rendering.

You will see nothing if you only use the default forward renderer in any Radium application (either the demos or Sandbox). To see something, you need a more complex graph with custom passes and different providers each adding their own defines per pass. You can see also the bug using the forward renderer when using our gltf plugin with advanced materials effects (e.g. clear_coat effect on a gltf material instance while others materials are not coated)... 

> Btw I changed the base branch from `master` to `release-candidate`.

oups ... thanks :~

PR #848, opened 2022-01-13
[Gui] Propage Mouse events using Qt signals

nmellado:

Feature

Mouse events are handled by the viewer and its inheriting classes

Mouse events are propagated through QT signals, which allows to use them in plugins

No

* **Other information**:

### DIFF
--- a/src/Gui/Viewer/Viewer.hpp
+++ b/src/Gui/Viewer/Viewer.hpp
@@ -181,6 +181,13 @@ class RA_GUI_API Viewer : public WindowQt, public KeyMappingManageable<Viewer>
 
     void needUpdate();
 
+    /// Event sent after a mouse press event has been processed, but before emitting needUpdate()
+    void onMousePress( QMouseEvent* event );
+    /// Event sent after a mouse release event has been processed, but before emitting needUpdate()
+    void onMouseRelease( QMouseEvent* event );
+    /// Event sent after a mouse move event has been processed, but before emitting needUpdate()
+    void onMouseMove( QMouseEvent* event );
+
   public slots:
     /// Tell the renderer to reload all shaders.
     void reloadShaders();
--- a/src/Gui/Viewer/Viewer.cpp
+++ b/src/Gui/Viewer/Viewer.cpp
@@ -541,11 +541,13 @@ void Viewer::mousePressEvent( QMouseEvent* event ) {
     m_depthUnderMouse = result.getDepth();
 
     handleMousePressEvent( event, result );
+    emit onMousePress( event );
     emit needUpdate();
 }
 
 void Viewer::mouseReleaseEvent( QMouseEvent* event ) {
     handleMouseReleaseEvent( event );
+    emit onMouseRelease( event );
     emit needUpdate();
 }
 
@@ -561,6 +563,7 @@ void Viewer::mouseMoveEvent( QMouseEvent* event ) {
     m_depthUnderMouse = result.getDepth();
 
     handleMouseMoveEvent( event, result );
+    emit onMouseMove( event );
     emit needUpdate();
 }
 


nmellado:
This trick is already used in some plugins.
It is true it does not fit the keymapping system,  but that is a quick fix that enable many interesting functionalities, to control user interaction from Radium plugins.

dlyr:
I do not get the point why do we need to different kind of behavior, one with callback for keyboard, one with signal for mouse ?
Both are fine for me, but maybe having something coherent is better ?

nmellado:
I do agree.
But this allows to use existing code with low effort and now, while the other option will take more time to be developed (and it should be).

dlyr (approved):

MathiasPaulin (commented):
Simple, working but questionable :
  - how to prevent the needUpdate (or indirecly frameUpdate) to be called twice (or even  once) if the listener is implemented such that it asks for update on some cases but do not want  to update the frame in another situation. 
   - how to develop an application/plugin/listener object using this functionality? Please update the demo or propose a simple one. 
    - this method being  Qt specific, it should require to develop more  code than needed to have portable component or to share the mouse behavior with keyboard behavior. 
    
Based on this question, I propose you tag the signals as « deprecated  » and the PR as « temporary fix(hack) » so  that developpers will not use this « hack  » too much in their apps 

As I understand the functionallity as being important and quite urgent, i do not argue against this PR but it should definitely be replaced as soon as possible by a more robust/general approach based on Qt independant callbacks. (Similar to keyboard évent propagation)

MathiasPaulin:
Simple, working but questionable :

how to prevent the needUpdate (or indirectly frameUpdate) to be called twice (or even once) if the listener is implemented such that it asks for update on some cases but do not want to update the frame in another situation.
how to develop an application/plugin/listener object using this functionality? Please update the demo or propose a simple one.
- this method being Qt specific, it should require to develop more code than needed to have portable component or to share the mouse behavior with keyboard behavior.
Based on this question, I propose you tag the signals as « deprecated  » and the PR as « temporary fix(hack) » so that developers will not use this « hack  » too much in their apps

As I understand the functionality as being important and quite urgent, i do not argue against this PR but it should definitely be replaced as soon as possible by a more robust/general approach based on Qt independent callbacks. (Similar to keyboard évent propagation)

nmellado:
> how to prevent the needUpdate (or indirectly frameUpdate) to be called twice (or even once) if the listener is implemented such that it asks for update on some cases but do not want to update the frame in another situation.

As written is the signal documentation, `needUpdate` will be called after the signal is emitted anyway. So there is no need to call `needUpdate` explicitly in the listener.

> how to develop an application/plugin/listener object using this functionality? Please update the demo or propose a simple one.
> this method being Qt specific, it should require to develop more code than needed to have portable component or to share the mouse behavior with keyboard behavior.

This thing is useful only for plugins: currently, anyone writing his own application can override the `handleMouseXXXEvent` in his custom class inhering `Ra::Gui::Viewer`. So yes, it is Qt specific, but the plugin system is currently relying on Qt. Also, it is a Qt signal which can be connected to any Qt slot: I can write it in the documentation, but it is very very standard and I don't see the point. Writing a demo app is not working either, as it makes no sense for apps.

@dlyr has open an issue to remember that we need a better implementation (#850).

So, what I can do: 
 - mark the signals as deprecated
 - add _ temporary fix(hack)_ in the commit message.

Does that work for you both ?

dlyr:
I'm not for deprecated since there is no current alternative.

MathiasPaulin:
> This thing is useful only for plugins: currently, anyone writing his own application can override the `handleMouseXXXEvent` in his custom class inhering `Ra::Gui::Viewer`. So yes, it is Qt specific, but the plugin system is currently relying on Qt. Also, it is a Qt signal which can be connected to any Qt slot: I can write it in the documentation, but it is very very standard and I don't see the point. Writing a demo app is not working either, as it makes no sense for apps.

One can develop a specific app without inheriting from viewer or something else. All DemoApps do this. Just use the BaseApplication and its Viewer, define a specific window factory (or use the provided ones) and implement application specific logic without worrying about Qt, Viewer ... take a look on, e.g. KeyEventDemoApp the defines its own Qt window to manage events and implement actions on the event (not separated from Qt-related stuff but this can be done quite easily and is a recommandation of all CHI software development workflow). 
One can also separate Qt-related event management (for mouse or keyboard actions) from application/library/plugin logic using observer/observable pattern. 

But this is out of topic of this PR even if a simple demo of the usage might be helpfull.

> So, what I can do:
>
> - mark the signals as deprecated
> - add _ temporary fix(hack)_ in the commit message.
>
> Does that work for you both ?

Fine for me.

nmellado:
Tell me what to do with the deprecated flag.

src/Gui/Viewer/Viewer.hpp
```diff
@@ -181,6 +181,13 @@ class RA_GUI_API Viewer : public WindowQt, public KeyMappingManageable<Viewer>
 
     void needUpdate();
 
+    /// Event sent after a mouse press event has been processed, but before emitting needUpdate()
+    void onMousePress( QMouseEvent* event );
```
MathiasPaulin, line 185:
```suggestion
    [[deprecated]] void onMousePress( QMouseEvent* event );
```

Note that for deprecated method, some IDE will not automatically complement their name from any prefix and will display the available method as strikethrough. This might discourage users to abuse from ready-to-be-removed features ;)

src/Gui/Viewer/Viewer.hpp
```diff
@@ -181,6 +181,13 @@ class RA_GUI_API Viewer : public WindowQt, public KeyMappingManageable<Viewer>
...
+    /// Event sent after a mouse release event has been processed, but before emitting needUpdate()
+    void onMouseRelease( QMouseEvent* event );
```
MathiasPaulin, line 187:
```suggestion
    [[deprecated]] void onMouseRelease( QMouseEvent* event );
```

src/Gui/Viewer/Viewer.hpp
```diff
@@ -181,6 +181,13 @@ class RA_GUI_API Viewer : public WindowQt, public KeyMappingManageable<Viewer>
...
+    /// Event sent after a mouse move event has been processed, but before emitting needUpdate()
+    void onMouseMove( QMouseEvent* event );
```
MathiasPaulin, line 189:
```suggestion
    [[deprecated]] void onMouseMove( QMouseEvent* event );
```

PR #856, opened 2022-01-18
[tests] fix #855 epsilon so that io test work with double

### LINKED ISSUE #855: Test IO/VolumesLoader fails when compiled with Scalar=double
Error logs: https://github.com/STORM-IRIT/Radium-Engine/actions/runs/1709898742

The problems seems to come from the fact that volume files are stored in `float`: numbers are loaded from files, casted to `double`, and then compared to `double` constants with an epsilon value in `double`. This test is very likely to fail and would require a `float` epsilon value.

dlyr:
Increase epsilon booster so that io tests work with Scalar=double, while the reader and stored data are floats.

### DIFF
--- a/tests/unittest/IO/volumeloader.cpp
+++ b/tests/unittest/IO/volumeloader.cpp
@@ -93,7 +93,12 @@ TEST_CASE( "IO/VolumesLoader", "[IO]" ) {
         auto binsize = volumeData->binSize();
         REQUIRE( ( Math::areApproxEqual( binsize[0], 1._ra ) &&
                    Math::areApproxEqual( binsize[1], 1._ra ) &&
-                   Math::areApproxEqual( binsize[2], 1.4_ra ) ) );
+                   // if Scalar = double, since reader use float, increase epsilon by float/double
+                   // epsilon ratio (test hack ;))
+                   Math::areApproxEqual( binsize[2],
+                                         1.4_ra,
+                                         Scalar { std::numeric_limits<float>::epsilon() /
+                                                  std::numeric_limits<double>::epsilon() } ) ) );
 
         LOG( logINFO ) << "computing gradients for data/Lobster.pvm.";
         volumeData->computeGradients();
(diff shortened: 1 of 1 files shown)


tests/unittest/IO/volumeloader.cpp
```diff
@@ -93,7 +95,9 @@ TEST_CASE( "IO/VolumesLoader", "[IO]" ) {
         auto binsize = volumeData->binSize();
         REQUIRE( ( Math::areApproxEqual( binsize[0], 1._ra ) &&
                    Math::areApproxEqual( binsize[1], 1._ra ) &&
-                   Math::areApproxEqual( binsize[2], 1.4_ra ) ) );
+                   // if Scalar = double, since reader use float, increase epsilon by float/double
+                   // epsilon ratio
+                   Math::areApproxEqual( binsize[2], 1.4_ra, 436870912_ra ) ) );
```
nmellado, line 100: Couldn't we have something like this (with the proper syntax):
```
Math::areApproxEqual( binsize[2], 1.4_ra, epsilon<float>()/epsilon<double>() ) ) );
```

dlyr: yes we can

tests/unittest/IO/volumeloader.cpp
```diff
@@ -93,7 +93,12 @@ TEST_CASE( "IO/VolumesLoader", "[IO]" ) {
         auto binsize = volumeData->binSize();
         REQUIRE( ( Math::areApproxEqual( binsize[0], 1._ra ) &&
                    Math::areApproxEqual( binsize[1], 1._ra ) &&
-                   Math::areApproxEqual( binsize[2], 1.4_ra ) ) );
+                   // if Scalar = double, since reader use float, increase epsilon by float/double
+                   // epsilon ratio (test hack ;))
+                   Math::areApproxEqual( binsize[2],
+                                         1.4_ra,
+                                         Scalar { std::numeric_limits<float>::epsilon() /
+                                                  std::numeric_limits<double>::epsilon() } ) ) );
```
nmellado, line 101:
```suggestion
                                                  std::numeric_limits<Scalar>::epsilon() } ) ) );
```

dlyr: This works for both float and double on my computer.

nmellado (approved):

PR #857, opened 2022-01-19
Set correct alpha values for Renderer exported frames

MathiasPaulin:

When exporting an image captured from the Renderer, the alpha channel depends on the exported AOV and do not allow proper displaying of the picture due to alpha blending problems when some transparent objects were drawn.

This PR fix this by 
 - modifying ForwardRenderer AOV initialization (Color::Alpha, e.g. {0, 0, 0, 0} instead of Color::Black, e.g. {0, 0, 0, 1}) so that there is no background in the AOV exported picture,
 - modifying the alpha-channel of an exported frame so that any alpha value greater than 0 is set to 1 while any 0 alpha value is preserved.

These two modifications allow to export png images with no background but solid color, even for transparent objects

Exported frame has an alpha channel that depends on drawn objects

Exported frame has alpha value of 0 in the background but 1 for all drawn pixels.

no

* **Other information**:

### DIFF
--- a/src/Engine/Rendering/ForwardRenderer.cpp
+++ b/src/Engine/Rendering/ForwardRenderer.cpp
@@ -272,7 +272,7 @@ void ForwardRenderer::renderInternal( const Data::ViewingParameters& renderData 
     else {
         glDisable( GL_POLYGON_OFFSET_FILL );
     }
-    static const auto clearZeros = Core::Utils::Color::Black().cast<GL_SCALAR_PLAIN>().eval();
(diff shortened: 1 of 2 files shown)


dlyr:
I understand the problem, but I'm not fine with the solution of discarding alpha when writing file.
We should compute final alpha correctly, what is shown in the issue is that the lines' alpha are set as final value instead of blending with background alpha. As shown in the issue "expected" results, there are black borders around silhouettes wireframe lines.

PR #871, opened 2022-02-10
Version file auto update

dlyr:
This PR add a VERSION file that contains current radium version
The version is auto increase on master pushes (increase feature e.g. 1.0.7 -> 1.1.0)
and release-candidate pushes (increase bug 1.2.3 -> 1.2.4)

Check [here](https://github.com/STORM-IRIT/Radium-Engine/actions/runs/1823524409) to see an example run

I have removed VERSION_TYPE since it's not handle correctly by cmake project_version (but github action actually handles -alpha -beta -rc ...). Adding this kind of version will add some ad hoc code in the cmake file to setup variables correctly.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -59,20 +59,6 @@ if(DOXYGEN_FOUND)
     set(RADIUM_DOC_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/)
     set(RADIUM_MAIN_DOC_FILE ${CMAKE_CURRENT_SOURCE_DIR}/main.md)
 
-    find_package(Git QUIET)
-    if(GIT_FOUND AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/../.git")
-        execute_process(
-            COMMAND ${GIT_EXECUTABLE} rev-parse --short HEAD
-            WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} OUTPUT_VARIABLE GIT_CHANGESET
-        )
-        if(GIT_CHANGESET)
-            # remove new line sometime appearing in git changeset
-            string(REGEX REPLACE "\n$" "" GIT_CHANGESET "${GIT_CHANGESET}")
-        endif()
-    else()
-        set(GIT_CHANGESET "")
-    endif()
-
     # Get awesome css
     set(RADIUM_DOXYGEN_AWESOME_CSS_ROOT
         "${CMAKE_CURRENT_SOURCE_DIR}/../external/doxygen-awesome-css"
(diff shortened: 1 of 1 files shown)


.github/workflows/push-master-ci.yml
```diff
@@ -13,6 +13,36 @@ env:
...
+      - name: Read VERSION
+        id: read
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: ./VERSION
+      - name: trim VERSION
+        id: set
+        run: |
+          echo "::set-output name=version::${{ steps.read.outputs.content }}"
+          [[ "${{ github.ref_name }}" == "master" ]] && echo "::set-output name=fragment::feature" || echo "::set-output name=fragment::bug"
```
nmellado, line 29: Can you add a comment explaining how to set the `bug` and `feature` keys ? Should it be in the commit message, or in the PR label ?

.github/workflows/push-master-ci.yml
```diff
@@ -13,6 +13,36 @@ env:
...
+      - name: Bump release version
+        id: bump_version
+        uses: christian-draeger/increment-semantic-version@1.0.2
+        with:
+          current-version: "${{ steps.set.outputs.version }}"
```
nmellado, line 34: What happens when the variables are not set at the previous step ? It can happen, right ?

.github/workflows/push-master-ci.yml
```diff
@@ -13,6 +13,36 @@ env:
...
+      - uses: actions/checkout@v2
+      - name: Read VERSION
+        id: read
+        uses: juliangruber/read-file-action@v1
+        with:
+          path: ./VERSION
+      - name: trim VERSION
+        id: set
+        run: |
+          echo "::set-output name=version::${{ steps.read.outputs.content }}"
```
nmellado, line 28:
```suggestion
          echo "::set-output name=version::${{ steps.read.outputs.content }}"
          # the variable `name` encodes the branch identifier: master (feature) or release-candidate (bug)
```

.github/workflows/push-master-ci.yml
```diff
@@ -13,6 +13,36 @@ env:
   ext-dir: $GITHUB_WORKSPACE/external/install/
```
nmellado, line 14:
```suggestion
# This pipeline computes the version number as follow: V1.featureNumber.bugNumber
#  featureNumber is incremented when pushing on the `master` branch,
#  bugNumber is incremented when pushing on the `release_candidate` branch.
```

nmellado (approved):

nmellado:
I was thinking that we could replace the commit sha by the version number in the documentation.
I can do the changes if it is fine for you.

PR #878, opened 2022-02-25
[cmake] First use robust deployqt and then fixup bundle

nmellado:

Bugfix: cmake deployment on windows

`Fixup_bundle` is ran before `windeployqt`. In some cases, `fixup_bundle` fails to find Qt libraries, and so the deployment.

`windeployqt` is ran first, and `mixup_bundle` only has to take care of plugins.

No

* **Other information**:
This PR fixes the problem by changing the instruction order, however it seems that there are some cases where `fixup_bundle` is not needed anymore. As already mentioned, a cleaning pass would be great on these files, but I believe this is out of the scope of this PR.


dlyr (approved):

MathiasPaulin (approved):
There should be problems when installing plugins alongside the application, but only for some plugins that uses Qt components that are not used by the application and/or Radium lib. But as such plugins are unhappily to be deployed now, we can merge this and work after that on full plugin integration.

PR #879, opened 2022-02-26
Manage the installation of packages with full version capabilities support.

MathiasPaulin:

Now that we have an automatic version management, Client of the Radium package or of any Radium related packages might require for any compatible Radium version they need and have access to the version number of the installed packages.
This PR gives access to the version of installed Radium packages (Radium libs as well as Radium extension) and allow version checking when searching for packages.

This allow to use the installed package in any client by requesting a given version of Radium.

typical uses are : 
 - in the client `CMakeLists.txt' 
```
find_package(Radium 1.0.2 COMPONENTS ....) # note that option number is still optional
```
This will success if the requested major version (here `1` is equal to the installed version, while the minor version (here `0` is less or equal to the installed one and the patch version, if the minor version is equal, is less or equal to the installed patch version.

Once found, any radium package named `Package` (e.g. `Radium` for the main package) will export the cmake variable 
`Package_VERSION`, `Package_VERSION_MAJOR`, `Package_VERSION_MINOR`, `Package_VERSION_PATCH` (e.g. `Radium_VERSION`, `Radium_VERSION_MAJOR`, `Radium_VERSION_MINOR`, `Radium_VERSION_PATCH`

This is done using standard cmake functions and variables, see cmake doc at 
https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package
and
https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html

No version information was fully exposed from installed Radium packages and no version management was allowed when searching for Radium package

see PR description


CMakeLists.txt
```diff
@@ -13,14 +13,19 @@ set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
...
+file(STRINGS "${CMAKE_CURRENT_SOURCE_DIR}/VERSION" INPUT_Radium_VERSION)
+project(radiumproject VERSION ${INPUT_Radium_VERSION})
 
 site_name(VERSION_HOST) # read hostname to VERSION_HOST
 set(VERSION_HOST "${VERSION_HOST}" CACHE STRING "host of build" FORCE)
 
-set(RADIUM_VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH})
-set(RADIUM_VERSION_MAJOR ${PROJECT_VERSION_MAJOR})
+# Sets the Radium_VERSION* variables such that all components built in the Radium-buildtree are able
+# to access them. Client applications, libraries and pluginq will access these variable after their
```
dlyr, line 23:
```suggestion
# to access them. Client applications, libraries and plugin will access these variable after their
```

tests/ExampleApps/HeadlessLibAndApp/Application/CMakeLists.txt
```diff
@@ -3,7 +3,16 @@ if(APPLE)
     cmake_policy(SET CMP0042 NEW)
 endif(APPLE)
 
-project(HeadlessDemo VERSION 1.0.0)
+if(IN_LIBHEADLESS_BUILD_TREE)
+    project(HeadlessDemo VERSION ${RadiumHeadless_VERSION})
+else()
+    project(HeadlessDemo VERSION 1.0.0)
+endif()
```
dlyr, line 10:
```suggestion
project(HeadlessDemo VERSION 1.0.0)
```
Well I think we should not have different version depending on how apps or libs are compiled.
Let's keep this 1.0.0 version (same for other apps/lib) and keep Radium_VERSION for bundled lib  (core, engine, ...)

tests/ExampleApps/HeadlessLibAndApp/libHeadless/CMakeLists.txt
```diff
@@ -3,7 +3,19 @@ if(APPLE)
     cmake_policy(SET CMP0042 NEW)
 endif(APPLE)
 
-project(RadiumHeadless VERSION 1.0.0)
+project(RadiumHeadless VERSION 1.0.2)
+
+# Defines the RadiumHeadless version in the parent scope so that on can use this lib doing
+# add_subdirectory(...) and access to the
```
dlyr, line 9:
```suggestion
# Defines the RadiumHeadless version in the parent scope so that one can use this lib doing
# add_subdirectory(...) and access RadiumHeadless_VERSION
```

dlyr (commented):
Small typo, except that I approve.
Regarding the different proposition, having each example using its own version is cleaner than trying to get radium version. The example Apps are "good" boilerplate to start a new project and I think we need to keep them compilable on their own easily.

MathiasPaulin:
> The example Apps are "good" boilerplate to start a new project and I think we need to keep them compilable on their own easily.

I agree with that but it is not the case right now. Only a few apps demo have their own `CMakeLists.txt`.
Do we add them in this PR or do we let this question for a future PR ?

MathiasPaulin:
> For a future PR it's ok, I just argue not to add code dependent on directory structure in the `CMakeLists.txt` so lasts changes are fine for me.

Application demo are now usable as  standalone projects or within the `ExampleApps` project. Only the root project test for being called within the Radium build (without checking for any directory structure) to set some variables. The others have all the same `CMakeLists.txt`(except for their project name and dependencies).

tests/ExampleApps/CMakeLists.txt
```diff
@@ -6,121 +6,88 @@ project(RadiumExampleApps)
...
+    # add a convenience macro for adding a run target
+    macro(add_run_and_debug_targets TARGET)
+        add_custom_target(
+            run_${TARGET}
+            WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+            USES_TERMINAL
+            DEPENDS ${TARGET}
+            COMMAND ./${TARGET}
+        )
+    endmacro()
```
dlyr, line 29: I don't get the point here, either move this macro from CleanCppExtensions to RadiumSetupFunction, or remove the use of this macro.

tests/ExampleApps/CoreExampleApp/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+    set(CMAKE_INSTALL_PREFIX
+        "${CMAKE_CURRENT_BINARY_DIR}/installed-${CMAKE_CXX_COMPILER_ID}-${CMAKE_BUILD_TYPE}"
+        CACHE PATH "Install path prefix, prepended onto install directories." FORCE
+    )
+    message("Set install prefix to ${CMAKE_INSTALL_PREFIX}")
+endif()
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_DISABLE_SOURCE_CHANGES ON)
+set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
dlyr, line 29: Could we move all this duplicated code to some radium_function ?

tests/ExampleApps/CMakeLists.txt
MathiasPaulin: I prefer to remove the use of this macro. This kind of target seems useless to me. I think you added them some time ago, so I keep them.

tests/ExampleApps/CoreExampleApp/CMakeLists.txt
MathiasPaulin: No. 

If you want that each demo has its own CMakeLists.txt, there will be lot of duplicated code. But this kind of configuration is application specific (here following the "recommandation" of Radium) and should not be set by the Radium package : 
 - cmake_policy are not intended to be permanent as they are set by some version of cmake to the wanted default value
 - Enabling in source build or changes, default build type  and default installation directory  should be left to the discretion of the application developper.

dlyr (approved):

PR #881, opened 2022-03-01
[headless] Fix installation issue and segfault when loading a scene with no camera

MathiasPaulin:
I forgot a commit in PR #879. Here is the fixes
- when installing the RadiumHeadless library, the setting of Radium_DIR in the config.in cmake script was incomplete. This fix either set the Radium_DIR given when compiling the lib as a standalone library (then with a valid Radium_DIR) or uses the same directory assuming Radium is installed in the same directory (which happens when RadiumHeadless is compiled as part of Radium.
- when, in the headless demo, a scene is loaded and does not contains a camera, the application segfault.

These fixes were tested using an external radium app that find the package RadiumHeadless either from the Radium bundle or from another installation directory.


tests/ExampleApps/HeadlessLibAndApp/Application/main.cpp
```diff
@@ -42,8 +42,8 @@ int main( int argc, const char* argv[] ) {
         viewer.setDataFileName( filename );
     }
     viewer.loadScene();
-    viewer.compileScene();
     viewer.setCamera();
+    viewer.compileScene();
```
dlyr, line 46: Is the order here important ? if so maybe add a comment
```suggestion
// need to be called after setCamera() because ...TODO
    viewer.compileScene();
```

dlyr (approved):

tests/ExampleApps/HeadlessLibAndApp/Application/main.cpp
MathiasPaulin: You're right. done in the last push.

PR #890, opened 2022-03-29
New demos

dlyr:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Add an example that draw a textured quad.
The objective is also to discuss here best practices.

* **What is the current behavior?** (You can also link to an open issue here)
Fix #887 

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
Nope

* **Other information**:
See commit history, I have added comment and remove some code from the original example from slack's discussion.

PR #895, opened 2022-03-31
Generate texture coordinates for parametric primitives

hiergaut:

Update quad creation-> moved to new PR #914 
Create texCoords for all parametric primitives.

No texture coordinates are created by default

Texture coordinates are created for all parametric primitives, it allows to apply texture on these primitives

TODO
- [ ] fix parametric sphere texCoord generation (@dlyr : I'm on it)
- [ ] fix parametric torus texCoord genration (once sphere is ok)
- [ ] use texCoord generation in src/Engine/Data/DrawPrimitives (done in // with previous todo)
- [ ] add demo in tests/ExampleApps/DrawPrimitivesDemo (done in // with previous todo)


dlyr:
Thanks for the PR. However, I think texCoords generation should be optional (e.g. with a bool generateTexCoord=false as arg) since one may want pure geometry. texCoords will roughly double geometry consuption, and when we'll generalize to all type of primitives, this could be costly.

src/Core/Geometry/MeshPrimitives.cpp
```diff
@@ -3,6 +3,8 @@
 #include <Core/Math/Math.hpp> // areApproxEqual
 #include <Core/Types.hpp>
 
+#include <Engine/Data/Mesh.hpp>
```
MathiasPaulin, line 6: Use only Core data structure here. The Core Library should not be dependent on Engine

src/Core/Geometry/MeshPrimitives.cpp
```diff
@@ -719,6 +726,7 @@ TriangleMesh makePlaneGrid( const uint rows,
     result.setVertices( std::move( vertices ) );
     result.setNormals( std::move( normals ) );
     result.setIndices( std::move( indices ) );
+    result.addAttrib( Ra::Engine::Data::Mesh::getAttribName( Ra::Engine::Data::Mesh::VERTEX_TEXCOORD ), std::move( texCoords ) );
```
MathiasPaulin, line 729: Instead of the dependence to Engine, use here "standard"  tex coord attrib name here : in_texcoord

MathiasPaulin (dismissed):
This proposal might be helpful but, right now, do not correspond to the Radium component dependency graph constraint.
  - remove the introduced dependency from Core to Engine
  - allow caller to specify if tex_coord is needed (see how color is managed)
  - propose this for all predefined parametric primitives (they are not so numerous)

src/Core/Geometry/MeshPrimitives.cpp
hiergaut: I can't do that, writing this seems to be a future delayed bug even if it is a standard name.
result.addAttrib( "in_texcoord", std::move( texCoords ) );
There are a duplication in the source code of the same information "in_texcoord". If one day a developer wants to change the names of shader attributes, it is highly probable that this one will be not aware to change this line too.
This is why I prefer to move the attribute name information to the Core. If you have any preferences for the location in the Radium hierarchy, I'm interested in it. For example in Core::Utils::Attribs (Attribs.hpp) or new Core::Mesh or new Core::Geometry::Mesh. For the moment I have positioned the update code in Core::Geometry (MeshPrimitives.hpp).
Thanks for your feedback.

hiergaut:
I have not been able to test the PR compilation on a windows system yet.

src/Core/Geometry/MeshPrimitives.cpp
dlyr: Even if it seems a good improvement, I think you can split the PR in two, since the proposal of using a const array in Core will break some external programs.
The second PR can fix generic attrib name management, and move it on the Core side. Then we can test it so one can update his own code before accepting the second PR.

dlyr (approved):

nmellado:
Please update wrt to `release_candidate`

nmellado (approved):

nmellado:
@hiergaut could you please rebase on `release_candidate`, so we can merge your proposal ?

hiergaut:
Sorry I didn't see this message, rebase done.

dlyr:
Before merging : I'll rewrite history
maybe add example in DrawPrimitivesDemo if takes less than 30 min

nmellado:
@dlyr we have a demo planned here: https://github.com/STORM-IRIT/Radium-Engine/pull/890

dlyr:
Ok, I'll do the full demo (all primitive) later ;)

hiergaut:
I can't see the texture on the disk in the centre of the scene, despite the RADIUM_IO_ASSIMP option being enabled on my setup. This could take me more than 30 minutes. Otherwise I can display the texture coordinates directly during rendering for parametric type objects. Do you have a way to make custom shader quickly ?

nmellado:
@dlyr what do you want to do with the history, it seems to be fine to me.

dlyr:
add tag ;)

dlyr (changes requested):
Maybe need refactor.

dlyr:
Sure, it's a hurry lock before merge. I'll do a review after a quick check

src/Core/Geometry/MeshPrimitives.cpp
```diff
@@ -11,24 +11,27 @@ namespace Core {
 namespace Geometry {
 
 TriangleMesh makeXNormalQuad( const Vector2& halfExts,
-                              const Utils::optional<Utils::Color>& color ) {
+                              const Utils::optional<Utils::Color>& color,
+                              const bool& generateTexCoord ) {
```
dlyr, line 15: not sure if it's efficient to have const bool&, why not just bool ?

nmellado: yes, `bool` is better indeed.

nmellado: https://stackoverflow.com/a/11932667

hiergaut: Because this is a const parameter. I said a const value like you. Not a reference.

dlyr:
Ok I finally got my first test, but it seems to have some problems. The example below show a parametric sphere textured.
Screenshot from 2022-04-20 18-47-10

Possible solutions
1. only accept quad tex coord generation, postpone other parametric objects tex coord generation
2. fix texture coords computation and make a complete example before accepting the PR

I prefer solution 1. since it was the original purpose of the PR. The fix could be done quite quickly afterward.

src/Core/Geometry/MeshPrimitives.cpp
hiergaut: Sorry, I didn't see your message when I wrote mine. I probably forgot to refresh the page when I sent mine. So I wrote a duplicate message.

hiergaut:
I forgot to add the pole texCoords, it is easy to fix, I'll do this soon

dlyr:
@hiergaut that's why I think it's out of scope of this PR. There is no de facto easy solution, and IMHO the best one is to TopologicalMesh facilities that handles this kind of computation easily.

hiergaut:
I was able to calculate the texture coordinates for the torus, unfortunately I have to duplicate the vertices at the envelope. In the same way I can make the UV sphere without using a Topology mesh David.
I couldn't rebase from release-candidate, there is a merge conflict. I tried to solve the conflict directly in the interface, and I think I made a mess.
screen

dlyr:
Moved to #915 , @hiergaut please check torus tex coord are correctly merged on the new PR (I think it's ok)

dlyr:
I can do it by tomorrow

++ david

PR #899, opened 2022-04-07
Headless as component

MathiasPaulin:

This PR moves Headless library from example to Radium-Engine components.
It also improves the documentation on the library classes as well as allowing future development of interactive applications relying on glfw.

Headless library is in the Examples subdirectory

Headless library is a Radium component in the src subdirectory

No

* **Other information**:


src/Headless/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+        CLI/Option.hpp
+        CLI/Split.hpp
+        CLI/StringTools.hpp
+        CLI/Timer.hpp
+        CLI/TypeTools.hpp
+        CLI/Validators.hpp
+        CLI/Version.hpp
+    )
+
+    set(${ra_headless_target}_sources CLIBaseApplication.cpp CLIViewer.cpp
```
nmellado, line 39: Shouldn't we use `cmake/filelistHeadless.cmake` instead ?

src/Headless/OpenGLContext/OpenGLContext.cpp
```diff
@@ -62,6 +62,34 @@ OpenGLContext::OpenGLContext( const std::array<int, 2>& size ) {
...
+            context->mouseEventCalback(
+                button, action, mods, int( xpos / xscale ), int( ypos / yscale ) );
+        };
+        glfwSetMouseButtonCallback( m_glfwContext, mouseCB );
+
+        auto scrollCB = []( GLFWwindow* window, double xoffset, double yoffset ) {
+            auto context = static_cast<OpenGLContext*>( glfwGetWindowUserPointer( window ) );
+            float xscale, yscale;
+            glfwGetWindowContentScale( window, &xscale, &yscale );
+            // TODO : verify if this is correct
```
nmellado, line 88: This comment is a bit frustrating ;)
I looked at the doc, but I don't understand what could be wrong. Could you be more explicit ?

src/Headless/OpenGLContext/OpenGLContext.hpp
```diff
@@ -0,0 +1,145 @@
...
+struct GLFWwindow;
+
+namespace Ra {
+namespace Headless {
+
+/**
+ * This class defines GLFW based OpenGL Context.
+ * There are two use cases of such a context :
+ *   - headless rendering application.
+ *   - simple window with limited interaction
```
nmellado, line 17: Do we really need to write these use case ?
I mean: if new use case are added, we will have to update the list, which we will likely forget to do.

src/Headless/OpenGLContext/OpenGLContext.cpp
MathiasPaulin: I also read the doc and think this is correct. But as this was not tested, and due to past experience of wrong HDPI management (also called 1/4 window bug :p ) I just put here a reminder. But, sure, we can remove it without regrets.

src/Headless/OpenGLContext/OpenGLContext.hpp
MathiasPaulin: I think we do. This class is, in fact, a wrapper to glfw OpenGL context and window management but exposes only a few of glfw interaction capabilities. 
So either we use it for headless rendering (without user interaction) or we use it with simple interaction based on simplified keyboard, mouse, scroll and resize event wrt glfw capabilities. 
And even if one can develop quite complex interactive application with this OpenGLContext by using an external GUI system, e.g.,  imGUI, the interaction capabilities will be limited by the wrapping we do. 
If we change this wrapping in the future, we'll change also the doc and if a developper forget to do this, PR reviewers will tell them to add doc ....

nmellado: ok, it was not clear to me if these use cases were the only ones possible, or if other variants could be implemented.

src/Headless/CMakeLists.txt
MathiasPaulin: I'm not a great fan of the pattern 'cmake/filelist.cmake' which I found harder to maintain than direct list of files and that prevent some IDE facilities (when adding a new class, the filelist must be updated manually). As this is not part of the [Coding conventions for Radium](https://storm-irit.github.io/Radium-Engine/release-candidate/develCodingConvention.html) I did this. But feel free to modify and commit/push on this branch.

nmellado (approved):
Ready to merge after CI is over

PR #901, opened 2022-04-10
Improve texture management

hiergaut:

Improve code, new feature, new docs

I have problems when using the texture object alone without using the texture manager. The texture pointer management does not take into account the fact that the data is generated on a separate thread, in fact I want to modify the texel data on a different thread than the one that renders the texture. I would like to retrieve a texture by its name with the texture manager and not the textureParameters. I would like to guarantee that textureParameters of a texture is constant during the life of a texture and not give the possibility for a user to modify this attribute there giving to the texture an incoherent state. I would like the name of a texture not to be in the textureParameters but rather in the textureManager because in fact the creation of a texture does not require a name, only the textureManager needs this information. I have encountered other problems with texture management, my current goal is to display a stream on a texture, so I'll contribute to Radium in order to have the possibility of using an efficient solution.

I can update a texture in a separate thread the way I wanted to. To do this I added an Image class (Core) containing only the information specific to an image (resolution, format, image data). This class has been implemented with the OpenImageIO library features, it also has been implemented in another way. Today an image can be attached from an engine texture. This allows to not modify the behavior of the existing code as well as the subjacent codes (software using Radium for rendering, and using massively textures). This PR does not modify the behavior of the textures on the engine side. On the other hand, attaching an image to a texture implies a slight overhead when using these textures during the bind process (which is done several times per frame for each displayed texture). That's why it would be necessary to consider a texture engine as an image (inheritance) and therefore to manage not a texture manager but rather an image manager in the Core side. Otherwise another solution is to build a texture attached to an image, the texture is a view of the image with its own parameter OpenGL (Wrap, mignify, magnify). Several textures can watch the same image for example, with different parameters. But changing the way to build a texture means changing all the calling codes. My opinion is this PR is not ready for a merge. Before the merged, we should measure the impact of changing the constructor of a texture. In the Radium code we have to change some lines in (ForwardRenderer, Renderer), and discuss about the modification impacts. The problem comes rather from sub-adjacent applications like Painty or others that I don't know yet.

* **Other information**:


MathiasPaulin:
Thanks to think about generalizing texture management. BTW, this PR makes me think you go toward more advanced CPU Texture management than we have right now. If this is the case, and as the needs should also helps other developers, perhaps we should think about using openImageIO (https://github.com/OpenImageIO/oiio) as base for our image/texture manager instead of maintaining and refactoring code, that was not developed for general texture management and will quickly become sub-optimal. I'll open an issue to discuss this.

dlyr:
Reading this PR make me think we can move texels linearization out of Texture, but only in TextureManager. This kind of transformation should be on eventually on Core side (when Core have full Image data management, Engine should only take care of the GPU representation of the texture, agnostic to the actual data semantic)

src/Core/CMakeLists.txt
```diff
@@ -47,7 +51,7 @@ target_include_directories(
 )
 target_link_libraries(
     ${ra_core_target} PUBLIC OpenMeshCore OpenMeshTools Threads::Threads Eigen3::Eigen
-                             cpplocate::cpplocate PRIVATE std::filesystem
+                             cpplocate::cpplocate OpenImageIO PRIVATE std::filesystem
```
hiergaut, line 53: Private or Public ?, I don't know what is the difference.

hiergaut: "All the include directories following PRIVATE will be used for the current target only, i.e., appending the directories to INCLUDE_DIRECTORIES."
I think we only use OpenImageIO in Core, so I have to change to private. Right ?

hiergaut, line 53:
```suggestion
                             cpplocate::cpplocate PRIVATE std::filesystem OpenImageIO
```

MathiasPaulin: I've in mind a use case where a radium extension will need to access some oiio features directly : we are implementing a ray-tracer that should use the oiio TextureSystem to have proper mip-map management on CPU (among other oiio features). This might require to have the oiio dependency as a PUBLIC dependency of RadiumCore.
And if oiio is exposed in a public Core header, it should be PUBLIC here to have proper transitivity when linking with Core.

hiergaut:
Why not place the OpenImageIO lib in IO instead of Core?

MathiasPaulin (changes requested):
I recognize there are some important improvements drafted by this PR. 
But I' have concerns about the design choices made here. This should need a live (in real life and in front of a white board) discussion.

hiergaut:
I agree to talk about it, there are several ways to design this.

hiergaut:
testTexture

During the writing of this PR, I wrote a test case to validate the features I wanted to have for my application (40Hz texture update, resizing, update of a local transformation matrix to 60Hz or if possible 120Hz). It turns out that this test shows some artifacts, one of which does not depend on the implementation of this PR. It is a matter of moving an object at 120Hz, for example with a single thread I decide to update the local transformation matrix of an object to simulate a position sensor. In this test I simply perform a rotation around a main axis of a white quad without texture. And I observe an artifact. I suppose that updating the local transformation matrix is not limited by the renderer refresh rate, indeed if I make several modifications of this matrix per frame, the renderer will take the most recent modification. I also detected another artifact on a texture that changes from white to black at 30Hz, I was expecting to see gray but it's not the case. This artifact may be related to my implementation though. So before closing this PR, I ask you to take into account what is said here in order to be able to correct these problems in a future implementation of an Image class. Thanks for reading.

Rotate quad at 120Hz
rotateQuad120Hz

Blink tetxure 30HZ (white <-> black)
blinkTexture

nmellado:
Superseded by #982

PR #909, opened 2022-04-13
Use m_vertexAttrib instead of deprecated attributes

Ventona666:

Moving all deprecated attributes (m_normal, m_tangent, m_bitangent, m_texCoord) and m_vertex to m_vertexAttribs (which is an attribManager).

File loaders use deprecated attributes to load data.

Now file loaders use an attribManager (m_vertexAttrib) to store this data instead of deprecated attributes, all tests passed.

There is no breaking change.

### DIFF
--- a/src/Core/Asset/GeometryData.hpp
+++ b/src/Core/Asset/GeometryData.hpp
@@ -273,6 +273,8 @@ class RA_CORE_API GeometryData : public AssetData
 
     template <typename Container>
     inline void setVertexAttrib( std::string vertexAttribName, const Container& vertexAttribList );
+
+    inline bool hasVertexAttrib( std::string vertexAttribName ) const;
 };
 
 } // namespace Asset
--- a/src/Core/Asset/GeometryData.hpp
+++ b/src/Core/Asset/GeometryData.hpp
@@ -125,7 +125,8 @@ class RA_CORE_API GeometryData : public AssetData
     template <typename Container>
     inline void setPolyhedra( const Container& polyList );
 
-    /// Return the list of vertex normals
+    /// Return the list of vertex normals.
+    /// \note This list must be unlock.
     inline Vector3Array& getNormals();
 
     /// Return the list of vertex normals.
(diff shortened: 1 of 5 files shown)


nmellado:
Hum, weird, there is a commit incrementing Radium version. what happened ?

dlyr:
I propose to review the PR as is ... and I'll clean the history.

nmellado:
yep ok.
It's mentioned WIP in the title, so let's wait until it is ready.

src/Core/Asset/GeometryData.hpp
```diff
@@ -214,21 +234,74 @@ class RA_CORE_API GeometryData : public AssetData
...
+     *
+     * @tparam Container
+     * @param name
+     * @return Get container base on name (lock).
+     * @warning If AttribHandle corresponding to name doesn't exist, it's created
+     * and return. By using this method, user has read-write access to data, data is lock, when
+     * done call attribDataUnlock( std :: std::string name )
+     */
+    template <typename Container>
+    inline Container& getAttribDataWithLock( std::string name );
```
nmellado, line 266: Consider 
```suggestion
    inline Container& getAttribDataWithLock( const std::string& name );
```
instead (same for all the other methods).

nmellado: thx

Ventona666:
I must also modify libGltf plugin by adding attribDataUnlock() where is needed.

MathiasPaulin:
> In this proposal, several functionalities of the Attrib manager are replicated (e.g. access to an attrib from its name). I am not sure if it is a good idea to to that, as the loaders will have, _in fine_, to play with the attribute manager directly.
> 
> What do you think @dlyr @MathiasPaulin ?

I set the "WIP" tag as this PR is not yet ready to review. If you want to review, wait until the WIP tag is removed and the tag "Ready to review" is set.

I recall you that we already discussed (IRL) about splitting  the work @Ventona666 do into two PRs. So we do. 
  - This PR only removes deprecated attribs in GeometryData by implementing all the interfaces using the attribManager then mark these interfaces as deprecated and proposes a much more simpler wrapper over attribArrayGeometry as an alternative. Then, sure, it keeps (except only a few renaming) the same (redundant) API than before but depreciate this API.
  - A second PR will remove all these deprecated (and redundant) methods by using directly the wrapper on a MutliArrayGeometry.
 
Right now, this is WIP and subject to numerous changes before the "Ready To Review" tag will be set.

nmellado:
Please convert to draft if you don’t want reviews.

Issue #912, opened 2022-04-19
Memory Leak : Viewer makeCurrent() and doneCurrent() has side effect on Texture initialization


Yasoo31:
Example that reproduce the leak :
 ```cpp

    uint width = 3024;
    uint height = 2138;

    for(int i = 0; i < 100; ++i)
    {
        viewer->makeCurrent();
        std::cerr << "Boucle n° " << i << "\n";

        float* buffer = new float[width * height];
        memset( buffer, 0, sizeof( float ) * width * height );

        TextureParameters texParams = {
                    "Tex_0",
                     GL_TEXTURE_2D,
                     width,
                     height,
                     1,
                     GL_RED,
                     GL_R32F,
                     GL_FLOAT,
                     GL_CLAMP_TO_BORDER,
                     GL_CLAMP_TO_BORDER,
                     GL_CLAMP_TO_BORDER,
                     GL_LINEAR,
                     GL_LINEAR,
                     buffer };

        std::unique_ptr<Ra::Engine::Data::Texture> pTex = std::make_unique<Texture>( texParams );
        pTex->initializeGL();

        delete [] buffer;

        viewer->doneCurrent();
    }
 ```
 Visible memory leak when `makeCurrent()` and `doneCurrent()` are inside the loop, and not visible when they are outside the loop

dlyr:
As stated on support chat, `unique_ptr` dtor is called outside gl context, hence the associated memory is note cleaned from gpu (and maybe cpu/driver) memory causing the leak.
To fix this exemple, either move `makeCurrent/doneCurrent` outside of the loop or call explicitly `pTex.reset()` before `doneCurrent`

dlyr:
(todo add comment to Texture stating that dtor must be called in gl context)

dlyr:
(todo bis, use "contextObject" that makeCurrent at ctor, and doneCurrent at dtor, hence they will follow block ctor/dtor order)

PR #904, opened 2022-04-11
Adding environment texture support on Radium (skyboxes, irradiance maps, cube map loading)

MathiasPaulin:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)
- [ ] Clean git history

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This PR adds support for environment textures in Radium. By offering loading of several environment maps parameterization (per face cube map, cross cube maps, equirectangular parameterization) as well as simple pre-filtering operations (projection on 3 spherical harmonic bands for low frequency diffuse lighting), this PR will allow to augment the rendering capabilities on Radium.
This PR also adds support for loading/writing OpenEXR images of various format (simple scanline, multi-part images, deep images) through the use of tinyEXR (fork and patch for integrating tinyEXR as external in Radium-Engine at https://github.com/MathiasPaulin/tinyexr/tree/radium)
This PR also slightly extend the Forward Renderer API to allow any application using this renderer to draw skyboxes.
A demonstrator of this extension is added in the example apps.
Capture d’écran 2022-04-11 à 18 56 08

* **What is the current behavior?** (You can also link to an open issue here)

No environment map support in the Radium-Engine

* **What is the new behavior (if this is a feature change)?**

Environment map support in Radium and skybox drawing capabilities using a simple inheritance over ForwardRenderer.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

No

* **Other information**:

PR #913, opened 2022-04-19
Scoped context

dlyr:

Add scoped context behavior. One can ask WindowQt an OpengGLContext than is bound at variable creation, and unbound when the variable is deleted (end of scope).

Have to manage context activation with makeCurrent and doneCurrent.
In a block, some gl related variable might be deleted on block end, while the doneCurrent have to be called before block end explicitly, see #912 .

Such block binded context are obtained as an ScopedContext object with WindowQt::createScopedContext, ScopedContext ctor calls makeCurrent, while dtor calls doneCurrent. 
Hence the "make current count" is still in use, and multiple ScopedContext can be imbricated without issues.

Nope, but this new behavior can replace makeCurrent/doneCurrent in a futur PR.

* **Other information**:
Also fix possible segfault on cube texture gpu allocation when no data provided

### DIFF
--- a/src/Gui/Viewer/WindowQt.inl
+++ b/src/Gui/Viewer/WindowQt.inl
@@ -4,20 +4,20 @@
 namespace Ra {
 namespace Gui {
 
-class WindowQt::ScopedContext
+class WindowQt::ScopedGLContext
 {
   public:
-    explicit ScopedContext( WindowQt* window ) : m_window( window ) { window->makeCurrent(); }
-    ~ScopedContext() { m_window->doneCurrent(); }
-    ScopedContext( const ScopedContext& ) = delete;
-    ScopedContext& operator=( ScopedContext const& ) = delete;
+    explicit ScopedGLContext( WindowQt* window ) : m_window( window ) { window->makeCurrent(); }
+    ~ScopedGLContext() { m_window->doneCurrent(); }
+    ScopedGLContext( const ScopedGLContext& ) = delete;
+    ScopedGLContext& operator=( ScopedGLContext const& ) = delete;
 
   private:
     WindowQt* m_window;
 };
 
-inline WindowQt::ScopedContext WindowQt::activateScopedContext() {
-    return ScopedContext { this };
+inline WindowQt::ScopedGLContext WindowQt::activateScopedContext() {
+    return ScopedGLContext { this };
 }
 
 } // namespace Gui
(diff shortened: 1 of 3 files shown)


src/Engine/Data/Texture.cpp
```diff
@@ -132,7 +132,14 @@ void Texture::updateData( const void* data ) {
     } break;
     case GL_TEXTURE_CUBE_MAP: {
         // Load the 6 faces of the cube-map
-        void** texels = (void**)data;
+        void* nullTexels[6] { nullptr, nullptr, nullptr, nullptr, nullptr, nullptr };
```
nmellado, line 135: Maybe use 
```suggestion
        static const void* nullTexels[6] { nullptr, nullptr, nullptr, nullptr, nullptr, nullptr };
```
instead ?

src/Gui/Viewer/WindowQt.inl
```diff
@@ -0,0 +1,22 @@
+#pragma once
+#include <Gui/Viewer/WindowQt.hpp>
+
+namespace Ra {
+namespace Gui {
+
+class WindowQt::ScopedContext
```
nmellado, line 7: Question about the naming: is this necessary related to GL ? If yes, it could be named `ScopedGLContext`.

src/Engine/Data/Texture.cpp
```diff
@@ -132,7 +132,14 @@ void Texture::updateData( const void* data ) {
...
+        void** texels;
+
+        if ( data == nullptr )
+            texels = (void**)data;
+        else
+            texels = nullTexels;
```
hiergaut, line 141:
```suggestion
        if ( data == nullptr )
            texels = nullTexels;
        else
            texels = (void**)data;
```

src/Gui/Viewer/WindowQt.inl
dlyr: Yes it's related to GL. Do you suggest to rename the inner class only or also the activateScopedContext method ?
Other Qt/GL related fonction are named context, makeCurrent and doneCurrent, following the Qt's convention. 
So my first opinion is ok to rename inner class, keep method name

src/Engine/Data/Texture.cpp
dlyr: moved to #904

src/Engine/Data/Texture.cpp
dlyr: moved to #904

src/Gui/Viewer/WindowQt.inl
nmellado: Ok, sounds good, I guess this will be more readable.

nmellado (approved):
Ready to merge once updated wrt `release_candidate`

PR #920, opened 2022-04-27
Resources

dlyr:

Feature: retrun absolute (canonical) path instead of relative one. Ease comparison to check if opened file are from resrouces dir.

getResroucesPaht may return relative (with ../) path

* **Other information**:
May break user code if one rely on the fact that the path is relative.


nmellado:
Sounds good to me, but I do not use the system of ressources right now, so cannot really test. Maybe @MathiasPaulin will be able to provide more feedback.

dlyr:
I have a second concern, `CXX_FILESYSTEM_HAVE_FS` is define as private to core target, while user code can include `Core/Utils/StdFilesystem.hpp`. Do we expect client cmake to define filesystem on its own ?

dlyr:
I've made filesystem dependecy public, so user code can use filesystem (I don't know what happens if fs is not the same between user and radium-engine's build, for instance when one grab a pre-compiled radium lib).

MathiasPaulin (approved):

PR #921, opened 2022-04-27
Render parameters

dlyr:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix and feature

* **What is the current behavior?** (You can also link to an open issue here)
render parameter concat do not replace values and misses some concatenation

* **What is the new behavior (if this is a feature change)?**
all paramters are concatenated, and a new "copy" feature is added which replaces existing parameters values.

PR #931, opened 2022-04-29
[engine] fix #930, taking care if Scalar is double

### LINKED ISSUE #930: #921 do not compile with Scalar = double
since renderparameters use float as scalar vector, and there is no conversion done.

dlyr:

Kind of quick fix, improvement needed to do only necessary conversion, using template programming, is_same and enable_if, but I hade trouble making it right.

Bug fix

see #930 

### DIFF
--- a/src/Engine/Data/ShaderProgram.cpp
+++ b/src/Engine/Data/ShaderProgram.cpp
@@ -336,15 +336,30 @@ void ShaderProgram::setUniform( const char* name, const Scalar& value ) const {
     m_program->setUniform( name, static_cast<GL_SCALAR_PLAIN>( value ) );
 }
 
-template <>
-void ShaderProgram::setUniform( const char* name, const std::vector<Scalar>& value ) const {
+template <typename T,
+          typename std::enable_if<!std::is_same<T, GL_SCALAR_PLAIN>::value>::type* = nullptr>
+void scalarVectorAdapter( globjects::Program* prog,
+                          const char* name,
+                          const std::vector<T>& value ) {
     std::vector<GL_SCALAR_PLAIN> convertedValue;
-    /// \todo use is_same and enable_if to prevent copy when unnecessary
     std::transform( value.begin(),
                     value.end(),
                     std::back_inserter( convertedValue ),
                     []( Scalar c ) -> GL_SCALAR_PLAIN { return c; } );
-    m_program->setUniform( name, convertedValue );
+    prog->setUniform( name, convertedValue );
+}
+
+template <typename T,
+          typename std::enable_if<std::is_same<T, GL_SCALAR_PLAIN>::value>::type* = nullptr>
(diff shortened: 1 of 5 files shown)


src/Engine/Data/ShaderProgram.cpp
```diff
@@ -333,6 +334,16 @@ void ShaderProgram::setUniform( const char* name, const Scalar& value ) const {
     m_program->setUniform( name, static_cast<GL_SCALAR_PLAIN>( value ) );
 }
 
+template <>
+void ShaderProgram::setUniform( const char* name, const std::vector<Scalar>& value ) const {
+    std::vector<GL_SCALAR_PLAIN> convertedValue;
+    /// \todo use is_same and enable_if to prevent copy when unnecessary
+    for ( const auto& v : value ) {
+        convertedValue.emplace_back( v );
```
nmellado, line 342: Could you use https://en.cppreference.com/w/cpp/algorithm/transform instead ?

MathiasPaulin (commented):
Is float/double compilation tested in CI ? As the answer should be yes (if I understand all the CI workflow), why #921 passed the Ci ? there is perhaps a problem in the CI also ?

src/Engine/Data/ShaderProgram.hpp
```diff
@@ -47,6 +47,8 @@ class RA_ENGINE_API ShaderProgram final
     /// Uniform setters
     template <typename T>
     void setUniform( const char* name, const T& value ) const;
+    template <typename T, typename E>
```
nmellado, line 50: What is `E`?

dlyr: maybe something I should have deleted, from by test.
If you have any guess how to specialize with enable_if and same_type to have only conversion if T is std::vector< Scalar>  and Scalar is double

dlyr: maybe simply specialize for std::vector<double> the typename replacement do the rest ...

nmellado: We need to check, but a simple `if (same_type )` should be enough: it is a constexpr that will be known at compile time, and thus probably optimized.

dlyr: It’s not sufficient since both branch of the if have to be valid code for the template parameters. And we want to make a simple copy which is not valid if Scalar is not the same as GL_SCALAR_PLAIN

nmellado: Hum right,
Then we can have a local template function with a unique template parameter, a boolean (computed with same_type), which is specialized with the two behaviors.

nmellado:
Not sure to understand how it works, but apparently it does, so.. fine for me.

dlyr:
I’m not 100% sure it’s the best way to do it, see https://en.cppreference.com/w/cpp/types/enable_if 
and SFINAE

PR #932, opened 2022-04-30
Named color

dlyr:

Feature, add the possibility to initialize color from svg names.

TODO 
- [ ] clean history before merge


src/Core/Utils/Color.hpp
```diff
@@ -10,12 +10,20 @@ namespace Ra {
 namespace Core {
 namespace Utils {
 
+/**
+ * Get a named color from its name. Return the uint value of named svg color. The value is of the
+ * form 0xRRGGBB. RR is red, GG is green and BB is blue. \param name: the name of the color as given
+ * from svg spec. \return Color code, or 0 if \a name is not a valid svg color name (0 is
+ * corresponds to black). \see https://www.december.com/html/spec/colorsvg.html
```
nmellado, line 17:
```suggestion
 * Get a named color from its name. Return the uint value of named svg color. The value is of the
 * form 0xRRGGBB. RR is red, GG is green and BB is blue. 
 * 
 * \param name: the name of the color as given from svg spec. 
 * \return Color code, or 0 if \a name is not a valid svg color name (0 corresponds to black). 
 * \see https://www.december.com/html/spec/colorsvg.html
```

nmellado, line 17:
```suggestion
 * Get a named color from its name. Return the uint value of named svg color. The value is of the
 * form 0xRRGGBB. RR is red, GG is green and BB is blue. 
 * 
 * \param name: the name of the color as given from svg spec. 
 * \return Color code, or 0 if \a name is not a valid svg color name (0 corresponds to black). 
 * \see https://www.december.com/html/spec/colorsvg.html
```

src/Core/Utils/Color.hpp
```diff
@@ -10,12 +10,20 @@ namespace Ra {
...
+ */
```
nmellado, line 18: Do we need this function to be public ?
The private static function of the Color class seems to be enough to me.

src/Core/Utils/Color.hpp
```diff
@@ -10,12 +10,20 @@ namespace Ra {
...
+uint32_t getNamedColorCode( const std::string& name );
+
 /*!
  * Colors are defined as vector4, i.e. 4 Scalars in RGBA order.
  * displayable colors should have all their coordinates between 0 and 1.
  *
  * \warning Vector arithmetics can be used to add, substract, multiply or divide colors with
- * scalars and other colors. In that case, alpha is treated as any other component. It is fine for
+ * scalars and other colors. In that case, alpha is treate as any other component. It is fine for
```
nmellado, line 26: typo ?

dlyr:
```suggestion
 * scalars and other colors. In that case, alpha is treated as any other component. It is fine for
```

src/Core/Utils/Color.hpp
dlyr: maybe the clang-format will reformat this (at least when no newline... but let's try.

src/Core/Utils/Color.hpp
dlyr: Well the private static function need to use this one. My idea here is to have the map as a static in cpp instead of having it in .inl
I'm not sure how to do without getting the name to color map in headers. Suggestion welcome

nmellado: I don't understand: I see the conversation has been marked resolved, however the public function is still there.
Did you try @MathiasPaulin's proposal ?

nmellado (approved):

PR #934, opened 2022-05-09
[tests] fix #927

### LINKED ISSUE #927: ${RadiumHeadless_VERSION} is not set when configure Radium
`tests/ExampleApps/HeadlessDemo/CMakeLists.txt` refers `${RadiumHeadless_VERSION}` which is not set
cmake output is 
`-- [Tests] Configuring demo application HeadlessDemo with RadiumHeadless  and Radium 1.2.21`
(while the RadiumHeadless_VERSION is expected after RadiumHeadless ^^)

MathiasPaulin:
Closes #927


dlyr (approved):
Well that means that this "version variable" is something outdated ?

MathiasPaulin:
> Well that means that this "version variable" is something outdated ?

Yes, this is a remnant of the old independent library proposal. Now, as it is a component, the version is the same than the Radium version.

PR #939, opened 2022-05-12
Fix #902 Component file lists 

### LINKED ISSUE #902: Component file lists
As discuss in #899, cmake source file lists could be probematic, and headless lib defines directly the file list in it's CMakeLists.txt

I open this issue to track source list management choices, and discuss alterantive.

The current evolution of component management make me think it can be better to have source list defined in the same place as the component, not in a general `/cmake/` dir.
It's also good practice to list files explicitly and not using GLOB like stuff.
I like the idea to generate such file list with a script.
I don't get the mentionned problem : "prevent some IDE facilities (when adding a new class, the filelist must be updated manually)" is it fine to have the file list variable defined in some .cmake file or does it have to be in a CMakeLists.txt ? 

We can have a script that generate the file list in CMakeLists.txt, but it makes huge lists for some components.

We can have file lists in Component  base directory under `src/Component`, generated by a script.

We can stick to the current approach, each component have its file list in `cmake/`.

We can get rid of scripts, and let each component manage how the file lists are defined.

Comments welcom.

dlyr:

Fix #902 

source filelist are in /cmake

source filelist are in module dir


scripts/generateFilelistForModule.sh
```diff
@@ -1,114 +1,130 @@
...
 
 BASE=$1
 LOWBASE=${BASE,,}
-OUTPUT="../cmake/filelist${BASE}.cmake"
+OUTPUT="../src/${BASE}/filelist.cmake"
 
 echo "generate [${BASE}] filelist"
 echo "-- input files from         [../src/${BASE}]"
-echo "-- output vars prefiex with [${LOWBASE}]"
+echo "-- output vars prefixwith   [${LOWBASE}]"
```
nmellado, line 14:
```suggestion
echo "-- output vars prefixed with   [${LOWBASE}]"
```

scripts/generateFilelistForModule.sh
```diff
@@ -1,114 +1,130 @@
...
-echo  "# ----------------------------------------------------" >> ${OUTPUT}
-echo  "# ---------- DO NOT EDIT -----------------------------" >> ${OUTPUT}
-echo  "# ----------------------------------------------------" >> ${OUTPUT}
-echo  "# This file is autogenerate from a script:" >> ${OUTPUT}
-echo  "# run form scripts directory" >> ${OUTPUT}
-echo  "# $0 ${BASE}" >> ${OUTPUT}
-echo  "# ----------------------------------------------------" >> ${OUTPUT}
+echo  "# ----------------------------------------------------" >> "${OUTPUT}"
+echo  "# This file can be generated from a script:" >> "${OUTPUT}"
+echo  "# run form scripts directory" >> "${OUTPUT}"
```
nmellado, line 21:
```suggestion
echo  "# ran from the scripts directory" >> "${OUTPUT}"
```

scripts/generateFilelistForModule.sh
```diff
@@ -1,114 +1,130 @@
...
-echo  "# ---------- DO NOT EDIT -----------------------------" >> ${OUTPUT}
-echo  "# ----------------------------------------------------" >> ${OUTPUT}
-echo  "# This file is autogenerate from a script:" >> ${OUTPUT}
-echo  "# run form scripts directory" >> ${OUTPUT}
-echo  "# $0 ${BASE}" >> ${OUTPUT}
-echo  "# ----------------------------------------------------" >> ${OUTPUT}
+echo  "# ----------------------------------------------------" >> "${OUTPUT}"
+echo  "# This file can be generated from a script:" >> "${OUTPUT}"
+echo  "# To do so, run form ./scripts directory" >> "${OUTPUT}"
+echo  "# $0 ${BASE}" >> "${OUTPUT}"
```
MathiasPaulin, line 22: I propose to replace the two lines above by : 
echo  "# To do so, run \"$0 ${BASE}\" from ./scripts directory" >> "${OUTPUT}"

MathiasPaulin (approved):

PR #942, opened 2022-05-16
[engine] let ShaderProgram manager control default glsl version.

dlyr:

Feature: let control the default glsl "#version" string.

"#version 410" along with some struct def is injected automatically in all shader. Version verison can be controlled by shader configuration.

Let the user specify the default version. Do not inject version nor default struct if the shader code already starts with "#version"
Also switch to a default version coherent with opengl context version.

Maybe depending on user shader code.

* **Other information**:

- [ ] maybe specify version as major, minor, and generate the string according to OpenGL doc https://www.khronos.org/opengl/wiki/Core_Language_(GLSL)
- [ ] Check version compatibility with OpenGL context (maybe in another PR)
- [ ] Discuss if we need to insert version/default struct in shaders
- [ ]  use glslangValidator during CI ? but include directive are badly supported

### DIFF
--- a/src/Engine/Data/ShaderProgramManager.hpp
+++ b/src/Engine/Data/ShaderProgramManager.hpp
@@ -49,13 +49,14 @@ class RA_ENGINE_API ShaderProgramManager final
      * The shader sources corresponding to the configuration will be compiled, linked and verified.
      *
      *
-     * @param config the configuration of the programm to add to the collection
+     * @param config the configuration of the programm to add to the collection. If no version is
+     * set in config, getDefaultVersion() is used as version.
      * @return the created shader program. In case of compile/link/verify error, return false
      * @note ownership on the returned pointer is keep by the manager.
      * @warning this method is *not* reentrant
      */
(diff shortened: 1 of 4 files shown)


nmellado (commented):
Seems fine, however I have low expertise on this, it would be better to get validation from @MathiasPaulin

src/Engine/Data/ShaderProgramManager.hpp
```diff
@@ -112,6 +118,8 @@ class RA_ENGINE_API ShaderProgramManager final
     std::map<std::string,
              std::pair<std::unique_ptr<globjects::File>, std::unique_ptr<globjects::NamedString>>>
         m_namedStrings;
+
+    std::string m_defaultVersion { "#version 440" };
```
MathiasPaulin, line 122: Default version should be 410 to be supported on macos. 
This default version might be updated at runtime, once opengl context is active, by using `gl::glGetString( gl::GLenum( GL_SHADING_LANGUAGE_VERSION )` and transforming the returned string to remove the dot (e.g. 4.10 --> 410)

PR #947, opened 2022-05-18
fix multiple resources directories issue (#946)

### LINKED ISSUE #946: Missing resources when installing library with multiple resources directories
When installing a plugin which depends on an helper libs having multiple resources directories, only the last declared resource dir is installed alongside the plugin.
This is related on how the resource installation and registration is performed in RadiumSetupfunctions.cmake. 
Right now, only one resource dir is associated with a target.
Associating several resource directories to a target (library, plugin or application) can help developers in their resource management.

MathiasPaulin:
closes #946 

Allows to manage several resources directories per Radium artifact (library, plugin, application)

When a library with several resources directories declared (using the radium-cmake function install_target_resources), only the last declared is installed if this library is used as an helper-lib for a plugin or a dependency for other installable artifact.

A target store a list of associated resource directories instead of only one directory

no

* **Other information**:

### DIFF
--- a/cmake/RadiumSetupFunctions.cmake
+++ b/cmake/RadiumSetupFunctions.cmake
@@ -22,9 +22,9 @@ include(CMakePackageConfigHelpers)
 # dependency resources when configuring/installing a target.
 define_property(
     TARGET
-    PROPERTY RADIUM_TARGET_RESOURCES_DIRECTORY
-    BRIEF_DOCS "Identify the optional resource directory associated with a target."
-    FULL_DOCS "Contains a directory that will be linked "
+    PROPERTY RADIUM_TARGET_RESOURCES_DIRECTORIES
+    BRIEF_DOCS "Identify the optional resource directories associated with a target."
+    FULL_DOCS "Contains a list of directories that will be linked "
               "or installed when building or installing the target."
 )
 define_property(
@@ -483,8 +483,13 @@ ${CMAKE_INSTALL_PREFIX}/${RESOURCES_DESTINATION_DIR}/${RESOURCES_INSTALL_DIR}/${
 
     # Install in the install tree Identify the individual files( to preserve directory structure )
     # set the target properties
+    get_target_property(resourceDir ${ARGS_TARGET} RADIUM_TARGET_RESOURCES_DIRECTORIES)
+    if(${resourceDir} STREQUAL "resourceDir-NOTFOUND")
+        set(resourceDir "")
+    endif()
+    list(APPEND resourceDir ${ARGS_RESOURCES_DIR})
     set_target_properties(
-        ${ARGS_TARGET} PROPERTIES RADIUM_TARGET_RESOURCES_DIRECTORY ${ARGS_RESOURCES_DIR}
+        ${ARGS_TARGET} PROPERTIES RADIUM_TARGET_RESOURCES_DIRECTORIES "${resourceDir}"
     )
 
     # why do not set tgt prop if files is not set, even if there is files in resources directory ?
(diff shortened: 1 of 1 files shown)


MathiasPaulin:
> Before going deeper in the review, this mechanism works fine with shared libs and apps right ? 

Only tested with share libs and plugins (see below). All the apps I have were not impacted by the bug. If you have one, I'll do the tests.

> Could this PR take benefit of install_target_resources as done in e.g. configure_cmdline_radium_app ?

I'm not sure to understand your question as this PR modify install_target_resources function to fix the resource installation problem. 

Bellow a use case that allows us to identify the bug and to correct it.

Given a library, say `TheLibrary`, configured as 
```
find_package(Radium REQUIRED Engine)
...
target_link_libraries(
        ${PROJECT_NAME}
        PUBLIC
        Radium::Engine
)

configure_radium_library(
        TARGET ${PROJECT_NAME}
        TARGET_DIR ${PROJECT_NAME}
        PACKAGE_DIR ${CMAKE_INSTALL_PREFIX}/lib/cmake
        NAMESPACE ${PROJECT_NAME}
        FILES "${public_headers};${public_inlines}"
)
configure_radium_package(
    NAME ${PROJECT_NAME}
    PACKAGE_DIR ${CMAKE_INSTALL_PREFIX}/lib/cmake
    PACKAGE_CONFIG ${CMAKE_CURRENT_SOURCE_DIR}/Config.cmake.in
    PACKAGE_VERSION ${PROJECT_VERSION}
)

install_target_resources(
        TARGET ${PROJECT_NAME}
        RESOURCES_INSTALL_DIR ${PROJECT_NAME}
        RESOURCES_DIR ${CMAKE_CURRENT_SOURCE_DIR}/Shaders
)

install_target_resources(
        TARGET ${PROJECT_NAME}
        RESOURCES_INSTALL_DIR ${PROJECT_NAME}
        RESOURCES_DIR ${CMAKE_CURRENT_SOURCE_DIR}/MetaData
)

```

When installing this library, the two resources directories are installed in the right place. This was true before this PR and still true after.

Now, consider a plugin configured by 
```
find_package(Radium REQUIRED COMPONENTS PluginBase)
find_package(TheLibrary REQUIRED)

...
 target_link_libraries(
        ThePlugin
        Radium::PluginBase
        TheLibrary
    )

configure_radium_plugin(
        NAME ThePlugin
        HELPER_LIBS TheLibrary
        INSTALL_IN_RADIUM_BUNDLE
)

```

When installing this plugin (here in the Radium bundle), it is expected that the helper library `TheLibrary` (the .so or .dylib) and its resource (here, Shaders and MetaData) are copied in the directory of the plugin, (Radium-Bundle/Plugins/lib for the .so, Radium-Bundle/Plugins/Resources/TheLibrary/xxx for the resources directories xxx). 
This is the way plugins are installed and this worked fine before this PR if the library requires only one resource directory. 
But before this PR, in the above example, only the last registered resource directory (MetaData) was copied in the Plugin directory. 
This PR fix this and make all the resources directories to be copied at the right place.
This was tested with a library you have access to (libGltf - branch release-candidate) that can be included in all the Radium applications through a simple plugin or linked directly for applications that do not support plugins. 
Both usage works now. 
But I've not tested to install bundled application not relying on plugins to use this library. 
For installed bundled applications that rely on plugins to use this library, all works fine (on macos).

MathiasPaulin:
And, I forgot, resources associated with an app are installed with the app. So, there were no problems with app resources neither before this PR, neither after.

dlyr:
Maybe I still miss something, but I exepect `configure_radium_plugin` to call `install_target_resources`, as `configure_radium_app` do. Which do not seems to be the case. But maybe the question is out of the PR scope.

dlyr (approved):

nmellado (approved):

MathiasPaulin:
> Maybe I still miss something, but I exepect `configure_radium_plugin` to call `install_target_resources`, as `configure_radium_app` do. Which do not seems to be the case. But maybe the question is out of the PR scope.

You're right, you miss something : the problem is not with the resources associated with the plugin, that are managed - and this PR do nothing for this as it works well -  exactly as you expect,  but with the resources of the **helper libs** of the plugin that are different things.
 In the documentation, helper libs are defined as 
```
HELPER_LIBS Optional. List of libraries (local target or imported targets) the plugin depends on
```
and such library can have its own resources the plugin has not to be aware of.

If needed, I'll clarify this at our new RL meeting.

PR #948, opened 2022-05-24
Update doc

dlyr:
One doxygen doc is miss assigned to a namespace. I moved the relevant information in class documentation
Current behavior, see here: https://storm-irit.github.io/Radium-Engine/release-candidate/namespaceUi.html

I also set EXTRACT_ALL to NO and HIDE_UNDOC_CLASSES to YES, to prevent from have documentation for e.g. 
https://storm-irit.github.io/Radium-Engine/release-candidate/classQWindow.html or https://storm-irit.github.io/Radium-Engine/release-candidate/classdouble.html

BTW, what about moving CLI to externals ? I don't tink having doc https://storm-irit.github.io/Radium-Engine/release-candidate/namespaceCLI.html is a good idea.
For this PR I add CLI to exclude pattern.


dlyr:
After a small test, there is now warning for undoc class.

PR #950, opened 2022-06-01
Material Edition

rainDiX:

New feature

There is no widget to edit material. Editing a material's renderParameter does not update it's associated attributes. There is no way to get the constraints of a parameter.

Introduce an interface to edit materials with constrains stored in a JSON file and add a updateState() method to update the material attributes with the values of the renderParameters.
Add a  GUI component to edit material's parameter.

No

* **Other information**:

Screenshot of the MaterialEditionDemoApp : 

EditorDemo


src/Engine/Data/RenderParameters.hpp
```diff
@@ -193,6 +193,24 @@ class RA_ENGINE_API RenderParameters final
...
+     * @param name The name of the parameter to get
+     * @return true if the parameter exists
+     */
+    template <typename T>
+    bool containsParameter( const std::string& name ) const;
+
+    /**
+     * Get a typed parameter
+     * @tparam T the type of the parameter to get
+     * @param name The name of the parameter to get
```
nmellado, line 208:
```suggestion
     * @param name The name of the parameter to get
     * @throw std::out_of_range if the container does not have an parameter with the specified name
```

src/Engine/Data/Material.hpp
```diff
@@ -114,6 +116,22 @@ class RA_ENGINE_API Material : public Data::ShaderParameterProvider
     std::string m_materialName;
 };
 
+/**
+ * Editable interface for the material class
+ */
+class RA_ENGINE_API EditableMaterial
```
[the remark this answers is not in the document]
MathiasPaulin: We need to discuss this AFK. The edition capability offered by this PR is more general than for Material as it allows to edit any set of `Ra::Engine::RenderParameters`. This decorator is here to simplify material edition by requiring Materials to define their edition constraints.

cmake/RadiumSetupFunctions.cmake
```diff
@@ -484,7 +484,7 @@ ${CMAKE_INSTALL_PREFIX}/${RESOURCES_DESTINATION_DIR}/${RESOURCES_INSTALL_DIR}/${
     # Install in the install tree Identify the individual files( to preserve directory structure )
     # set the target properties
     get_target_property(resourceDir ${ARGS_TARGET} RADIUM_TARGET_RESOURCES_DIRECTORIES)
-    if(${resourceDir} STREQUAL "resourceDir-NOTFOUND")
+    if("${resourceDir}" STREQUAL "resourceDir-NOTFOUND")
```
dlyr, line 487: should it simply be 
```suggestion
    if(NOT "${resourceDir}")
```

src/Gui/MaterialParameterEditor/MaterialParameterEditor.cpp
```diff
@@ -0,0 +1,356 @@
...
+                                                   const std::string& name ) {
+
+    m_matInfoGroup->setVisible( false );
+    if ( !name.empty() ) { m_matInstanceNameLabel->setText( QString::fromStdString( name ) ); }
+    else {
+        m_matInstanceNameLabel->setText( "Parameter set" );
+    }
+    // clear the old control panel
+    m_matParamsLayout->removeWidget( m_parametersControlPanel );
+    delete m_parametersControlPanel;
```
dlyr, line 216: I hade some weird segfault on windows (in other code base), directly deleting widget, I haven't tested this one but maybe use 
```suggestion
    m_parametersControlPanel->deleteLater();
```
see https://doc.qt.io/qt-6/qobject.html#deleteLater

cmake/RadiumSetupFunctions.cmake
MathiasPaulin: According to https://cmake.org/cmake/help/latest/command/if.html, yes, this should be simplified.

src/Gui/MaterialParameterEditor/MaterialParameterEditor.cpp
MathiasPaulin: uhm ... not sure deleteLater is the solution. When deleting a widget, all child widgets are first deleted. Can you show me, by private channel, what you're doing in the other code-base ? Do you add custom widgets to the panel ? How are they managed ? I'll verify this on windows ASAP, with the edition demonstrator to see if the crash is due to this delete or to an eventually added externally added widget.I'll also remind in the documentation that the ControlPanel takes ownership of the added widgets.

cmake/RadiumSetupFunctions.cmake
MathiasPaulin: I verified, this should not be simplified. I don't know why but doing this breaks the installation of the Resources (when more than one resource directory is associated with a target) of client libraries and plugins. I'll investigate more deeply later. Right now, we need to keep the proposal.

src/Gui/MaterialParameterEditor/MaterialParameterEditor.cpp
dlyr: The crashed I had are not due to MaterialParameterEditor but to other similar widget.
If I delete directly the widget, they still receive some event, but maybe here it's not the case.
deleteLater simply ensure there is no current event propagation that can trigger event from a widget that will be deleted.
The following snippet crash on windows if I call delete child->widget() instead of deleteLater
```
        auto vlayout = layout();
        if ( vlayout ) {
            QLayoutItem* child;
            while ( ( child = vlayout->takeAt( 0 ) ) != nullptr ) {
                if ( child->widget() ) {
                    child->widget()->disconnect();
                    child->widget()->deleteLater(); // delete the widget
                }
                delete child; // delete the layout item
            }
        }
        else {
            vlayout = new QVBoxLayout( this );
        }
```

MathiasPaulin: I've tested on windows, the MaterialEditionDemo is now operational. I've not meet your problem.

MathiasPaulin:
Trying to make a simple example that show all edition capabilities and how to use the control panel independently of materials, I found several problems that should be fixed before this PR could be merged. I put the WIP tag until I fix everything.

dlyr (commented):
First round of review, ok for what I've seen, plus some proposal on my fork (for enum values)
 
https://github.com/dlyr/Radium-Engine/tree/material_json_metadata_v2

examples/MaterialEdition/main.cpp
```diff
@@ -0,0 +1,209 @@
...
+                 on_materialParametersModified );
+
+        m_dock->setWidget( m_matParamsEditor );
+        addDockWidget( Qt::LeftDockWidgetArea, m_dock );
+
+        connect(
+            getViewer(), &Ra::Gui::Viewer::rightClickPicking, this, &DemoWindow::handlePicking );
+    }
+
+    // Handle the picking event generated when clicking on the window with  <RightButton>
```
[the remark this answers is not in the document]
MathiasPaulin: Simple typo in the comment. The signal we connect to is named Ra::Gui::Viewer::rightClickPicking but we do not care how user action is translated into this signal by Radium Gui.
Then this do not depends on the keymapping config (at least not directly) but only on the selection/manipulation signal emitted by the viewer in the method Viewer::processPicking.

MathiasPaulin:
Still WIP : some TODO : range iterator over bijective association, some code factorization using template parameters ... will be done soon.

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Ra {
+namespace Core {
+namespace Utils {
+
+/**
+ * \brief Bijective association on a finite set of <key, value> pairs.
+ * \tparam T1 Type of one element of the pair
+ * \tparam T2 Type of the other element of the pair
+ * \note T1 must be different from T2
```
dlyr, line 14: Do this limit comes only from operator() ?
There is currently a bijective association in `Engine::Mesh` which would benefit of this new class. But it's string <> string.
```c++
    /// Core::Mesh attrib name to Render::Mesh attrib name
    using TranslationTable = std::map<std::string, std::string>;
    TranslationTable m_translationTableMeshToShader;
    TranslationTable m_translationTableShaderToMesh;
```

MathiasPaulin: This limit is not only from operator () but also from 'addAssociation 'and the constructors. 
If the two types are the same, compiler generates an error such as `error: multiple overloads of 'addAssociation' instantiate to the same signature`.
This can be solved by removing the capability to add pair whatever their elements' order (e.g. {T1, T2} or {T2, T1}) and restricting the construction using only pari (or parameters in the order {T1, T2}.
We can also change the operators () to two different operator names, e.g inject : T1 -> T2 and surject : T2 -> T1, instead of overloading.
But if types are different, I found very practical to have the () operator.
Perhaps we can rewrite this class using std::enable_if<std::is_same<T1, T2> .... > to define only the alowed functions. I'll try to do that.

MathiasPaulin: done using partial specialisation ( I do not success in using std::enable_if :( ). 
If you want to port your use case, you can  push your commit here or wait another PR specific to your fix.

dlyr: Well the PR is big enough to wait for Engine::Mesh.

dlyr:
To ease review, it may be a good idea to split the PR so that bijective association comes afterward.

First: material edition without enum and bijective association
Second: bijective association used for enum and other places in radium (eg Mesh attrib translator).

MathiasPaulin:
> it may be a good idea to split the PR so that bijective association comes afterward.

I don't think so. 
This PR is about RenderParameter edition, with a demonstration covering both pure parameter set and Material parameters. As there are uses cases of enum as Parameter, with big issues in edition without having a bijective association between the enum value (correctly typed and stored in the RenderParameter) and the string (used in the Gui), the PR should cover this uses cases. 

What I agree for, is  the usage of bijective association outside of RenderParameter edition  (e.g. on your mesh use case) that can come afterward.

dlyr:
Fine with that, so I'll review your proposal and your use cases in its current form.

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * \brief Gets the value associated to the key
+     */
+    Type2 operator()( const Type1& k ) const;
+
+    /**
+     * \brief Gets the value associated to the key
+     */
+    Type2 inject( const Type1& k ) const;
```
dlyr:
```suggestion
    Type2 value( const Type1& k ) const;
```

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * \brief Gets the value associated to the key
+     */
+    Type1 surject( const Type2& k ) const;
```
dlyr:
```suggestion
    Type1 key( const Type2& k ) const;
```

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,136 @@
...
+ * \tparam T1 Type of one element of the pair
+ * \tparam T2 Type of the other element of the pair
+ * \note If T1 and T2 refer to the same type, some operators will be removed.
+ */
+template <typename T1, typename T2, bool TypesAreEquals = std::is_same_v<T1, T2>>
+class BijectiveAssociation
+{
+  public:
+    using Type1 = T1;
+    using Type2 = T2;
```
dlyr:
```suggestion
    using key_type = T1;
    using value_type = T2;
```

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * \brief Constructor from { <T1, T2> } pairs
+     */
+    explicit BijectiveAssociation( std::initializer_list<std::pair<T1, T2>> pairs );
+    /**
+     * \brief Constructor from { <T2, T1> } pairs
+     * \note If T1 and T2 are the same type, this constructor is disabled.
+     */
+    explicit BijectiveAssociation( std::initializer_list<std::pair<T2, T1>> pairs );
```
dlyr: I think it might be confusing to forget about the direction key - value.
Having only one ctor is sufficient.
This comment is valid for other methods as well.

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,136 @@
...
+    /**
+     * \brief Gets the value associated to the key
+     * \note If T1 and T2 are the same type, this operator is disabled.
+     */
+    Type1 operator()( const Type2& k ) const;
+
+    /**
+     * \brief Gets the value associated to the key
+     */
+    Type2 operator()( const Type1& k ) const;
```
dlyr: Also this overloaded operator might be confusing, as I found in the RenderParameter usage.
If we consider the bijective association as a function y = f(x) != x = f(y)

MathiasPaulin: yes if its a function, but here, its an association.

src/Core/Utils/BijectiveAssociation.hpp
MathiasPaulin: I'll remove all overload and partial specialisation. I found some of them useful from a user point of vue (Given a unordered pair, if a have one element of this pair, I can get the other without inquiring which one is the key and which one is the value). But I agree to consider this as a function.

dlyr (commented):
Review in progress.

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * \brief Gets the key  associated to the value
+     */
+    key_type key( const value_type& k ) const;
+
+    /**
+     * \brief Gets an const iterator at beginning of the key to value map
+     */
+    typename std::map<key_type, value_type>::const_iterator begin();
```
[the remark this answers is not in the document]
MathiasPaulin: Well, I'll change this to `const_iterator begin() const noexcept;` (as in C++11 reference) to prevent the modification of the association through the iterator. The objective here is to have a read-only iterator.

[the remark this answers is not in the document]
dlyr: hence cbegin might be sufficient ?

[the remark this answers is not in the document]
MathiasPaulin: is cbegin sufficient for range based for ? I'll test it. But we can follow c++11 reference and define only 
```
const_iterator begin() const noexcept; 
const_iterator cbegin() const noexcept;
```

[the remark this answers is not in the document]
dlyr: Fine for me. And the user get a warning/error if it tries to get a non const iterator.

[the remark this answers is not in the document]
MathiasPaulin: `begin()` and `end()` functions are required for range-based for and algorithms. So, we have to define begin as well as cbegin with the above mentioned profiles.

src/Core/Utils/BijectiveAssociation.hpp
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * \brief Constructor from { <T1, T2> } pairs
+     */
+    explicit BijectiveAssociation( std::initializer_list<std::pair<key_type, value_type>> pairs );
+
+    /**
+     * \brief Add a pair to the relation.
+     * \return false if the pair was not added due to collision with an already inserted pair
+     */
+    bool addAssociation( std::pair<key_type, value_type> p );
```
dlyr, line 31: I think we could follow std::map naming, e.g. insert instead of addAssociation.

MathiasPaulin: I agree.

src/Engine/Data/RenderParameters.hpp
```diff
@@ -218,7 +363,23 @@ class RA_ENGINE_API RenderParameters final
...
 
     UniformBindableSet<TextureParameter> m_texParamsVector;
-    /**@}*/
+    /**\}*/
+};
+
+/**
+ * Interface to define metadata (constraints, description, ...) for the edition of parameter set
+ */
+class RA_ENGINE_API ParameterSetEditionInterface
```
dlyr, line 372:
```suggestion
class RA_ENGINE_API ParameterSetEditingInterface
```
As far as I know, edition is related to printing, editing is related to the edit action.

src/Engine/Data/Material.hpp
```diff
@@ -53,6 +53,9 @@ class RA_ENGINE_API Material : public Data::ShaderParameterProvider
      */
     inline void setMaterialName( std::string newName );
 
+    /// Load the material parameter description
+    static void loadMetaData( const std::string& basename, nlohmann::json& destination );
```
dlyr, line 57: Should this one moved to ParameterSeetEdit(ing)Interface ?

tests/unittest/Core/bijectiveassociation.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+        auto n = myTranslator.size();
+        REQUIRE( n == 4 );
+    }
+
+    SECTION( "Bijection between same types" ) {
+        BijectiveAssociation<std::string, std::string> myTranslator { { "One", "Un" },
+                                                                      { "Two", "Deux" } };
+        myTranslator.insert( "Three", "Trois" );
+        myTranslator.insert( { "Four", "Quatre" } );
```
dlyr, line 32:
```suggestion
        REQUIRE( myTranslator.insert( { "Four", "Quatre" } )); // first insert pass 
        REQUIRE( !myTranslator.insert( { "Four", "Quatre" } )); // seconid insert failed since already present
```

src/Engine/Data/Material.hpp
MathiasPaulin: I don't think so. MetaData are a per-class informations, so a static method in each class.

src/Gui/Widgets/ConstrainedNumericSpinBox.hpp
```diff
@@ -0,0 +1,52 @@
...
+     * \brief Validate a value
+     * \param s
+     * \return true if the value is valid according to the stored predicate, false otherwise.
+     */
+    inline bool isValid( T s ) const { return m_p( s ); }
+
+  private:
+    Predicate m_p = []( T ) { return true; };
+};
+
```
dlyr, line 33: Maybe move these function directly to ConstrainedNumericSpinBox

src/Engine/Data/RenderParameters.cpp
```diff
@@ -87,6 +112,20 @@ void RenderParameters::addParameter( const std::string& name, Data::Texture* tex
     m_texParamsVector[name] = TextureParameter( name, tex, texUnit );
 }
 
+void RenderParameters::addParameter( const std::string& name, const std::string& value ) {
+    auto it = m_enumConverters.find( name );
+    if ( it != m_enumConverters.end() ) { it->second->setEnumValue( *this, name, value ); }
```
dlyr, line 117:
```suggestion
   addParameter( name, it->second->getEnumerator( value );
```
Not tested, but seems clearer for me (I don't get why the converter should set the value, instead of the parameter set get the conversion results and set it).

dlyr (commented):
My last comments, I'm fine with the PR apart from theses.

src/Engine/Data/RenderParameters.cpp
MathiasPaulin: Asking the converter to set the value (instead of setting it directly) allows to preserve the underlying type of the enumeration. 

See the documentation about getEnumerator : 
```
* \note  This method does not respect the underlying type of enumerations as it returns
* always an int. Use setEnumValue to modify the parameter set in a type safe way.
```

nmellado (approved):

PR #925, opened 2022-04-28
Use multiIndexedLayer instead of deprecated attributes

Ventona666:
_Check if you branch history is PR compatible_
- Your branch need to be up to date with origin/master AND to have linear history (i.e. no merge commit).
- Update your git repository `git fetch origin` if origin is this remote
- Check with the script provided in `scripts/is-history-pr-compatible.sh`
- You must use clang-format style
_These checks are enforced by github workflow actions_
_Please refer to the corresponding log in case of failure_

_UPDATE the form below to describe your PR._

* **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Moving all depracated attributes (m_normal, m_tangent, m_bitangent, m_texCoord) and m_vertex to the attribManager of the multiIndexedLayer and m_edge, m_faces, m_polyhedron into layers.

* **What is the current behavior?** (You can also link to an open issue here)
File loaders use deprecated attributes to load data. 

* **What is the new behavior (if this is a feature change)?**
Now file loaders use an attribManager and layers to store this data instead of deprecated attributes, all tests passed. This will then allow to use this structure directly in the engine and avoid data duplication.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
There is no breaking change except unlock attrib an layer when user lock it to have write access on it.

 - Other information:
 Numerous functions were marked deprecated by this PR and their implementation were changed to correspond to the recommended alternative, sometime using some introduced helper functions (marked as such by a note in their documentation). 
 At the end of the deprecation process, all these functions (and the related helper function) will be removed from the source code.

PR #951, opened 2022-06-01
Add QuadMesh

Ventona666:

This PR add quadMesh support by templating the current class polyMesh, so if the type of mesh is know at compilation time, we are able to alloc the real size of a quadMesh instead of using VectorNui of polyMesh.

A quadMesh is loaded as a polyMesh, so we loose some time and space during the execution.

Now quadMesh is fully support, a part of calculation time is during the compilation and reduce memory usage.

Yes, by templating getFaces() and the class polyMesh.

* **Other information**:
* This PR can only be remove from draft when the PR #925 will be merged into release candidate.


MathiasPaulin:
This PR was already included in #925.

PR #954, opened 2022-06-13
[engine] Add convenience function to enable/disable pervertex color

nmellado:

Convenience feature

`PointCloudComponent` computes if colors need to be rendered per vertex when at loading.
Changing the value of this parameter afterwards (for instance when adding colors computationally) involves several actions downside.

Add convenience functions to get and set the `pervertexcolor` attribute of the default `BlinnPhongMaterial` set by the `PointCloudComponent`.

No

* **Other information**:

### DIFF
--- a/src/Core/Geometry/MeshPrimitives.inl
+++ b/src/Core/Geometry/MeshPrimitives.inl
@@ -72,26 +72,47 @@ makeParametricSphere( Scalar radius, const Utils::optional<Utils::Color>& color,
     };
 
     // For general vertices retrieve normals and texCoords from vhIndex
-    auto wedgeSetterGeneric =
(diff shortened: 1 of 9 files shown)


MathiasPaulin (commented):
I recognize the usefulness of the proposal. BTW, I think that the implementation is not satisfying. If you agree, I'll push another proposal on this PR extending the Material API to add a `ColoredByVertexAttrib` (or similar) method, thus removing the need fo the dynamic_cast. This will also make any material/BSDF supporting per-vertex color parameter  properly managed.

MathiasPaulin:
The PR description might be updated to match the PR proposal (even if my commits are rejected).

dlyr (approved):

PR #956, opened 2022-06-21
[gui] prevent BaseApplication to load a plugin twice

MathiasPaulin:

It could happen that an application loads a plugin twice : first, a plugin installed in the Plugin Directory of the Radium Bundle, and second the same plugin from a directory stored in the QtSettings but that refer to the Bundle's plugin directory.
This can cause crash of the application depending on the plugin initialization.

When a plugin is loaded from a directory stored in the QtSettings refering to the plugin installation directory into the Radium Bundle, any plugin in this directory is loaded twice.

Plugins are loaded only once from the Bundle's Plugin directory

* **Other information**:


dlyr:
If I understand well the PR's code, it skips directory that are equivalent to default plugin directoy.
If a plugin file is duplicated in two locations, then it's still loaded twice, right ?
Also if two use directories points to the same location, then the whole directories are loaded twice, right ?

MathiasPaulin:
> If I understand well the PR's code, it skips directory that are equivalent to default plugin directoy.

yes. 

> If a plugin file is duplicated in two locations, then it's still loaded twice, right ?

yes, if the two locations are registered. 
Preventing this will require more work on plugin management to be able to identify a plugin before loading it, to manage set of loaded plugins in order to prevent double loading, and, to allow plugin unloading/reloading. But this is far more work than this simple fix that solve the following problems : 
  when you run a example app from the install tree, this app loads installed plugin from the Radium bundle. When you run the same example app from the build tree, this app can't load installed plugins from the bundle (this is due on how we search path for the plugins directory in Resources.cpp). The only way you have to load plugins from an app ran from the buildtree is to add the plugin directory in the QSettings file. This simple fix just handle the case one have added the Bundle's Plugin directory in the QSettings to be able to load plugin from the buildtree and prevent double loading, when the app is ran from install tree or ran linked to the installed bundle. 

> Also if two use directories points to the same location, then the whole directories are loaded twice, right ?

No, two directories pointing to the same location are "equivalent" : https://en.cppreference.com/w/cpp/filesystem/equivalent.
but yes if we register twice the same directory, it will be loaded twice.

I'll push in the next few hours another fix, this one being removed, to prevent double loading based on plugin filename instead of just Bundle-Registered directory equivalence. But proper plugin load/unload/replace/explore should be a good "should have" for future.

dlyr:
The test isn't done only between default plugin path and user registered plugin path ?
What happens if one add the same directory (even with different path) with `addPluginDirectory` ?

PR #957, opened 2022-06-22
[ci] add qtversion to exteral cache key

dlyr:
Now external (i.e PowerSlider) uses Qt.
This Pr adds qtversion to the cache key of externals, such that the correct version is pulled from cache database.

### DIFF
--- a/.github/workflows/build-matrix.yml
+++ b/.github/workflows/build-matrix.yml
@@ -185,7 +185,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: external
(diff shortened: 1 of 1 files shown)


MathiasPaulin:
Having externals depending on Qt is not the case right now. 
This needs is introduced in the PR Material Edition #950. 
Perhaps this fix should be included in the PR that require it as only #950 will be able to test the efficiency of this fix ...

dlyr:
Right, you can cherry pick the commit in PR#950 (and close this PR when done).

dlyr:
(oups I already closed it ;))

MathiasPaulin:
I'll take the fix and make the commit ;)

PR #965, opened 2022-07-13
Set default GLSL version for shaders according to detected OpenGL context

MathiasPaulin:

instead of default `#version 410` added to all managed shaders, use the GLSL version corresponding to the OpenGL version of the context bound when initializing engine.

shader header default to `#version 410`

shader header default to `#version context-dependant-version`

NO

* **Other information**:

Need to be tested as some shader might not compile if GLSL version is less than 410


dlyr (approved):
Next step, let BaseApplication and CLIViewer have configurable version.

PR #893, opened 2022-03-30
Components management

dlyr:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines
- [x] Docs have been added / updated (for bug fixes / features)

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This PR introduce a cleaner way to define components for the Radium-Engine distribution and configure Core, IO, PluginBase, Gui and Engine as components of the RAdium package.
This is done by : 
 - modifying the cmake configuration scripts of all the base components : Core, Engine, IO, PluginBase and Gui
 - modifying the global project installation procedure.
 - defining an unified way to add new components to the Radium distribution

* **What is the current behavior?** (You can also link to an open issue here)

Components are configured globally exhibits some robustness issues (mainly concerning components inter-dependencies)

* **What is the new behavior (if this is a feature change)?**

Component can be configured individually and requested individually in a `find_package` command.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

No

* **Other information**:

PR #969, opened 2022-07-15
[cmake] Take into account QT_DEFAULT_MAJOR_VERSION.

dlyr:

Feature: adds the possibility to choose qt version, with both qt libs installed system wide.

if qt6 is installed system wide, there is no option to use qt5

QT_DEFAULT_MAJOR_VERSION is checked to choose specified qt version, as stated in qt doc. 
Nonetheless, we need to include only this version, otherwise I got a build issue

Nope.

* **Other information**:
Maybe need to add doc


src/Gui/CMakeLists.txt
```diff
@@ -10,6 +10,8 @@ include(QtFunctions)
 
 # Find packages
 find_qt_package(COMPONENTS Core Widgets OpenGL Xml REQUIRED)
+set(QT_DEFAULT_MAJOR_VERSION ${QT_DEFAULT_MAJOR_VERSION} PARENT_SCOPE)
```
nmellado, line 13:
```suggestion
if(QT_DEFAULT_MAJOR_VERSION)
     set(QT_DEFAULT_MAJOR_VERSION ${QT_DEFAULT_MAJOR_VERSION} PARENT_SCOPE)
 endif()
```

nmellado (commented):
Sounds good.
Do we really need to check for qt version issues in both Gui and PluginBase `Config.cmake.in` ?

dlyr:
> Sounds good. Do we really need to check for qt version issues in both Gui and PluginBase `Config.cmake.in` ?
Not sure, I have not checked if plugin base is always included before Gui, at least Gui is not sufficient (I have tested first with Gui only, and have some troubles ...)

nmellado:
PluginBase requires Gui if I remember well.

MathiasPaulin:
TL;DR 
  **it works and dependency between Radium component are well managed since PR Components management #893. Just search for the one you need and the components its depends on will be configured.**

Reminder on the way components are configured : 

When searching for the package Radium, the user need to specify the component he needs. 
If no components are given only Core is configured by the cmake script. 
If any component is given, the dependencies of this component are correctly managed. 
So, here : 
 - PluginBase depends on Gui
    - If the user  just do a find_package(Radium COMPONENTS PluginBase), the Gui is first searched as a dependency (lines 7 to 16 of PluginBase/Config.cmake.in), resulting in lines 8 to 17 of the generated RadiumPluginBaseConfig.cmake available in the build tree or installed with the Bundle
    - If the user search first for Gui, by doing (Radium COMPONENTS Gui PluginBase), the dependency for pluginBase is here and no more work is needed, only configuring PluginBase
    - If the user search for Gui after PluginBase, by doing by doing (Radium COMPONENTS PluginBase Gui) , as Gui was already configured as a dependency of pluginBase, all is OK.

MathiasPaulin:
> Ok maybe I have not explained well.

OK, I saw. In this case, just move the lines 36-37
```
    include(${CMAKE_CURRENT_LIST_DIR}/../QtFunctions.cmake)
    find_qt_dependency(COMPONENTS Core Widgets OpenGL Xml REQUIRED)
```
in Gui/Config.cmake.in just before line 32 (`if(EXISTS "${CMAKE_CURRENT_LIST_DIR}/../PluginBase/RadiumPluginBaseConfig.cmake")` so that Qt is configured before including PluginBase. Like this, no need to duplicate `check_and_set_qt_version("@QT_DEFAULT_MAJOR_VERSION@")`

MathiasPaulin:
> Like this, no need to duplicate `check_and_set_qt_version("@QT_DEFAULT_MAJOR_VERSION@")`

Well, thinking twice, not sure this will be enough. If your proposal works, stay with it right now until we solve the circular dependencies between Gui and PluginBase later

dlyr:
If everything is ok, could someone approve the PR ?
I've fix error message, and colon spacing btw.

cmake/QtFunctions.cmake
```diff
@@ -20,7 +20,9 @@ macro(find_qt_package)
         set(MY_OPTIONS_COMPONENTS "")
     endif()
 
-    find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} QUIET)
+    if(NOT QT_DEFAULT_MAJOR_VERSION OR QT_DEFAULT_MAJOR_VERSION STREQUAL "6")
```
nmellado, line 23: We may have a problem here, when the user set `QT_DEFAULT_MAJOR_VERSION` to 6, but find_package does not find it. As we check it as `quiet`,  we will go to line 28 (`find_package(Qt5 5.15 ...)`. If Qt5 is found, we may have a mismatch between the user requirement and the actual configuration, without any error or warning.

cmake/QtFunctions.cmake
```diff
@@ -20,7 +20,9 @@ macro(find_qt_package)
         set(MY_OPTIONS_COMPONENTS "")
     endif()
```
MathiasPaulin, line 22:
```suggestion
if (QT_DEFAULT_MAJOR_VERSION STREQUAL "6")
    set(QT6_OPTION REQUIRED)
else()
    set(QT6_OPTION QUIET)
endif()    
```

cmake/QtFunctions.cmake
```diff
@@ -20,7 +20,9 @@ macro(find_qt_package)
...
 
-    find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} QUIET)
+    if(NOT QT_DEFAULT_MAJOR_VERSION OR QT_DEFAULT_MAJOR_VERSION STREQUAL "6")
+        find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} QUIET)
```
MathiasPaulin, line 24:
```suggestion
        find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} ${QT6_OPTION})
```

MathiasPaulin: What do you think of this proposal. I can't test it

dlyr:
Now, if during Radium configuration, Qt6 is asked, and not found, and required, an error is raised. If not required and Qt6 is asked, there is no search for Qt5 ...

If client code do not specify Qt version, the version defined during radium compilation is used.

cmake/QtFunctions.cmake
```diff
@@ -20,40 +20,41 @@ macro(find_qt_package)
...
+    else()
+        set(QT_SEARCH_MODE QUIET)
+    endif()
+
+    if(QT_DEFAULT_MAJOR_VERSION STREQUAL "6")
+        find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} ${QT_SEARCH_MODE})
+    elseif(QT_DEFAULT_MAJOR_VERSION STREQUAL "5")
+        find_package(Qt5 5.15 COMPONENTS ${MY_OPTIONS_COMPONENTS} ${QT_SEARCH_MODE})
+    else() # QT_DEFAULT_MMAJOR_VERSION not set, first search 6, then 5.
+        find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS})
```
nmellado, line 34: Don't we need to have this one `QUIET` to avoid failure and allow to go to 5 ?
```suggestion
        find_package(Qt6 COMPONENTS ${MY_OPTIONS_COMPONENTS} QUIET)
```

nmellado (approved):

PR #608, opened 2020-10-13
Add a simple window and its factory to the GuiBase to ease writing client applications

MathiasPaulin:
_Check if you branch history is PR compatible_
- Your branch need to be up to date with origin/master AND to have linear history (i.e. no merge commit).
- Update your git repository `git fetch origin` if origin is this remote
- Check with the script provided in `scripts/is-history-pr-compatible.sh`
- You must use clang-format style
_These checks are enforced by github workflow actions_
_Please refer to the corresponding log in case of failure_

_UPDATE the form below to describe your PR._

* **Please check if the PR fulfills these requirements**
- [ ] The commit message follows our guidelines
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

Be aware that the PR request cannot be accepted if it doesn't pass the Continuous Integration tests.

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)

This PR brings a simpler way to developp client application. The demo `HelloRadium`ans `SimpleAnimation`where updated to use the SimpleWindow introduced here.

This PR is far from being finished and will serve as a draft and experiment environment towar simpler Radium client API.

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)

* **Other information**:

PR #974, opened 2022-07-25
Fix draw primitives

dlyr:

Bug fix, on linux, qt 6, DrawPrirmitives examples fails to have working glbinding.
It seems to be related to the missing "main window" since DrawPrimitives create its own minimap app, and use directly Viewer as main widget.
Not sure why, but I've tested to use Viewer::create, and it didn't solve the problem.
With createWindowContainer, it works (both qt5 and qt6)

DrawPrimitives examples segfault (after displaying there is no opengl available for glbinding, while qt openglcontext seems ok.

Works.
This PR also cleanup the code a little bit.

Nope.

* **Other information**:

TODO 
- [x] Test on windows (with and without modifications introduce with this PR).


nmellado (commented):
Code looks ok, but needs to be tested on all OS

examples/DrawPrimitives/main.cpp
```diff
@@ -5,53 +5,34 @@
 #include <QOpenGLContext>
 
 #include <Engine/Scene/EntityManager.hpp>
-#include <Gui/Viewer/RotateAroundCameraManipulator.hpp>
 
 #include <minimalapp.hpp>
 #include <minimalradium.hpp>
 
-int main( int argc, char* argv[] ) {
+#include <Gui/BaseApplication.hpp>
```
MathiasPaulin, line 12:
```suggestion
```

examples/DrawPrimitives/main.cpp
```diff
@@ -5,53 +5,34 @@
...
+#include <Gui/RadiumWindow/SimpleWindowFactory.hpp>
```
MathiasPaulin, line 13:
```suggestion
```

examples/DrawPrimitives/minimalapp.cpp
```diff
@@ -3,76 +3,107 @@
...
 
-MinimalApp::~MinimalApp() {
-    // need to clean up everithing before engine is cleaned up.
-    m_task_queue.reset( nullptr );
-    m_viewer.reset( nullptr );
-    m_engine->cleanup();
-    m_engine.reset( nullptr );
+    // Create a window container for the viewer.
+    // using viewer directly as main windows fail to have valid glbinding context
+    // on Debian bullseye, Qt version 6.2.4
```
MathiasPaulin, line 74: I'm not sure this is only related to Debian bullseye but rather to the Qt version (and its implementation ?). See SimpleWindow::SimpleWindow, added by the PR #608, that do the same.

examples/DrawPrimitives/minimalradium.hpp
```diff
@@ -5,9 +5,6 @@
 #include <Engine/Scene/System.hpp>
 #include <Engine/Scene/SystemDisplay.hpp>
 
-/* This file contains a minimal radium/qt application which shows the
-classic "Spinning Cube" demo. */
-
 /// This is a very basic component which holds a spinning cube.
```
MathiasPaulin, line 8: Not sure this comment is correct. There is no spinning cube in the component but all the geometric primitives Radium can handle.

examples/DrawPrimitives/minimalapp.cpp
dlyr: Well, it works on debian + qt 5, so it's related at least to qt version, and if current release-candidate works on macos, it may be related to the qt os specific implementation.

examples/DrawPrimitives/minimalapp.cpp
```diff
@@ -3,76 +3,107 @@
...
+
+    KeyMappingManager::createInstance();
+}
+
+MinimalApp::~MinimalApp() {
+    // need to clean up everithing before engine is cleaned up.
+    m_taskQueue.reset( nullptr );
+    m_viewer.reset( nullptr );
+    m_engine->cleanup();
+    m_engine.reset( nullptr );
```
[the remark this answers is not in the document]
dlyr: Well, since it's a singleton, I prefer to use raw ptr "view" to engine.

MathiasPaulin (approved):

nmellado (approved):

PR #976, opened 2022-07-25
Dynamic texture update in tasks

nmellado:

Feature

TextureManager as a method to update textures, but it is never called

Call updatePendingTextures at each frame.
Add example showing how to update textures in Tasks.

No

* **Other information**:

### DIFF
--- a/examples/TexturedQuadDynamic/main.cpp
+++ b/examples/TexturedQuadDynamic/main.cpp
@@ -1,0 +1,100 @@
+// Include Radium base application and its simple Gui
+#include <Gui/BaseApplication.hpp>
+#include <Gui/RadiumWindow/SimpleWindowFactory.hpp>
+
+// include the Engine/entity/component interface
+#include <Core/Asset/BlinnPhongMaterialData.hpp>
+#include <Core/Geometry/MeshPrimitives.hpp>
+#include <Core/Tasks/Task.hpp>
+#include <Engine/Data/TextureManager.hpp>
+#include <Engine/FrameInfo.hpp>
+#include <Engine/Scene/EntityManager.hpp>
+#include <Engine/Scene/GeometryComponent.hpp>
+#include <Engine/Scene/GeometrySystem.hpp>
+
+#include <QTimer>
+
+/// Compute a texture wrt to a given timestamp
+class DynamicTextureSystem : public Ra::Engine::Scene::System
+{
+    unsigned char data[192 * 512];
+    Ra::Engine::Data::TextureManager* m_texManager { nullptr };
+    /// Name of the texture, used as identifier by the TextureManager
+    const std::string textureName { "myTexture" };
+    /// Copy of the texture parameters, used by the TextureManager to identify the pending texture
+    Ra::Engine::Data::TextureParameters textureParam;
+
+    inline void computeTexture( Scalar time ) {
+        // fill with some function
+        for ( int i = 0; i < 192; ++i ) {
+            for ( int j = 0; j < 512; j++ ) {
+                data[( i * 512 + j )] =
+                    (unsigned char)( 255.0 * std::abs( std::sin( j * i * M_PI / 64.0 + time ) *
+                                                       std::cos( j * i * M_PI / 96.0 + time ) ) );
+            }
+        }
+    }
+
+  public:
+    inline DynamicTextureSystem( Ra::Engine::Data::TextureManager* mng ) : m_texManager( mng ) {
+        assert( m_texManager != nullptr );
+
+        /// Create texture
+        this->computeTexture( 0 );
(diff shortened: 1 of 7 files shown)


examples/TexturedQuadDynamic/main.cpp
```diff
@@ -0,0 +1,99 @@
...
+    inline DynamicTextureSystem( Ra::Engine::Data::TextureManager* mng ) : m_texManager( mng ) {
+        assert( m_texManager != nullptr );
+
+        /// Create texture
+        this->computeTexture( 0 );
+        auto& myTexture = m_texManager->addTexture( textureName, 192, 512, data );
+        // these values will be used when engine initialize texture GL representation.
+        myTexture.format         = gl::GLenum::GL_RED;
+        myTexture.internalFormat = gl::GLenum::GL_R8;
+        m_texManager->loadTexture( myTexture, false );
```
hiergaut, line 45: Use getOrLoadTexture here

src/Gui/BaseApplication.cpp
```diff
@@ -498,6 +498,9 @@ void BaseApplication::radiumFrame() {
 
     timerData.tasksEnd = Core::Utils::Clock::now();
 
+    // Update texture (cpu->gpu)
+    m_engine->getTextureManager()->updatePendingTextures();
```
hiergaut, line 502: Use this line after startRendering call, your BaseApplication works for my application, but not for your example

examples/TexturedQuadDynamic/main.cpp
nmellado: If I do so, the texture is not accessible in the shader anymore (quad is black)

nmellado: And if I use `LoadTexture` I crash at the CORE_ASSERT when calling `updateTextureContent` in the task.

dlyr:
I have a working solution, but I don't get why image2d do not update actual displayed texture

nmellado:
> I have a working solution, but I don't get why image2d do not update actual displayed texture

Sorry I don't understand what you mean.
The function used to compute the image takes the frame time stamp, but do not use it yet. So the image is not expected to change over time (with this version, should be updated on the issue is fixed)

dlyr:
Yes I've used timestamp, and track that the data is well modified during Texture::updateData, but the final render do not change (I've also check the texture id is coherent between task manager and blinn phong material)

nmellado:
@hiergaut is going to push a fix, let see if it looks like yours
thx guys

dlyr:
My version is working, but also diverge from current.
Main problem was a missing viewer->makeCurrent before updatePendingTextures

examples/TexturedQuad/main.cpp
```diff
@@ -23,35 +25,106 @@ int main( int argc, char* argv[] ) {
...
     //! [Tell the window that something is to be displayed]
 
+    auto viewer = app.m_mainWindow->getViewer();
+    viewer->makeCurrent();
+    auto texture = app.m_engine->getTextureManager()->getOrLoadTexture( textureParameters );
+    viewer->doneCurrent();
+
+    // terminate the app after 4 second (approximatively). Camera can be moved using mouse moves.
+    // use different thread as render thread
+    auto thread = std::thread( [&app]() {
```
hiergaut, line 75:
```suggestion
    auto thread = std::thread( [=,&app]() {
```

src/Engine/Data/TextureManager.cpp
```diff
@@ -141,17 +141,18 @@ void TextureManager::deleteTexture( Texture* texture ) {
...
 }
 
 void TextureManager::updatePendingTextures() {
     if ( m_pendingData.empty() ) { return; }
 
     for ( auto& data : m_pendingData ) {
-        LOG( logINFO ) << "TextureManager::updateTextures \"" << data.first << "\".";
-        m_textures[data.first]->updateData( data.second );
+        auto itr = m_textures.find( data.first );
+        if ( itr != m_textures.end() ) { itr->second->updateData( data.second ); }
```
hiergaut, line 152: I don't know if it's a good idea to make the function updatePendingTextures more complex (adding a condition) here because this function with the contribution of this PR will be called for each rendered frame. Would it be possible that the Texture::updateData function can initialize a texture with his textureParameters if it is not already initialized before affecting the image data without overcost ?

dlyr: Well, in any case you have to test for something (the test will be done only if pending texture is not empty, do every frame, only for updated texture ...)
Another way to do it is to unsure that updateTextureData is called after the texture is initialized.

hiergaut:
Please close this PR as said here #982.

PR #980, opened 2022-07-26
Fix issue #979

### LINKED ISSUE #979: Missing guards in HeadlessExample
When glfw3 is not installed, the example HeadlessExample is still included in the cmake build chain, causing cmake to fail at configure time (headless component is not found).

MathiasPaulin:
closes #979 
 
Add the missing guard in integration test when not all components are configured.
See #979


nmellado (commented):
Sounds good.
@dlyr does it solve the problem with `examples` on your side ?

dlyr:
Yep.
I'm a bit confused by changing RADIUM_GENERATE_LIB things, I'll rather add a line under actual configured compoents, stating explicitly which requested component are not build (something like `-- missing components: Headless (see log for details)`

MathiasPaulin:
We can do without modifying the  RADIUM_GENERATE_LIB (with a bit more logic) but then the final status message might show some components as configured even if they are not. For me, RADIUM_GENERATE_LIB serves as a hint for the user to set the wanted components and after configuration should reflect the actual configured ones. So, I've no concerns in modifying this variable and use it to test which integration test to use.
For examples configuration, the find_package tricks is still needed as they can be configured outside of the build tree.

CMakeLists.txt
```diff
@@ -208,7 +209,8 @@ if(RADIUM_ENABLE_TESTING)
     # prepare integration tests
     if(NOT RADIUM_ENABLE_EXAMPLES)
         add_subdirectory(examples/CoreExample EXCLUDE_FROM_ALL)
-        if(RADIUM_GENERATE_LIB_HEADLESS)
+        find_package(Radium COMPONENTS Headless QUIET)
```
MathiasPaulin, line 209: Why doing a `find_package` here as we have means to test for the presence of Headless component without side effect.

dlyr: well, it's another proposal, I've kept most of you proposal in //
My proposal is not to change RADIUM_GENERATE_LIB_HEADLESS ... but I let you decide what you think is best.
If it's to change RADIUM_GENERATE_LIB_HEADLESS, then you can also remove RADIUM_COMPONENTS related stuff, since they are redudant

MathiasPaulin: I dont really care, simple difference of opinion, except for the side effect I removed in the next commit. I also make final message less disturbing printing missing components only if they are some.

dlyr (approved):

src/CMakeLists.txt
```diff
@@ -46,6 +47,7 @@ endif()
 
 add_dependencies(RadiumLibs ${RADIUM_COMPONENTS})
 set(RADIUM_COMPONENTS ${RADIUM_COMPONENTS} PARENT_SCOPE)
+set(RADIUM_MISSING_COMPONENTS ${RADIUM_MISSING_COMPONENTS} PARENT_SCOPE)
```
nmellado, line 50: Why is the variable `RADIUM_MISSING_COMPONENTS` set twice in the same file ?

dlyr: Right, it's uneeded.

nmellado (approved):

PR #981, opened 2022-07-27
Cmake lint

### LINKED ISSUE #962: cmake style
Currently there is a cmake-lint configuration describe here https://github.com/STORM-IRIT/Radium-Engine/blob/master/.cmake-format It defines naming convention, which are not followed. Here are the rules
```
  function_pattern = '[0-9a-z_]+'
  macro_pattern = '[0-9a-z_]+'
  global_var_pattern = '[A-Z][0-9A-Z_]+'
  internal_var_pattern = '_[A-Z][0-9A-Z_]+'
  local_var_pattern = '[a-z][a-z0-9_]+'
  private_var_pattern = '_[0-9a-z_]+'
  public_var_pattern = '[A-Z][0-9A-Z_]+'
  argument_var_pattern = '[a-z][a-z0-9_]+'
  keyword_pattern = '[A-Z][0-9A-Z_]+'
```
It barely corresponds to cmake-format default, except for macro pattern, that is set to match default function pattern.

PR #958 propose to add cmake-lint as pre-commit, but it can be removed from the PR and added later, with style fix.
Before that, we can discuss/validate these naming convention.
(to skip one pre-commit check simpy use `SKIP=name-of-the-check git commit -m"message"` see https://pre-commit.com/#temporarily-disabling-hooks)

dlyr:
Apply cmake-lint to all cmake code.
WIP (see pre-commit report for todo modifications)
To check that everything looks fine.
Fix #962


dlyr:
For the review I propose to comment about specific line directly in code, and to the general style configuration in the linked issue.

nmellado:
Did you see that the coverage call seems to be broken ?

dlyr:
Right, might be fixed (it worked before, because of name clash. i.e. marco parameters where not substitued by they actual value, but since their names refer global variables it worked.)

dlyr:
After continuing this PR, I found cmake-lint to restrictive.
Either we set naming convention quite open, or we try to do our best "by hand".
Or we remove some files from the checks.

PR #982, opened 2022-07-27
Update texture data

hiergaut:

Added feature of updating texture data (cpu) for any thread and auto update the (gpu) data during the rendering loop.

No texture update is possible from a different thread as rendered thread.

You can update texture everywhere in your code.

updateData(const void*data) -> updateData(void *data)
before : The function updateData performed the updateGL cpu -> gpu with const user pointer. (not internal data, textureParameter.texels).
now : The function updateData performs the update of internal data pointer (only cpu).
The update of gpu representation will be done later during the rendering loop.

* **Other information**:


examples/TexturedQuad/main.cpp
```diff
@@ -23,35 +24,71 @@ int main( int argc, char* argv[] ) {
...
     app.m_mainWindow->prepareDisplay();
     //! [Tell the window that something is to be displayed]
 
+    auto viewer = app.m_mainWindow->getViewer();
+    viewer->makeCurrent();
+    auto texture = app.m_engine->getTextureManager()->getOrLoadTexture( textureParameters );
+    viewer->doneCurrent();
+
+    // terminate the app after 4 second (approximatively). Camera can be moved using mouse moves.
+    auto thread = std::thread( [&app, &texture]() { // not worked
```
hiergaut, line 72:
```suggestion
    auto thread = std::thread( [=, &app, &texture]() {
```

src/Engine/Data/DirtyableObject.hpp
```diff
@@ -0,0 +1,49 @@
...
+/// render, getNumFaces, getNumeVertices don't make sense for a texture.
+/// That's why I implemented a separate class.
+**/
+class Dirtyable
+{
+  public:
+    virtual ~Dirtyable() = default;
+
+    // implement this function updating data from model to view (cpu -> gpu).
+    virtual void clean() = 0;
```
MathiasPaulin, line 40: This correspond to the updateGL() method on already existing "dirtyable" objects.
As part of the renaming proposed in the overall comment on this PR, this should be renamed to `synchronize` or, at least, `updateGL`.

src/Engine/Data/DirtyableObject.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+    // call this method on client side when the model data is updated.
```
MathiasPaulin, line 42: use doxygen comments /// \brief .....

MathiasPaulin (commented):
I think this PR takes the good way toward asynchronous, non intrusive, quite efficient management of the coherence between CPU and GPU data.
As it is supposed to be replaced by this proposal, I propose that you close PR #976 and concentrate your effort on this.

The main concept exposed by this PR, DirtyableObject and WashingMachine is very closed to what I expected and the generalization of its usage to other objects of Radium (AttribArrayDisplayable, shaders, ...) might be done once this PR is merged. But there is still some work to do before that.

First, think about renaming Dirtyable and WashingMachine to names more related to what they do  in the radium universe.
Without definitive conviction, I propose, e.g.
 - Synchronizable (instead of Dirtyable) as objects implementing this interface might have several representations that need to be synchronized, here CPU data <-> GPU data,
 - Synchronizer (instead of WashingMachine) as the role of this class is manage the synchronization of synchronizable objects.

The proposed synchronizer (WashingMachine) is implemented as a singleton (with some concerns, e.g. the data - s_dirts and s_mtx - are outside of the scope of the singleton), using an implementation of this pattern that we do not use on Radium. There are also several singleton implementation, either using the RA_SINGLETON_INTERFACE or namespace functions as for EngineRenderTechniques and EngineMaterialConverters. I propose you implement it using namespace functions and data instead of static object.

Your functions and classes should be defined so that they export their symbol in the windows dll. So, use the RA_ENGINE_API macro as expected to define exported symbols.

I also put comments directly in the file reviews.

hiergaut:
Thanks for your feedback, I'll implement the mentioned points probably in the evening.

src/Engine/Data/SynchronizableObject.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Ra {
+namespace Engine {
+namespace Data {
+
+///////////////////////////////////////////////
+////            Synchronizer                ///
+///////////////////////////////////////////////
+
+namespace Synchronizer {
```
nmellado, line 16: Please add documentation explaining the purpose of this namespace

src/Engine/Data/SynchronizableObject.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+// Use set of object pointer to ensure there is only one reference/update
+// for each synchronizable object even if it added twice for the current frame.
+// Each synchronizable object is synchronized at most once per frame with the latest available data.
+static std::set<Synchronizable*> s_synchronizables;
+static std::mutex s_mtx;
+
+bool registerSynchronizable( Synchronizable* synchronizable ) {
```
nmellado, line 24: Add documentation

src/Engine/Data/SynchronizableObject.cpp
```diff
@@ -0,0 +1,73 @@
...
+    // cannot add synchronizable object if the synchronizer is running
+    s_mtx.lock();
+    s_synchronizables.insert( synchronizable );
+    //    LOG( Core::Utils::logINFO ) << "Synchronizer::registerSynchronizable(" << synchronizable
+    //    << ")";
+    s_mtx.unlock();
+    return true;
+}
+
+void start() {
```
nmellado, line 34: Add documentation

src/Engine/Data/SynchronizableObject.cpp
```diff
@@ -0,0 +1,73 @@
...
+Synchronizable::~Synchronizable() {
+    Synchronizer::s_mtx.lock();
+    Synchronizer::s_synchronizables.erase( this );
+    Synchronizer::s_mtx.unlock();
+    CORE_ASSERT( Synchronizer::s_synchronizables.find( this ) ==
+                     Synchronizer::s_synchronizables.end(),
+                 "Cannot synchronize dead ptr" );
+}
+
+void Synchronizable::needSync() {
```
nmellado, line 67: Add documentation

src/Engine/Data/SynchronizableObject.hpp
```diff
@@ -0,0 +1,55 @@
...
+RA_ENGINE_API void start();
+
+} // namespace Synchronizer
+
+///////////////////////////////////////////////
+////           Synchronizable               ///
+///////////////////////////////////////////////
+
+/**
+/// \brief The Synchronizable class
```
nmellado, line 26: Write a brief that explain what the class does

src/Engine/Data/Texture.cpp
```diff
@@ -94,7 +95,83 @@ void Texture::bindImageTexture( int unit,
         uint( unit ), level, layered, layer, access, m_textureParameters.internalFormat );
 }
 
-void Texture::updateData( const void* data ) {
+void Texture::updateData( void* newData ) {
```
nmellado, line 98: Why removing the const qualifier ?

nmellado (commented):
Thanks for the proposal (a few comments after a quick read)

As for the other PR, could you keep the TextureQuad example as simple as possible, and add another example showing how to use dynamic textures ?

src/Engine/Data/Texture.cpp
hiergaut: Because textureParameters.textels is void*, and could be update (no const).
There is a comment (Texture.hpp:79) says that the pointer is dangling and that it is up to the user to maintain the data. Risk of simultaneous reading and writing if the floating pointer is updated by the user at the same time as the updateGL. Should we improve this by adding an updateData function that copies the user data internally for future synchronization without simultaneous read/write in the image buffer ?

hiergaut: For information textureParameters.texels is now the internal representation of pixel data of Texture. Before this contribution updateData perform the gpu synchronization on the fly with random pointer and not the internal pointer.

hiergaut:
Thanks for your feedback, I have added the missing comments now.

src/Engine/Data/Texture.cpp
hiergaut: The updateData only update internal data. No updateGL is performed here within this function.

src/Engine/filelist.cmake
```diff
@@ -72,13 +73,15 @@ set(engine_headers
...
     Data/SimpleMaterial.hpp
+    Data/SynchronizableObject.hpp
     Data/Texture.hpp
     Data/TextureManager.hpp
     Data/ViewingParameters.hpp
     Data/VolumeObject.hpp
     Data/VolumetricMaterial.hpp
     FrameInfo.hpp
     OpenGL.hpp
+    pch.hpp
```
MathiasPaulin, line 84: This file should not be mentioned here. If `RADIUM_ENABLE_PCH` option is `OFF`, this file will not be parsed/compiled.

dlyr: Might come from generateFilelsitForModule.sh, PR #988 proposes a fix.

nmellado:
@hiergaut Can you please add tests to this PR, to cover the new classes added (can be inspired from the example).
Current coverage change is -2%.

hiergaut:
I have a serious problem, I can not rebase, there is a conflict that I do not understand the origin. I only added a test, I do not understand why git is so angry with me. Could someone help me to solve this problem together to solve this problem alone in the future ?

dlyr:
New proposal using taskQueue to handle gpuTasks, by the engine.
This new proposal misses some doc and polish.
If ok with the idea, and after accepting this PR, some work can be done to move other gpu sync task to the engine task queue.

src/Engine/Data/Texture.cpp
```diff
@@ -97,7 +100,17 @@ void Texture::bindImageTexture( int unit,
 
 void Texture::updateData( void* newData ) {
     m_textureParameters.texels = newData;
-    needSync();
+
+    // register gpu task to update opengl representation before next rendering
+    static Core::TaskQueue::TaskId updateDataTaskId;
```
MathiasPaulin, line 105: should this be in critical section as several threads might call updateData concurrently ?

MathiasPaulin: I vote for re-entrant as the caller can be unaware of the existence of other threads nor how they modified data (e.g. Plugins)

hiergaut:
Please, can you make both examples ("TexturedQuad", "TexturedQuadDynamic") work without having a segmentation fault during the registerTask with the new implemented method (TaskQueue) ? 
I would like to do some comparison tests with the previous implem.
The "TexturedQuadDynamic" example being very representative of what I can have in my application (raw data stream on different threads).

src/Engine/Data/Texture.cpp
dlyr: (btw static here is a bad idea, taskid should not be shared between instances of texture, I think this solve the problem)

dlyr:
@hiergaut my bad I forgot one file in my last commit, now should be ok, if not please report where the segfault occurs.

src/Engine/Data/Texture.cpp
```diff
@@ -94,7 +98,92 @@ void Texture::bindImageTexture( int unit,
...
+
+    // register gpu task to update opengl representation before next rendering
+    if ( m_updateDataTaskId.isInvalid() ) {
+        auto taskFunc = [this]() {
+            this->updateGL();
+            m_updateDataTaskId = Core::TaskQueue::TaskId::Invalid();
+        };
+        auto task          = std::make_unique<Core::FunctionTask>( taskFunc, getName() );
+        m_updateDataTaskId = RadiumEngine::getInstance()->addGpuTask( std::move( task ) );
+    }
```
MathiasPaulin, line 112: I still think this section should be a critical one. Two or more threads may ask for texture update simultaneously and the result should be unpredictable.
I propose (but have not tested nor verified) to have an instance member (a mutex) m_updateMutex and do the following to make any test_and_write on m_updateDataTaskId a critical section

```suggestion
    m_updateMutex.lock();
    if ( m_updateDataTaskId.isInvalid() ) {
        auto taskFunc = [this]() {
            this->updateGL();
            m_updateMutex.lock();
            m_updateDataTaskId = Core::TaskQueue::TaskId::Invalid();
            m_updateMutex.unlock();
        };
        auto task          = std::make_unique<Core::FunctionTask>( taskFunc, getName() );
        m_updateDataTaskId = RadiumEngine::getInstance()->addGpuTask( std::move( task ) );
    }
    m_updateMutex.unlock();
```

dlyr: Fine for me, I'll do it (with lock_guard). 
If I understand well, same shoud be done for setParameters, but I have a concern with non const ref acess with getParameters, since use code from different threads might gain write acess, so we may give access to the mutex (or to a lock) as well so that use code can lock the resource.

nmellado:
@hiergaut are you fine with this proposal ? Can you confirm the final version works for you ?

MathiasPaulin (approved):
As discussed AFK, this is only a step toward a better management of CPU<->GPU data synchronization. But right now, and before going further, this PR could be merged if it works for the external application that needs it.

hiergaut:
It works on my side
@nmellado, sorry for the late reply, I didn't remember receiving a notification email for your request, or I deleted the email by mistake.

PR #988, opened 2022-08-23
Update filelist script to skip pch.hpp

dlyr:

Remove pch.hpp from generate file list

pch.hpp is included in cmake target's header, which is not needed.

pch.hpp is removed from autogenerated header list.


.github/workflows/guideline-checks.yml
```diff
@@ -30,3 +30,18 @@ jobs:
...
+    steps:
+      - name: Checkout remote head
+        uses: actions/checkout@master
+        with:
+          fetch-depth: 0
+      - name: generate and check diff
+        run: |
+          pip3 install --upgrade pip && pip install cmake-format
+          cd scripts
+          bash ./generateAllFilelists.sh
```
MathiasPaulin, line 46: I've a question about this checking. I understand it as making the use of generateAllFilelists.sh mandatory for PR submitters. But is this script working an all supported systems (Windows, macos) with the same result ?
If yes, and even if I'm dubious about it, I may agree with this constraints. If no, I think we should not verify this.

dlyr: The idea is that anyone can edit by hand the file as long as it correspond to the generated file. So the use of the script is not mandatory.

We can perform the check, but have is as not necessary to accept a PR.

The script should work as long as  bash, cmake-lint, and other commands use in the script (e.g. sort, grep ...) are available, but I haven't tested on other OSes.

MathiasPaulin: OK, seems fine.

MathiasPaulin (approved):

PR #990, opened 2022-08-24
[scripts] remove trailing / for find commands

dlyr:
Try to fix generate filelist script on macos x

find command print a // if the argument have a trailing /

this pr should fix it.

### DIFF
--- a/scripts/generateFilelistForModule.sh
+++ b/scripts/generateFilelistForModule.sh
@@ -28,7 +28,7 @@ export LC_ALL=C
 function genList(){
     ext=$1
     suffix=$2
-    L=$(find  "../src/${BASE}/" -name "*.${ext}")
+    L=$(find  "../src/${BASE}" -name "*.${ext}")
     if [ ! -z "$L" ]
     then
         echo  "set(${LOWBASE}_${suffix}" >> "${OUTPUT}"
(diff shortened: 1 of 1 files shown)


MathiasPaulin (commented):
This proposal fixes the unwanted front `/` on the filelist but it still remains the `./generateFilelistForModule.sh: line 9: ${BASE,,}: bad substitution` error. 
This seems to be the result of a quite old version of bash (`GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)`) distributed by Apple. 

Another way (POSIX standard, working on macos) to convert BASE to LOWBASE could be (line 9 of the script)
``` 
LOWBASE=$(echo $BASE | tr '[:upper:]' '[:lower:]')
```

The actual substitution is only available for bash >= 4.0, not available on macos as /bin/bash.

dlyr:
On my install, when I run `bash ./generateFilelistForModule.sh Gui` it works, since it uses bash from homebrew install (version 5.x)
I'm fine with your proposal, maybe we can switch this kind of script in python or js to avoid such portability issues.

MathiasPaulin:
> On my install, when I run `bash ./generateFilelistForModule.sh Gui` it works, since it uses bash from homebrew install (version 5.x) 

Sure, it works also on mine using the macport installed version of bash (version 5.x). But neither macport nor homebrew replace /bin/bash. So, executing the script with just `./generateFilelistForModule.sh Gui` still not work.

> I'm fine with your proposal, maybe we can switch this kind of script in python or js to avoid such portability issues.

I agree with this, but I am not able to do it, my skills in python or js are way too low

MathiasPaulin (approved):
Works on Linux and macos. Not tested on Windows ;)

PR #992, opened 2022-08-26
Fix missing engine deletion in CLIVIewer that prevent OpenGL related unittest to work

MathiasPaulin:

The fix on Headless component allows to run unittests that require OpenGL context and fix a potential memory leak for some applications.
The RADIUM_ENABLE_GL_TESTING cmake option is added to enable OpenGL related unittests.

When running tests that use CLIViewer, as the engine singleton constructed when the CLIViewer is constructed is not destroyed when the CLI Viewer is destroy, the actual version segfault on the next engine singleton creation/deletion.

Engine singleton created when initializing CLIViewer is destroyed when deleting the CLIVIewer

No, only the dependent option RADIUM_ENABLE_GL_TESTING is added, OFF by default. 
OpenGL testing can be activated on any platform that allows to obtain an OpenGL context. This should be the case on your development machine. 
It is OFF by default as running applications or unittests that require OpenGL is not available on all CI platform.

* **Other information**:

Some code is commented out as it will be available only when PR #950 will be merged

### DIFF
--- a/.github/workflows/build-matrix.yml
+++ b/.github/workflows/build-matrix.yml
@@ -94,8 +94,9 @@ jobs:
           [[ "true" == "${{ inputs.qt5 }}" && "true" == "${{ inputs.qt6 }}" ]] && echo ','
           [[ "true" == "${{ inputs.qt6 }}" ]] && echo '{ "name" : "qt6", "value" : "6.2.0"}'
           echo '],'
-          echo '"coverage" :'
-          [[ "true" == "${{ inputs.coverage }}" ]] && echo '["ON"]' || echo '["OFF"]'
+          echo '"coverage" : ['
+          [[ "true" == "${{ inputs.coverage }}" ]] && echo '{ "value" : "ON", "extra-flags" : "-DRADIUM_ENABLE_GL_TESTING=ON" }' || echo '{ "value" : "OFF", "extra-flags" : "" }'
+          echo "]"
           echo "}"
           ) | jq -c .)
           echo $matrix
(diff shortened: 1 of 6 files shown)


dlyr (approved):
I have made a proposal to run gl unit test on ci coverage computation (linux only).

MathiasPaulin:
> I have made a proposal to run gl unit test on ci coverage computation (linux only).

fine for me, that's what I had in mind.

But I think that we could keep `target_compile_definitions(unittests PRIVATE ENABLE_GL_TESTING)` in tests/unittest/CMakeLists.txt if GL_TESTING is ON so that for some unit tests that contains non-opengl code AND openGL code (ex. we can test textures with or without openGL), there will be no need to write two source files.

dlyr:
Well, I'm not sure for the define.
What I had in mind is to split unit test with and without GL to have two executable, and add in cmake the launcher for gl unit test.
But it seems to be unsupported by catch (and cmake auto discover catch test) so I keep only one catch executable.

Having two kinds of test in the same source file is not a problem on its own, but if one use this define inside a given test, the test should pass without GL, while not pass with GL, and it can be quite confusing on unit test report.

e.g. 
```{.cpp}
#ifdef ENABLE_GL_TESTING
TEST_CASE( "Engine/Data/TOTO/WIthGL",
           "[Engine][TOTO]" ) {
// gl test
}
#endif
TEST_CASE( "Engine/Data/TOTO/WIthoutGL",
           "[Engine][TOTO] ) {
// non gl test
}
```
is fine while 
```{.cpp}
TEST_CASE( "Engine/Data/TOTO",
           "[Engine][TOTO]" ) {
#ifdef ENABLE_GL_TESTING
// gl test
#endif
// non gl test
}
```
is not

So removing the compile definition prevent from bad usage by design.

MathiasPaulin:
This makes sense. 
So, OK to merge this PR.

PR #1004, opened 2022-09-15
Custom types attrib manager

nmellado:
# Pull Request Desription

## Check if you branch history is PR compatible

  - [ ] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] other:

- **Other information**:


dlyr:
Custom attrib needs to implement AttribBase interface, with container introspection.
Clearly doc is missing, but one has to provide either a full CustomAttrib class implementation, or provides templated specialisation for the custom type, as done for float and double:
 ```

template <>
size_t Attrib<double>::getNumberOfComponents() const {
    return 1;
}
```

**Comment by nmellado** [comment]
Shall we provide a default implementation of `VectorArrayTypeHelper` for other types ?
I mean, it could interesting to have strings as vertex attributes.

But in that case, things would be messy:
 - `VectorArrayTypeHelper::Size` should be 0 (to avoid binding with GL)
 - GetData would return a `V*`, not a `Scalar*` (to still access data). Alternatively, it could return `nullptr` (I think I prefer this option).

What do you think ?

File: src/Core/Containers/VectorArray.hpp [hunk 1]
```diff
@@ -3,21 +3,34 @@
...
+        VectorArrayTypeHelper<V, std::is_integral<V>::value || std::is_floating_point<V>::value>;
+
   public:
     // Type shortcuts
-    using Vector         = V;
-    using Matrix         = Eigen::Matrix<typename V::Scalar, V::RowsAtCompileTime, Eigen::Dynamic>;
-    using MatrixMap      = Eigen::Map<Matrix>;
-    using ConstMatrixMap = Eigen::Map<const Matrix>;
+    static constexpr int ElementSize = TypeHelper::Size;
+    using Scalar                     = typename TypeHelper::Scalar;
```
MathiasPaulin, line 29: I prefer something like `value_type` instead of `Scalar` to prevent confusion with the Radium `Scalar` type definition (i.e. `float` or `double` depending on the building options)
```suggestion
    using value_type                     = typename TypeHelper::value_type;
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,21 +3,34 @@
...
+    using Vector                     = V;
```
MathiasPaulin, line 30: if V is a scalar type (integral of floating_point), seems strange to define an alias name as `Vector`.
In the previous version, the specialization of VectorArray for scalar types removed this alias .

src/Core/Containers/VectorArray.hpp
```diff
@@ -26,68 +39,30 @@ class VectorArray : public AlignedStdVector<V>
...
-    }
-
-    /// Returns the array as an Eigen Matrix Map (const version)
-    ConstMatrixMap getMap() const {
-        CORE_ASSERT( !this->empty(), "Cannot map an empty vector " );
-        return ConstMatrixMap( this->data(), 1, Eigen::Index( this->size() ) );
-    }
+template <typename V>
+struct VectorArrayTypeHelper<V, true> {
+    using Scalar              = V;
```
MathiasPaulin, line 55:
```suggestion
    using value_type              = V;
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,21 +3,34 @@
...
+
 /// This class is a wrapper around a std::vector of Core::Vectors.
 /// which allow to use the stdlib's dynamic array implementation, yet pass it as
 /// a matrix when Eigen needs it with the getMap() method.
 template <typename V>
 class VectorArray : public AlignedStdVector<V>
 {
+  private:
+    using TypeHelper =
+        VectorArrayTypeHelper<V, std::is_integral<V>::value || std::is_floating_point<V>::value>;
```
MathiasPaulin, line 24:
```suggestion
        VectorArrayTypeHelper<V, std:: is_scalar_v<V>>;
```
(see https://en.cppreference.com/w/cpp/types/is_scalar)
This will allow to have char* attributes (eg C strings), enum values and others as attribute values.

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,21 +3,34 @@
...
+
   public:
     // Type shortcuts
-    using Vector         = V;
-    using Matrix         = Eigen::Matrix<typename V::Scalar, V::RowsAtCompileTime, Eigen::Dynamic>;
-    using MatrixMap      = Eigen::Map<Matrix>;
-    using ConstMatrixMap = Eigen::Map<const Matrix>;
+    static constexpr int ElementSize = TypeHelper::Size;
```
MathiasPaulin, line 28:
```suggestion
    static constexpr int NumberOfComponents = TypeHelper::NumberOfComponents;
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -26,68 +39,30 @@ class VectorArray : public AlignedStdVector<V>
...
-
-    /// Returns the array as an Eigen Matrix Map (const version)
-    ConstMatrixMap getMap() const {
-        CORE_ASSERT( !this->empty(), "Cannot map an empty vector " );
-        return ConstMatrixMap( this->data(), 1, Eigen::Index( this->size() ) );
-    }
+template <typename V>
+struct VectorArrayTypeHelper<V, true> {
+    using Scalar              = V;
+    static constexpr int Size = 1;
```
MathiasPaulin, line 56:
```suggestion
    static constexpr int NumberOfComponents = 1;
```

MathiasPaulin (commented):
Some elements in this review answers (at least partly) to the question about having strings (restricted to C strings) as attributes, and even function pointer (could be helpful to have lambda as vertex attributes ;) ). But then, having the attrib bindable to openGL should be done using another decision process than the number of elements that will be 1 for this type of attribs.

src/Core/Containers/VectorArray.hpp
MathiasPaulin: To prevent openGL binding for non arithmetic types using the number of compoenents, we should also have something like 
```suggestion
    static constexpr int NumberOfComponents = std::is_arithmetic_v<V> ? 1 : 0;
```
optinally, we could try to allow also some enum bindings (e.g using the enums underlying type as value_type instead of the enum type itself ...

dlyr:
At first glance, I think that VectorArray should implement the ContainerIntrospectionInterface.
Moreover VectorArray have some functionalities that seems tailored for eigen types, maybe we use the base class (AlignedStdVector) to implement container introspection which will be used for attrib and common functionalites 
Attrib then can have a type helper that provides either AlignedStdVector or VectorArray for eigen types.
All these classes can have a clone method to perform deep copy, hence attrib manager could copy all attribs without knowing the types of the attrib.

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,20 +3,33 @@
...
+
 /// This class is a wrapper around a std::vector of Core::Vectors.
 /// which allow to use the stdlib's dynamic array implementation, yet pass it as
 /// a matrix when Eigen needs it with the getMap() method.
 template <typename V>
 class VectorArray : public AlignedStdVector<V>
 {
+  private:
+    using TypeHelper =
+        VectorArrayTypeHelper<V, std::is_integral<V>::value || std::is_floating_point<V>::value>;
```
MathiasPaulin, line 24:
```suggestion
        VectorArrayTypeHelper<V, is_arithmetic_v<T>>;
```
Well, I understand your restriction to arithmetic types (integral and floating point) so that this class is compatible with Eigen functionalities that do not work if we store pointers in this array. Am I right ?

But, in this case, why not changing also the class name (its VectorArray and is not representative of scalar array if V is just an arithmetic  type)  and documentation as it is no more a "std::vector of Core::Vectors" as  it can store scalars (any integral type or floating point type) as well  ? 
Why can't we have any scalar type (using the proposed std::is_scalar) as well as Eigen Vectors ? Some eigen functionalities could be disabled for non eigen compatible type as you did for the alias scalar_type.

MathiasPaulin:
> At first glance, I think that VectorArray should implement the ContainerIntrospectionInterface.
> Moreover VectorArray have some functionalities that seems tailored for eigen types, maybe we use the base class (AlignedStdVector) to implement container introspection which will be used for attrib and common functionalites
> Attrib then can have a type helper that provides either AlignedStdVector or VectorArray for eigen types.
> All these classes can have a clone method to perform deep copy, hence attrib manager could copy all attribs without knowing the types of the attrib.

Well, I agree with your proposal except that I think this class should no more be called `VectorArray`.
And `AlignedStdVectors` are just a specialisation of std::vectors with the `Eigen::aligned_allocator<T>>::vector` . What happens if we store something else than a vector in an AlignedStdVector ? Are the data well aligned ? is there some more "padding" in the data ?
So yes, we can think about a custom container class implementing the ContainerIntrospectionInterface and adding on top of that the eigein constraints. But, reading twice the code, why are we doing this as there are already some aliases of VectorArray<T> for int, and some others (e.g. `DEFINE_CONVENIENCE_MATRIX_ALIAS( Vector1iArray, int, 1 )` at line 101 of VectorArray.hpp ...

Contrary of what I  said some hours ago, I'm now not sure the objectives of this PR are clear and that the proposal is the good direction to take. We need to discusss this clearly at our next AFK meeting.

dlyr:
Concerning DEFINE_CONVENIENCE_MATRIX_ALIAS
Vector1iArray is a vector of Eigen::Matrix<int 1, 1> which is different from raw int.

For the naming, I propose we first work on the functionalities and behavior of the code, and then we can think of renaming if needed.

I'm not sure to catch "if we store something else than a vector in an AlignedStdVector", to what "vector" refer here ?
AlignedStdVector is a std::vector that uses aligned_allocator, it's suitable to store anything.
It ensure certain alignment of the first address of allocated blocks.

More about aligned allocator, see eigen doc
https://eigen.tuxfamily.org/dox/classEigen_1_1aligned__allocator.html
https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html

And as far as I understand, the goal of the PR is to extends possible type for attrib, which are currently restricted to float, double, or any Eigen::Matrix types.

MathiasPaulin:
I think that the naming (and doc) should  be changed as the objective of this class will change. But, yes, this should be done when functionalities will be clear and well implemented. 

If the goal of the PR is to extend possible types for attrib (and i understood this), I reopen the discussion which propose to enable at once any types for which std::is_scalar_v<T> is true (or, longer to write  std::is_scalar<T>::value). 

AS this PR is work in progress, I add the WIP tag.

src/Core/Containers/VectorArray.hpp
nmellado: We can replace `is_arithmetic_v` by `is_scalar`.
Regarding the name: `VectorArray` was already supporting `float` and `double`: this PR only add other types, nothing else. 

We might to want to change the name of this class and refactor its use in `Attrib`, however I believe this is out of the scope of this class.

dlyr:
With 98430e9 we can have any type in VectorArray.
NumberOfComponents are 1 for arithmetic types, N for fixed N dimensional Vector (any type), -1 for dymanic vector, 0 for other types (e.g. std::string, enum) 

```
Vector1Array : arithmetic(1), eigen(0), #components 1
Vector1uArray : arithmetic(0), eigen(1), #components 1
Vector2Array : arithmetic(0), eigen(1), #components 2
VectorArray<std::string> : arithmetic(0), eigen(0), #components 0
VectorArray<Ra::Core::VectorN> : arithmetic(0), eigen(1), #components -1
VectorArray<enumClass> : arithmetic(0), eigen(0), #components 0
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,20 +3,35 @@
...
 #include <Core/RaCore.hpp>
 
+#include <type_traits> // std::is_integral
+
 namespace Ra {
 namespace Core {
 
+/// Internal class used to compute types and values used by VectorArray, to handle both
+/// eigen matrices and integral types
+template <typename V, bool isIntegral, bool isEigen>
```
nmellado, line 13:
```suggestion
template <typename V, bool isArithmetic, bool isEigen>
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -26,68 +41,42 @@ class VectorArray : public AlignedStdVector<V>
     /// Returns the array as an Eigen Matrix Map
     MatrixMap getMap() {
         CORE_ASSERT( !this->empty(), "Cannot map an empty vector " );
-        return MatrixMap( this->data()->data(), V::RowsAtCompileTime, this->size() );
+        return MatrixMap( TypeHelper::getData( this ),
```
nmellado, line 44: Use `static_assert`to ensure that `TypeHelper::NumberOfComponents>1`

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,20 +3,35 @@
...
+
   public:
     // Type shortcuts
-    using Vector         = V;
-    using Matrix         = Eigen::Matrix<typename V::Scalar, V::RowsAtCompileTime, Eigen::Dynamic>;
-    using MatrixMap      = Eigen::Map<Matrix>;
+    static constexpr int NumberOfComponents = TypeHelper::NumberOfComponents;
+    // scalar_type is different from value_type defined in Base class
+    using scalar_type = typename TypeHelper::scalar_type;
+    using Matrix      = Eigen::Matrix<scalar_type, TypeHelper::NumberOfComponents, Eigen::Dynamic>;
```
MathiasPaulin, line 33: Is there a mean to disable this declaration (and others) for non eigen type ? using std::enable°if or something like this ? I'm not sure (and do not success in my writing tests but ...)

src/Core/Containers/VectorArray.hpp
```diff
@@ -3,20 +3,35 @@
...
 /// This class is a wrapper around a std::vector of Core::Vectors.
 /// which allow to use the stdlib's dynamic array implementation, yet pass it as
 /// a matrix when Eigen needs it with the getMap() method.
 template <typename V>
 class VectorArray : public AlignedStdVector<V>
 {
+  private:
+    using TypeHelper =
+        VectorArrayTypeHelper<V,
+                              std::is_arithmetic<V>::value,
```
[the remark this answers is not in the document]
dlyr: As this is only cosmetic, and not about functionality, we can keep c++14 only version, better for some platform.

dlyr:
TODO;
- [ ] add tests to cover new lines
- [ ] documentation check 
- [ ] static_assert or enable_if for getMap (@dlyr)

src/Core/Containers/VectorArray.hpp
```diff
@@ -2,92 +2,93 @@
...
+/// This class is implements ContainerIntrospectionInterface for AlignedStdVector.
+/// It provides Eigen::Map functionality if the underlying component allows it (i.e. fixed size)
 template <typename V>
-class VectorArray : public AlignedStdVector<V>
+class VectorArray : public AlignedStdVector<V>, public Utils::ContainerIntrospectionInterface
 {
+  private:
+    using TypeHelper =
+        VectorArrayTypeHelper<V,
+                              std::is_arithmetic<V>::value || std::is_enum<V>::value,
```
MathiasPaulin, line 25: why not using `std::is_scalar<V>::value` that will also allow pointers as `component_type` ?

src/Core/Containers/VectorArray.hpp
```diff
@@ -2,92 +2,93 @@
...
+                              std::is_base_of<typename Eigen::MatrixBase<V>, V>::value>;
+
   public:
     // Type shortcuts
-    using Vector         = V;
-    using Matrix         = Eigen::Matrix<typename V::Scalar, V::RowsAtCompileTime, Eigen::Dynamic>;
-    using MatrixMap      = Eigen::Map<Matrix>;
+    static constexpr int NumberOfComponents = TypeHelper::NumberOfComponents;
+    // compoenent_type is different from value_type defined in Base class, e.g. for
```
MathiasPaulin, line 31:
```suggestion
    // component_type is different from value_type defined in Base class, e.g. for
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -2,92 +2,93 @@
...
-    using MatrixMap      = Eigen::Map<Matrix>;
-    using ConstMatrixMap = Eigen::Map<const Matrix>;
-
-  public:
-    /// Inheriting constructors from std::vector
-    using AlignedStdVector<double>::AlignedStdVector;
+template <typename V>
+struct VectorArrayTypeHelper<V, true, false> {
+    using component_type                    = V;
+    static constexpr int NumberOfComponents = 1;
```
MathiasPaulin, line 71: if using `std::is_scalar<V>::value` above, should use `(std::is_arithmetic<T>::value || std::is_enum<T>::value) ? 1 : 0` here

src/Core/Containers/VectorArray.hpp
dlyr: done with enabe_if

src/Core/Containers/VectorArray.hpp
dlyr: Matrix typedef is valid for non eigen types, as long as you canmap data to eigen matrix. Check via getMatrix availability

dlyr (approved):

src/Core/Containers/VectorArray.hpp
```diff
@@ -2,91 +2,116 @@
...
+ */
 template <typename V>
-class VectorArray : public AlignedStdVector<V>
+struct VectorArrayTypeHelper : public VectorArrayTypeHelperInternal<
+                                   V,
+                                   std::is_arithmetic<V>::value || std::is_enum<V>::value,
+                                   std::is_base_of<typename Eigen::MatrixBase<V>, V>::value> {};
+
+/**
+ * @brief This class is implements ContainerIntrospectionInterface for AlignedStdVector.
```
nmellado, line 29:
```suggestion
 * @brief This class implements ContainerIntrospectionInterface for AlignedStdVector.
```

src/Core/Containers/VectorArray.hpp
```diff
@@ -2,91 +2,116 @@
...
-/// which allow to use the stdlib's dynamic array implementation, yet pass it as
-/// a matrix when Eigen needs it with the getMap() method.
+/** Internal class used to compute types and values used by VectorArray, to handle both
+ * eigen fixed sized matrix and scalar type as mappable to eigen matrices.
+ */
+template <typename V, bool isArithmetic, bool isEigen>
+struct VectorArrayTypeHelperInternal {};
+
+/**
+ * arithmetic and fixed sized eigen types are mappable to Eigen matrices.
```
nmellado, line 19:
```suggestion
 * Arithmetic and fixed sized eigen types are mappable to Eigen matrices.
```

MathiasPaulin (approved):
OK for me as soon as the last discussion will be closed and their typo-fix proposal committed.

PR #1007, opened 2022-09-29
Radium dataflow

MathiasPaulin:
# Pull Request Desription
Add dataflow programming capabilities to the Radium engine (draft)

TODO
 - [ ] clarify ownership transfer of the nodes to the graph 
 - [ ] clearly state that naked pointers got from the graph are non owning pointers.
 - [ ] add rendering component (NOT IN THIS PR)
 - [x] add documentation and programming guide (WIP)
 - [x] transform demos to unittests

Capture d’écran 2022-10-11 à 20 53 40


external/Core/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_custom_target(CoreExternals ALL)
 
 if(NOT DEFINED Eigen3_DIR)
     check_externals_prerequisite()
-    status_message("" "eigen3" "remote git")
+    status_message("[CoreExternal]" "eigen3" "remote git")
```
dlyr, line 20: I have made a proposal to have better external message in #1016, may superseed this.

src/Core/Random/RandomPointSet.hpp
```diff
@@ -0,0 +1,152 @@
...
+ */
+namespace Random {
+
+Scalar constexpr sqrtNewtonRaphsonhelper( Scalar x, Scalar curr, Scalar prev ) {
+    return curr == prev ? curr : sqrtNewtonRaphsonhelper( x, 0.5_ra * ( curr + x / curr ), curr );
+}
+Scalar constexpr ct_sqrt( Scalar x ) {
+    return sqrtNewtonRaphsonhelper( x, x, 0_ra );
+}
+
```
dlyr, line 25: Are these suppose to be used directly by client code ? if not, maybe move to .cpp
if yes, rename ct_sqrt to something more explicit, and following camel case naming convention.

dlyr (commented):
review in progress

src/Core/Random/RandomPointSet.hpp
MathiasPaulin: These are used, right now, only by the private implementation of, e.g. FibonacciSequence to initialise `static constexpr Scalar phi` in the class definition, and should be a constexpr. So this need to be defined here. But, your right, this should be renamed.

dlyr: isn't it possible to move them as static constexpr private member function of FibonacciSequence ?

MathiasPaulin: yes for this PR. But I've some other code that uses these helpers. In this PR, Random utilities are only used for unittests. I've some other code, that will be future PR that uses these class and some of them use these helpers. I propose to add them in a namespace called "internal" in the "Random" namespace.

MathiasPaulin: I can also remove the commits about Random utilities from this PR

dlyr: Maybe can be easier to review, but I'm fine with your explaination also.

MathiasPaulin: So, as soon as you finish on these 3 files (and the associated unittest), I'll push a commit taking into account our discussions. I also can show you AFK what I did right now.

doc/concepts/nodesystem.md
```diff
@@ -0,0 +1,199 @@
+\page nodeSystem Radium node system
+[TOC]
+
+# Radium node system
+
+Radium-Engine embed a node system allowing to develop computation graph using an adaptation of dataflow programming.
+This documentation explain the concepts used in the node system and how to develop computation graph using the Core
+node system and how to extend the Core node system to be used in specific Radium-Engine application or library.
+
+## Structure and usage of the Radium::Dataflow component
```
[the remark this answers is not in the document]
MathiasPaulin: I don't think so. This section is about usage of a specific component, not about compiling Radium nor general usage of Radium.
The first paragraph is just a reminder on how components are compiled into Radium and how their integration can be configured. 
The following is just indications on how to uses this component or only one of its sub-component.
So, this section is too specific to be moved in general Radium compilation and usage description.

examples/DataflowExamples/HelloGraph/main.cpp
```diff
@@ -0,0 +1,146 @@
...
+ * \brief Demonstrate how to use a graph to filter a collection
+ */
+int main( int argc, char* argv[] ) {
+    using RaVector = Ra::Core::VectorArray<Scalar>;
+    //! [Creating an empty graph]
+    DataflowGraph g { "helloGraph" };
+    //! [Creating an empty graph]
+
+    //! [Creating Nodes]
+    auto sourceNode    = new Sources::SingleDataSourceNode<RaVector>( "Source" );
```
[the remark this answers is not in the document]
MathiasPaulin: Nodes are own by the graph. 
The documentation should be clearer about this. 
There is an ownership transfer from the caller to the graph when nodes are added to a graph, and only if they are added. 
So, if we move a unique_ptr into the graph but the node could'nt be added, what happens on the unique_ptr ? 
We can add a function to test if a node can be added to a graph and then use this result as requirements of the addnode method. In this case, we can move a unique_ptr and return a optional with the naked pointer to the node.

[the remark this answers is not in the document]
MathiasPaulin: Node can't be added only if thereis a name conflict as node names should be unique in a graph in order to set links. So, in this exemple, everything goes fine. If a node could'nt be added, the user can test the return of add_node, change the name and add the node again. The raw pointer given to the graph is used to create a unique_ptr  managed by the graph(ownership is transfered to the graph) but the user can still use the pointer to further add links, set state, etc ... 
To prevent some problems, resulting from the current implementation or from the deletion of a unique pointer not added to the graph, I'll change the add_node that will return a pair <bool, raw ptr>. If the bool is true, the node was added to the graph and the user can use the raw pointer without ownership management (the graph owns it). If the bool is false, the raw pointer will be own by the caller and it can fix the name, embed the pointer in a unique ptr and add it again.
We can also change the ways links are defined to use the uuid (which is unique by construction) of the node instead of its name, then allowing non unique names and but it is clearly less user friendly ^^ and do not prevent all collisions if aliases are still allowed. So yep, I'll go for unique_ptr transfer with a returned pair

PR #1021, opened 2022-11-13
Simple skinning

dlyr:

  - [ ] bug fix
  - [x] feature
  - [ ] docs update
  - [ ] other:

It's hard to use skinning without loaded scene

Add new setters so that one can define all skinning data progamatically, with a simple example.

- **Other information**:
Fix checkInvalidNumbers (as far as I understand), add smoothstep

### DIFF
--- a/examples/SimpleSkinning/main.cpp
+++ b/examples/SimpleSkinning/main.cpp
@@ -1,0 +1,147 @@
+#include <Core/Animation/Skeleton.hpp>
+#include <Core/Geometry/MeshPrimitives.hpp>
+#include <Core/Tasks/Task.hpp>
+#include <Core/Types.hpp>
+#include <Engine/FrameInfo.hpp>
+#include <Engine/RadiumEngine.hpp>
+#include <Engine/Scene/CameraManager.hpp>
+#include <Engine/Scene/EntityManager.hpp>
+#include <Engine/Scene/GeometryComponent.hpp>
+#include <Engine/Scene/SkeletonBasedAnimationSystem.hpp>
+#include <Engine/Scene/SkeletonComponent.hpp>
+#include <Engine/Scene/SkinningComponent.hpp>
+#include <Engine/Scene/System.hpp>
+#include <Gui/BaseApplication.hpp>
+#include <Gui/RadiumWindow/SimpleWindowFactory.hpp>
+#include <Gui/Viewer/Viewer.hpp>
+
+using namespace Ra;
+using namespace Ra::Core;
+using namespace Ra::Core::Animation;
+using namespace Ra::Core::Math;
+using namespace Ra::Engine;
+using namespace Ra::Engine::Scene;
+
+class EntityAnimationSystem : public Scene::System
+{
+  public:
+    void setSkeleton( SkeletonComponent* skel ) { m_skel = skel; }
+    void setSkin( SkinningComponent* skin ) { m_skin = skin; }
+
+    void generateTasks( TaskQueue* q, const FrameInfo& info ) override {
+
+        if ( !m_skel || !m_skin ) return;
+
+        m_t = info.m_animationTime;
+
+        auto animTaskId = q->registerTask( new Ra::Core::FunctionTask(
+            [this]() {
+                Transform transform;
+                transform =
+                    Transform( AngleAxis( cos( m_t + Math::Pi ), Vector3( 1_ra, 0_ra, 0_ra ) ) *
+                               Translation( Vector3( 0_ra, 0_ra, 5_ra ) ) );
+
+                m_skel->setBoneTransform( 2, transform, Skeleton::SpaceType::LOCAL );
+                transform = Transform( AngleAxis( cos( m_t ), Vector3( 1_ra, 0_ra, 0_ra ) ) *
+                                       Translation( Vector3( 0_ra, 0_ra, 5_ra ) ) );
+
+                m_skel->setBoneTransform( 3, transform, Skeleton::SpaceType::LOCAL );
+            },
+            "AnimTask_" + m_skin->getMeshName() ) );
+
+        auto skinFunc = std::bind( &SkinningComponent::skin, m_skin );
+        auto skinTask = new Core::FunctionTask( skinFunc, "SkinnerTask_" + m_skin->getMeshName() );
+        auto endFunc  = std::bind( &SkinningComponent::endSkinning, m_skin );
+        auto endTask = new Core::FunctionTask( endFunc, "SkinnerEndTask_" + m_skin->getMeshName() );
+
+        auto skinTaskId = q->registerTask( skinTask );
+        auto endTaskId  = q->registerTask( endTask );
+        q->addDependency( animTaskId, skinTaskId );
+        q->addDependency( skinTaskId, endTaskId );
+    }
+
+  private:
+    Scalar m_t { 0_ra };
+    SkeletonComponent* m_skel { nullptr };
+    SkinningComponent* m_skin { nullptr };
+};
+
+void setupScene( Ra::Engine::RadiumEngine* engine ) {
+    auto animationSystem = new EntityAnimationSystem;
+    engine->registerSystem( "Simple animation system", animationSystem );
+
+    auto entity = engine->getEntityManager()->createEntity( "Cylinder" );
+
+    auto cylinder =
+        Geometry::makeSharpCylinder( { 0_ra, 0_ra, 0_ra }, { 0_ra, 0_ra, 10_ra }, 2_ra, 16, 32 );
+
+    // component ownership is transfered to entity in component ctor
+    auto meshComponent =
+        new TriangleMeshComponent( "Cylinder", entity, std::move( cylinder ), nullptr );
+
+    std::map<std::string, Core::Transform> boneMatrices;
+    std::map<std::string, std::vector<std::pair<uint, Scalar>>> boneWeights;
+
+    auto skel            = Ra::Core::Animation::Skeleton();
+    auto transform       = Transform::Identity();
+    auto rootId          = skel.addRoot( transform, "root" );
+    boneMatrices["root"] = skel.getTransform( rootId, Skeleton::SpaceType::MODEL );
+
+    transform         = Transform( Translation( Vector3( 0_ra, 0_ra, 0_ra ) ) );
+    auto id1          = skel.addBone( rootId, transform, Skeleton::SpaceType::MODEL, "1" );
+    boneMatrices["1"] = skel.getTransform( id1, Skeleton::SpaceType::MODEL ).inverse();
+    std::vector<std::pair<uint, Scalar>> w;
+    auto& vertices = meshComponent->getCoreGeometry().vertices();
+
+    transform         = Transform( Translation( Vector3( 0_ra, 0_ra, 5_ra ) ) );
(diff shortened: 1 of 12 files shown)


src/Core/Geometry/MeshPrimitives.hpp
```diff
@@ -93,6 +93,13 @@ RA_CORE_API TriangleMesh makeCylinder( const Vector3& a,
                                        uint nFaces                                = 32,
                                        const Utils::optional<Utils::Color>& color = {} );
 
+RA_CORE_API TriangleMesh makeSharpCylinder( const Vector3& a,
```
MathiasPaulin, line 96: Not sure the name is adapted to what is built. The cylinder does not seems "sharp" as normals are smoothed radially. Add documentation to explain the difference between this and the existing MakeCylinder

src/Engine/Scene/SkeletonComponent.cpp
```diff
@@ -158,8 +167,11 @@ void SkeletonComponent::update( Scalar t ) {
 
     m_animationTime = m_speed * t;
     Scalar lastTime = 0;
-    for ( auto boneAnim : m_animations[m_animationID] ) {
-        lastTime = std::max( lastTime, *boneAnim.getTimes().rbegin() );
+    if ( !m_animations.empty() ) {
+        // m_animationID is always < m_animation.size() unless m_animations.empty()
+        for ( auto boneAnim : m_animations[m_animationID] ) {
```
MathiasPaulin, line 172:
```suggestion
        for ( const auto& boneAnim : m_animations[m_animationID] ) {
```
There are several loops on this file that should benefit from this king a fixes. Use const ref whenever it is the meaning of the loop variable.

MathiasPaulin (commented):
This example is welcome but still need some work before being accepted. 
  - continue to update SkeletonComponent to make loops more efficient and robust
  - add some comments in the example toserve as reference to the documentation on how to build a skeleton

src/Core/Geometry/MeshPrimitives.hpp
dlyr: My idea is to remove the other makeCylinder, but i don't know if it's usefull for someone, any guess ?

src/Engine/Scene/SkeletonComponent.cpp
dlyr: maybe out of scope of this pr, but ok for this one

examples/SimpleSkinning/main.cpp
```diff
@@ -0,0 +1,147 @@
...
+    auto entity = engine->getEntityManager()->createEntity( "Cylinder" );
+
+    auto cylinder =
+        Geometry::makeSharpCylinder( { 0_ra, 0_ra, 0_ra }, { 0_ra, 0_ra, 10_ra }, 2_ra, 16, 32 );
+
+    // component ownership is transfered to entity in component ctor
+    auto meshComponent =
+        new TriangleMeshComponent( "Cylinder", entity, std::move( cylinder ), nullptr );
+
+    std::map<std::string, Core::Transform> boneMatrices;
```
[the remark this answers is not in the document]
dlyr: sure I'll add some comment, but I was not sure the way I've made the example is fine. (is there any other way to define skinning without input files, e.g with skinning system ?)

src/Core/Geometry/MeshPrimitives.hpp
MathiasPaulin: As your proposal covers also the previous makeCylinder (by setting `fillSegments` to 1), yes, your proposal should replace the previous one with the same name.

dlyr: Well the main difference is that the old one have the same normals for side and fill segment around sides. The new one have duplicated vertices with a sharp edge around side.

src/Engine/Scene/SkinningComponent.hpp
```diff
@@ -135,9 +139,16 @@ class RA_ENGINE_API SkinningComponent : public Component
     void setWeightBone( uint bone );
     /// \}
 
+    /// Set the associated weight for named bones.
+    /// Each pair is a vertex index from the skinned mesh (uint) and a weight (Scalar).
     void setPerBoneWeight( std::map<std::string, std::vector<std::pair<uint, Scalar>>>&& w ) {
         m_loadedWeights = w;
     }
+
+    /// Set the assiciated bind matrix of named bones.
```
MathiasPaulin, line 148:
```suggestion
    /// Set the associated bind matrix of named bones.
```

src/Engine/Scene/SkinningComponent.hpp
```diff
@@ -135,9 +139,16 @@ class RA_ENGINE_API SkinningComponent : public Component
...
+    /// A bind matrix is used to transform a vertex from object space to rest pose bone space.
+    /// Typical calue is obtained from skeleton getTransform( boneId, Skeleton::SpaceType::MODEL
```
MathiasPaulin, line 150:
```suggestion
    /// Typical value is obtained from skeleton getTransform( boneId, Skeleton::SpaceType::MODEL
```

src/Engine/Scene/SkeletonComponent.cpp
MathiasPaulin: When encountering some problems not so hard to fix nor so numerous, I prefer to fix them when discovered but it's just a matter of taste.

dlyr: Sure, feel free to do it, it's quicker to fix/commit it directly I think, than comment and ask for modification.
Unless you think we should discuss the proposed modification.
(btw I've done it for this file).

MathiasPaulin (approved):

dlyr:
I've tried to specialize template function for eigen matrix without success, for the record the pre non working implementation
```{.cpp}

/// Component-wise smoothstep() function on a floating-point vector.
/// \todo fix template resolution compared to scalar only version
/// https://www.fluentcpp.com/2017/08/11/how-to-do-partial-template-specialization-in-c/

template <typename DT, typename DU, typename DV>
inline typename DV::PlainMatrix
smoothstep<Eigen::MatrixBase<DT>, Eigen::MatrixBase<DU>, Eigen::MatrixBase<DV>>(
    const Eigen::MatrixBase<DT>& edge0,
    const Eigen::MatrixBase<DU>& edge1,
    const Eigen::MatrixBase<DV>& x ) {
    using std::clamp;
    typename DV::PlainMatrix t =
        clamp( ( x - edge0 ) / ( edge1 - edge0 ),
               DV::PlainMatrix::Constant( static_cast<typename DV::PlainMatrix::Scalar>( 0 ) ),
               DV::PlainMatrix::Constant( static_cast<typename DV::PlainMatrix::Scalar>( 1 ) ) );
    return t * t * ( 3.0 - 2.0 * t );
}


/// Component-wise smoothstep() function on a floating-point vector.
template <typename Derived>
inline typename Derived::PlainMatrix
smoothstep( const Scalar& edge0, const Scalar& edge1, const Eigen::MatrixBase<Derived>& v ) {
    return v.unaryExpr( [edge0, edge1]( Scalar x ) { return smoothstep( edge0, edge1, x ); } );
}


```

MathiasPaulin:
> I've tried to specialize template function for eigen matrix without success, for the record the pre non working implementation

Did you success ? I do not find the partial specialization for matrices in the PR.
By specializing, do you mean specializing
```
template <typename T, typename V>
V smoothstep( T edge0, U edge1, V x )
``` 
(I do not understand why `edge0` and `edge1` should be different type)
for the case `V = Eigen::MatrixBase<W>` for any W ?

did you try 
```
template <typename T, template<typename W> typename V>
V<W> smoothstep( T edge0, U edge1, V<W> x ) { ... }
``` 
Perhaps I'm wrong but this is  how I interpret the template template parameters
https://en.cppreference.com/w/cpp/language/template_parameters#:~:text=since%20C%2B%2B20)-,Template,-template%20parameter

dlyr:
It's not in the PR since I did not succeed. I've pasted my experiment here.
right we can restrict to the same type for edge0 and edge1.
The idea is to allow 
```
smoothstep(scalar edge0, scalar edge1, scalar x);
smoothstep(scalar edge0, scalar edge1, vector x);
smoothstep(vector edge0, scalar edge1, vector x);
```
as in glsl

I'll give a second try, but I think it's better to postpone to another pr.

dlyr:
let's postpone. 
last try was  
```
template <typename Derived>
inline typename Derived::PlainMatrix
smoothstep( const Scalar& edge0, const Scalar& edge1, const Eigen::MatrixBase<Derived>& v ) {
    return v.unaryExpr( [edge0, edge1]( Scalar x ) { return Ra::Core::Math::smoothstep( edge0, edge1, x ); } );
}
```
which gives
``` 
note: candidate template ignored: deduced conflicting types for parameter 'T' ('float' vs. 'Eigen::Matrix<float, 3, 1, 0, 3, 1>')
T smoothstep( T edge0, T edge1, T x ) {
  ^
```

src/Core/Math/Math.hpp
```diff
@@ -73,10 +73,10 @@ template <typename T>
 inline constexpr T lerp( const T& a, const T& b, Scalar t );
 
 /// As define by https://registry.khronos.org/OpenGL-Refpages/gl4/html/smoothstep.xhtml
-template <typename T, typename U, typename V>
-V smoothstep( T edge0, U edge1, V x ) {
+template <typename T>
```
MathiasPaulin, line 76: Why removing 2 template parameters ? Should have one for edges, one for the data. This last might be the same than the first, or not.

MathiasPaulin (changes requested):

src/Core/Math/Math.hpp
dlyr: the current implementation I propose could not handle smoostep( scalar, scalar, vector ) nor smoostep(vector, vector, vector) due to operation and cast undefined.
One way to handle this is to have a template specialization, if you have a possible implementation in mind, feel free to test. 

So to ensure the smoothstep works as expected, let's say edge and x are the same time for the moment.

MathiasPaulin (approved):

src/Core/Math/Math.hpp
MathiasPaulin: I'll give a try ASAP (say, 1 or 2 days).

MathiasPaulin: well, compile and works using llvm clang++ 11.1.0
```
template <typename T, template<typename, int... > typename M, int... p>
M<T, p...> smoothstep( T edge0, T edge1, M<T, p...> v ) {
    return v.unaryExpr( [edge0, edge1]( T x ) { return Ra::Core::Math::smoothstep( edge0, edge1, x ); } );
}
```

tested with : 
```
Vector3 v {0_ra, 0_ra, 0_ra};
std::cout << smoothstep(-1_ra, 1_ra, v ).transpose() << "\n"; 
```
this results in `0.5 0.5 0.5` (expected)

and 
```
Matrix2 m;
m << 0_ra, 0_ra, 0_ra, 0_ra;
std::cout << smoothstep(-1_ra, 1_ra, m ) << "\n";
```
this results in (expected also)
```
0.5 0.5
0.5 0.5
```

MathiasPaulin: I can push my proposal.

MathiasPaulin: pushed

PR #1027, opened 2022-11-25
Renderparameter uses Core::VariableSet as storage for parameters

MathiasPaulin:
# Pull Request Description

This PR changes the way render parameters are stored in class RenderParameters.
only small changes in public interface was made. Only changes in private interface.

This PR also  solve some design issues in   VariableSet that prevent  usage on Windows. The current  version works on all systems. 

Todo
  - [x] Fix windows compilation
  - [x] Make it work on windows.
  - [x] Implement ParameterSetEditor::setupFromParameters using a custom visitor 
  - [x] Test this proposal for real applications (only tested with unittests, integration tests and by running the examples with custom extension plugins)

  - [ ] bug fix
  - [x] feature
  - [ ] docs update
  - [ ] other:

RenderParameters uses explicit storage for all parameter types. This cause some redundant code and make extending the supported parameter types difficult

RenderParameters use the Core::VariableSet generic storage facilities and is, right now, only a wrapper to VariableSet.
This changes make code simpler to maintain (Well, not sure for the current version).

Only private interface changes but with minor changes to public interface (make some attributes private to ensure invariants) 

- **Other information**:

The current state is working on all the example apps, tests, and is also tested with external library (gltf).


src/Core/Utils/EnumConverter.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * \brief This class manage the bijective association between string and integral representation
+ * of an enumeration.
+ *
+ * \tparam EnumBaseType the underlying type (as returned by typename std::underlying_type_t<Enum>)
+ * of the enumeration to manage
+ *
+ * Management of string representation of enum type.
+ * \todo, reword the following
```
nmellado, line 20: Just in case it was missed: a todo is remaining

src/Core/Utils/TypesUtils.hpp
```diff
@@ -23,8 +25,121 @@ const char* demangleType() noexcept;
 template <typename T>
 const char* demangleType( const T& ) noexcept;
 
+// TypeList taken and adapted from
+// https://github.com/AcademySoftwareFoundation/openvdb/blob/master/openvdb/openvdb/TypeList.h
+// Only took small part of TypeList utilities
+// todo : do we need all the type list utilities ? I think yes but wait for other advices
```
nmellado, line 31: Another todo here: I guess we converged, and this comment can be removed ?

src/Engine/Rendering/ForwardRenderer.cpp
```diff
@@ -404,42 +410,50 @@ void ForwardRenderer::renderInternal( const Data::ViewingParameters& renderData
...
-            auto shader = m_shaderProgramManager->getShaderProgram( "ComposeVolume" );
-            shader->bind();
-            shader->setUniform( "volumeImage", m_textures[RendererTextures_Volume].get(), 0 );
-            m_quadMesh->render( shader );
+                for ( const auto& ro : m_volumetricRenderObjects ) {
+                    ro->render(
+                        passParams, renderData, DefaultRenderingPasses::LIGHTING_VOLUMETRIC );
+                }
+            }
+            // TODO make this more simpler
```
nmellado, line 435: Another todo here

nmellado (commented):
I spotted a few remaining `\todo` commands, otherwise looks ready to merge.

src/Engine/Scene/Light.hpp
```diff
@@ -103,14 +102,35 @@ class RA_ENGINE_API Light : public Component
...
+    m_color = color;
+    m_params.addParameter( "light.color", m_color );
+}
+
+inline const Light::LightType& Light::getType() const {
+    return m_type;
+}
+
+inline Data::RenderParameters& Light::getRenderParameters() const {
+    return const_cast<Data::RenderParameters&>( m_params );
```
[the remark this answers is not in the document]
MathiasPaulin: This is because RenderParameters only automatically binds references to other RenderParameters (and not const references, see `Ra::Engine::Data::RenderParameters::BindableTypes`. ) and some shaders (for volume rendering in ForwardRenderer but also others in external applications, extension libraries or plugins) need the parameters of the current light to be bound when rendering, without rendering a RenderObjects that manage the binding explicitly. 
So, either we return a non-const reference here, or we add automatic binding of const reference in the RenderParameters class. I'll try the second to check the impact of this choice and will push my proposal ASAP.

dlyr (approved):

PR #1037, opened 2023-01-22
Remove inlines

dlyr:
Remove `#include "foo.inl"` from `foo.hpp` and replace it by the inl content, using `noinl.py` script, run from `src/` directory.

Also remove inline references from `CMakeLists.txt` and scripts.

If new pushes on release-candidate, skip commit `run noinl.py` and `remove inl` to redo it on a clean basis (e.g. run the script and git rm inl files)

TODO
- [x] update doc to remove any ref to inl files 
- [x] move includes from inl to the top of hpp files 
- [x] merge namespaces


doc/developer/conventions.md
```diff
@@ -74,7 +75,6 @@ use `Scalar()` or `_ra` suffix when defining numbers from literals (e.g. `auto a
 
 * Constructors should be trival. All complex work goes in an `init()` function
```
MathiasPaulin, line 76: Out of scope of this PR but this should be avoided (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c41-a-constructor-should-create-a-fully-initialized-object)

MathiasPaulin (approved):

doc/developer/conventions.md
dlyr: Out of scope right, maybe simply do another pr

Issue #1039, opened 2023-02-02
Wrong viewport on HDPI displays on windows when the configured display scale is not 100%


MathiasPaulin:
## Expected Behavior

Rendering on a Radium SimpleWindow should fill the viewport with the computed image.

## Current Behavior

As seen on the picture below, rendering on hdpi display on windows is not correct. The viewport is not filled as expected.
![image](https://user-images.githubusercontent.com/11328578/216298458-65f2bffb-4cff-4a5a-a81b-1f7e80d81c30.png)

## Possible Solution

I've no idea ... any advice ?

## Steps to Reproduce (for bugs)

1. get windows with HDPI display, e.g. 3840x2160
2. in the display settings, configure Scale and Layout to 150% (recommended on my display so that text is readable)
3. launch any Radium app, you then get the above result
4. without quitting the app, in the display settings, configure Scale and Layout to 100% (then making text hard to read and everything very small, even on a 32"monitor) and you get the good result as shown below

![image](https://user-images.githubusercontent.com/11328578/216300664-b5977485-8656-47a8-a617-465df5fa571e.png)

PR #1040, opened 2023-02-03
Fix hdpi displays render size

dlyr:
# Pull Request Desription

  - [x] bug fix
  - [ ] feature
  - [ ] docs update
  - [ ] other:

see #1039 

  Takes into account device pixel ratio on any system so that OpenGL rendering and picking works with device (physical) coordinates. 
  tested on : 
  - [x] macos
  - [x] linux
  - [x] windows

Event management not related to openGL rendering  still remains based on Qt logical coordinates.

- **Other information**:

Gizmo manipulation computations remain in logical coordinate, even if the openGL part of the Gizmos relies on physical coordinates. 

The "scale" of the display (and then the devicePixelRatio) could be changed dynamically, without restarting the app. If two displays with different dpi are used, moving the app from one display to the other also get the right devicePixelRatio.


src/Gui/Viewer/Viewer.cpp
```diff
@@ -517,8 +517,8 @@ bool Viewer::initializeGL() {
 }
 
 void Viewer::resizeGL( QResizeEvent* event ) {
-    int width  = event->size().width();
-    int height = event->size().height();
+    int width  = event->size().width() * devicePixelRatio();
+    int height = event->size().height() * devicePixelRatio();
```
MathiasPaulin, line 521: Thanks, I'll give a try. I remember having tried something like this at the beginning of Radium dev but with some difficulties to take devicePixelRatio into account for picking and other pixel access.

dlyr: I've not tested picking nor anything else, but definitively need to be tested before merge.

src/Gui/Viewer/Viewer.cpp
```diff
@@ -517,8 +517,8 @@ bool Viewer::initializeGL() {
 }
 
 void Viewer::resizeGL( QResizeEvent* event ) {
-    int width  = event->size().width();
-    int height = event->size().height();
+    int width  = event->size().width() * devicePixelRatio();
```
MathiasPaulin, line 520:
```suggestion
    int width  = event->size().width();
```

src/Gui/Viewer/Viewer.cpp
MathiasPaulin, line 521:
```suggestion
    int height = event->size().height();
```

src/Gui/Viewer/Viewer.cpp
```diff
@@ -517,8 +517,8 @@ bool Viewer::initializeGL() {
...
+    int height = event->size().height() * devicePixelRatio();
     // Renderer should have been locked by previous events.
 #ifndef OS_MACOS
     gl::glViewport( 0, 0, width, height );
```
MathiasPaulin, line 524:
```suggestion
    gl::glViewport( 0, 0, width * devicePixelRatio(), height * devicePixelRatio() );
```

src/Gui/Viewer/Viewer.cpp
MathiasPaulin: Tested on windows. If this solve the viewport problem, as assumed, there are problems when picking :(. BTW, my suggestions here seems to work (at least for picking). I've tested it on windows and this solves my problem. My suggestion does not impact macos but needs to be tested on Linux.

MathiasPaulin: Suggestions tested on Linux/gnome, with hdpi display. Due to the way scaling seems to be managed by gnome, the suggestions do not impact Linux ( `devicePixelRatio()` seems to always return 1).

So, the solution is indeed simple, thanks @dlyr, but should be placed where I suggest (not in the computation of width and height but in the viewport definition). Once done, the I'll approve the PR.

dlyr: On my setup,  devicePixelRatio() returns 1 for 100% and 2 of 200% following gnome display setup. 
Changing only viewport make the rendering to be done as lower resolution, is this what we want ?
Changing viewport only
100%
view-100
200%
view-200
Changing dim
100%
dim-100
200%
dim-200

```{.bash}
$lsb_release -a
Description:	Debian GNU/Linux 11 (bullseye)
$ qmake6 -v
Using Qt version 6.3.1 in /usr/lib/x86_64-linux-gnu
```

MathiasPaulin: on my setting, `devicePixelRatio()` always return 1
scale100
scale200
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
$ qmake -v
QMake version 3.1
Using Qt version 5.15.2 in /usr/lib/x86_64-linux-gnu
```

If you execute the demo `Picking`, do you have the correct result with your 200% scale setting ? On windows (and Qt6.3.1), I haven't when changing dim, but it was OK when changing viewport (scaling was set at 150% as recommended by windows - and accepted by my eyes ^^ - and I didn't notice quality degradation).
I'll try to install Qt6 on linux to test with the same version than yours.

MathiasPaulin: After upgrading Qt version :
```
$ qmake6 --version
QMake version 3.1
Using Qt version 6.4.2 in /usr/lib/x86_64-linux-gnu
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
```

at 100% scale : `devicePixelRatio()` returns 1.25 (in tweakui, I've set the font to 125% to be able to read my screen)
original (without taking `devicePixelRatio` into account) - picking NOT OK: 
qt6-scale100-nopixelratio

viewport scaled by devicePixelRatio - picking OK
qt6-scale100-viewportscaled

at 200% scale : `devicePixelRatio()` returns 2.5
viewport scaled by devicePixelRatio - picking OK - quality degradation
qt6-scale200-viewportscaled

dim scaled by devicePixelRatio - picking NOT OK - quality OK
qt6-scale200-dimscaled

So, either we change all the picking management code to take pixelRatio into account, either we just scale the viewport, with a quality degradation acceptable for small scale (e.g 125%)

BTW, changing the dim to prevent quality lost (and changing the way picking is managed), seems strange to me as, e.g., for a scaling at 125%, we request a picture with size 800x600 and get one with size 1000x750 ... could be strange for some apps ...

MathiasPaulin:
I'll propose a fix soon.

MathiasPaulin (approved):

Issue #1012, opened 2022-10-17
Refactor Texture


dlyr:
Current Engine/Data/Texture implementation have some flaws.
This issue is to track/discuss solution, in continuation of the modification proposed in #982 

Concurrent multi-thread access might cause data inconsistency, between texture parameters and texture data.
Current implementation use texture parameters for multiples usage, and is a bit hacky.
Textures do not own its data, but may represent gpu data that was available on cpu side, but has been deleted afterward.

Proposed guidelines:
split sampler parameters and image data 
make texture the owner of image data
then methods can use class own mutex, be re-entrant...

Issue #1043, opened 2023-02-25
Maybe an issue in stask queue


dlyr:
Sometimes task queue unit test fail
i.e. https://github.com/STORM-IRIT/Radium-Engine/actions/runs/4269573806/jobs/7432847039

I couldn't reproduce, but there may be a bug somewhere.

PR #1056, opened 2023-07-02
Refactor textures

dlyr:
Update: some changes have been moved to specific PR, see #1072, #1073, #1074
## TODOS

- [x] Move MaterialTextureSet to its own header file
- [x] Clean history
- [x] test on bigger apps
- [x] update sandbox
- [x] Update docs

## description 

  - [x] feature

    See #1012 

    Start splitting texture image and sampler parameter

  - Use shared_ptr instead of naked new for texture data
  - Add texture sematics enum and MaterialTextureSet to manage material's texture in a uniform way
  - Delayed gpu texture init, remove material pending texture mechanism

- **Other information**:


src/Engine/Data/Texture.hpp
```diff
@@ -56,28 +75,24 @@ struct TextureParameters {
...
-    /// texels[3] <-- GL_TEXTURE_CUBE_MAP_NEGATIVE_Y <br/>
-    /// texels[4] <-- GL_TEXTURE_CUBE_MAP_POSITIVE_Z <br/>
-    /// texels[5] <-- GL_TEXTURE_CUBE_MAP_NEGATIVE_Z <br/>
-    /// @todo memory allocated for this pointer might be lost at texture deletion as ownership is
-    /// unclear.
-    void* texels { nullptr };
+
+    /// texels OR cubeMap, shared ownership
+    std::shared_ptr<void> texels { nullptr };
+    std::array<std::shared_ptr<void>, 6> cubeMap {};
```
MathiasPaulin, line 81: Why not using 
```suggestion
    /// texels OR cubeMap, shared ownership
    std::variant<std::shared_ptr<void>, std::array<std::shared_ptr<void>, 6> > texels;
```
This should simplify code below (variant has operators =, ==, ...) and make the access easier (always using texels). Validity of an ImageParameters could be checked with `texels.index() == (target != GL_TEXTURE_2D)`.

src/Engine/Data/Texture.hpp
```diff
@@ -215,28 +219,46 @@ class RA_ENGINE_API Texture final
...
-     * sufficient to update the GPU representation.
+     * If imageParameters is changed, the method call setImageParameters() to register update GPU
+     * representation task. If samplerParameter is change, the method call setSamplerParameters() to
+     * register update GPU sample task.
      */
     void setParameters( const TextureParameters& textureParameters ) {
+        m_updateMutex.lock();
+        bool test1 = ( textureParameters.sampler != m_textureParameters.sampler );
+        m_updateMutex.unlock();
+        if ( test1 ) setSamplerParameters( textureParameters.sampler );
```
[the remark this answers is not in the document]
dlyr: Actually setSamplerParameters need a free to lock m_updateMutex, hence this code.
I'll add a comment to clarify.

src/Engine/Data/Texture.hpp
dlyr: Done, I let you resolve the conversation if the proposed modification is fine for you.

src/Engine/Data/Texture.hpp
```diff
@@ -16,32 +17,32 @@ class Texture;
...
+    /// OpenGL minification filter ( GL_LINEAR or GL_NEAREST or GL_XXX_MIPMAP_YYY )
+    GLenum minFilter { GL_LINEAR };
+    /// OpenGL magnification filter ( GL_LINEAR or GL_NEAREST )
+    GLenum magFilter { GL_LINEAR };
+};
+
+struct ImageParameters {
+
+    using Image2DType = std::shared_ptr<void>;
+    using CubeMapType = std::array<std::shared_ptr<void>, 6>;
```
MathiasPaulin, line 37: Should we add type alias for `Image3D` (3D texture) and `ImageLayered` (layered texture) ?
If yes, also add getter specializations.

src/Engine/Data/VolumetricMaterial.hpp
```diff
@@ -94,6 +94,7 @@ class RA_ENGINE_API VolumetricMaterial final : public Material, public Parameter
 
   private:
     // The density matrix
+    /// \todo get rid of the raw ptr.
```
MathiasPaulin, line 97: Should we use same logic than for other materials. i.e. inheriting from `public MaterialTextureSet<TextureSemantics::VolumetricMaterial>` ?

src/Engine/Data/Texture.hpp
MathiasPaulin: OK, some comments in my last review (07-13-2023).

src/Engine/Data/VolumetricMaterial.hpp
dlyr: If I understand well, Volumetric Material has only one texture and manage it.
So in it's current form, using TextureSet is overkill.

src/Engine/Data/Texture.hpp
```diff
@@ -16,32 +17,32 @@ class Texture;
...
+    const Image2DType* getImage2D() const {
+        return std::get_if<ImageParameters::Image2DType>( &texels );
+    }
+
+    const CubeMapType* getCubeMap() const {
+        return std::get_if<ImageParameters::CubeMapType>( &texels );
+    }
```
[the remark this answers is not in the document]
dlyr: Right, the current verion is a first step (trying to change texture implem, keeping the interface), and will be adapted.
I'll prefer on my side trying to avoid exception as much as possible, but the proposed interface is ok.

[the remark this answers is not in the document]
dlyr: I'll prefer on my side trying to avoid exception as much as possible, but the proposed interface is ok.
We can use std::optional.
Why the const * to shared is dangerous ?

src/Engine/Data/Material.hpp
```diff
@@ -5,15 +5,64 @@
 #include <map>
 #include <string>
 
-// This include brings only the macro EIGEN_MAKE_ALIGNED_OPERATOR_NEW in the file scope.
+// This include brings only the macro  in the file scope.
 // Need to be separated to reduce compilation time
 #include <Core/Types.hpp>
```
MathiasPaulin, line 10:
```suggestion

```

src/Engine/Data/VolumetricMaterial.hpp
MathiasPaulin: Right, this was due to the old way textures were managed and loaded. Here, the 3D texture is loaded as part of the Asset::VolumeData and converted to VolumetricMaterial by the VolumeComponent. But I think that unifying material interface should help even if this specific part can be delayed for a future PR.

src/Engine/Data/Texture.hpp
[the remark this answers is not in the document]
dlyr: Done, feel free to reopen if needed

src/Engine/Data/Texture.hpp
dlyr: Done, feel free to reopen if needed

src/Engine/Data/VolumetricMaterial.hpp
dlyr: Let's delay this for a futur PR.

src/Engine/Data/Texture.cpp
```diff
@@ -57,278 +154,209 @@ void Texture::initializeGL( bool linearize ) {
...
 }
 
 // let the compiler warn about case fallthrough
-void Texture::updateParameters() {
+void Texture::sendSamplerParametersToGpu() {
     switch ( m_texture->target() ) {
     case GL_TEXTURE_CUBE_MAP:
     case GL_TEXTURE_3D:
-        m_texture->setParameter( GL_TEXTURE_WRAP_R, m_textureParameters.wrapP );
+        m_texture->setParameter( GL_TEXTURE_WRAP_R, m_textureParameters.sampler.wrapP );
```
MathiasPaulin, line 302: Why `wrapP` in the sampler and not `wrapR` to match OpenGL naming ?

src/Engine/Data/Texture.hpp
```diff
@@ -16,195 +17,181 @@ class Texture;
...
+    /// Types for texels variant
+    using ImageType   = std::shared_ptr<void>;
+    using CubeMapType = std::array<std::shared_ptr<void>, 6>;
+
+    /// check which type is held by texels
+    template <typename TexelType>
+    bool isTexelOfType() const {
+        return std::holds_alternative<TexelType>( texels );
+    }
+
```
MathiasPaulin, line 44: I think the two following functions could be helpful
```suggestion
template<typename ContentType>
ContentType getContent() const {
    CORE_ASSERT( isTexelOfType<ContentType>(), "texture variant does not hold the requested type." );
    return std::get<ContentType>( texels );
}

template<typename ContentType>
void setContent( ContentType image ) { texels = image; }
```
and then {get|set}Image and {get|set}CubeMap should be simple wrapper (syntactic sugar) over these two.
This should enable client to manage texture in a templated way without testing explicitly the type of the texture.

dlyr: Since data is public, why don't just use
`std::get<ContentType>(imageParam.texels)`
?

dlyr: Let's delay to a use case, if needed we can add those get / set.

MathiasPaulin (dismissed):

dlyr:
I don't get why there is some freeze on texture unittest, I could reproduce it on my computer, randomly happens.
Turn into draft until not solved.

dlyr:
I have fixed the task queue problem (we already had some times ago on some tests  #1043).
To this end I have take inspiratoin from https://github.com/Liam0205/toy-threadpool/
With one mutex only, read and write lock.
Indeed sometimes threads start to work after dtor call, and some write/read inconsistency. 
The fix might be overkill since I have done a lot of tests to find what happened, a second analysis might simplify the code

src/Engine/Data/BlinnPhongMaterial.hpp
```diff
@@ -2,12 +2,14 @@
...
-#include <map>
-#include <string>
-
 #include <Core/Utils/Color.hpp>
 #include <Engine/Data/Material.hpp>
 #include <Engine/Data/Texture.hpp>
+#include <Engine/Data/TextureManager.hpp>
+#include <Engine/RadiumEngine.hpp>
+
+#include <map>
```
MathiasPaulin, line 11: not needed.

src/Engine/Data/SimpleMaterial.hpp
```diff
@@ -2,26 +2,39 @@
...
-#include <map>
-#include <string>
-
 #include <Core/Utils/Color.hpp>
 #include <Engine/Data/Material.hpp>
 #include <Engine/Data/Texture.hpp>
+#include <Engine/Data/TextureManager.hpp>
+#include <Engine/RadiumEngine.hpp>
+
+#include <map>
```
MathiasPaulin, line 11: not needed.

src/Engine/Data/Texture.hpp
```diff
@@ -16,230 +17,288 @@ class Texture;
...
-    /// texels[4] <-- GL_TEXTURE_CUBE_MAP_POSITIVE_Z <br/>
-    /// texels[5] <-- GL_TEXTURE_CUBE_MAP_NEGATIVE_Z <br/>
-    /// @todo memory allocated for this pointer might be lost at texture deletion as ownership is
-    /// unclear.
-    void* texels { nullptr };
 };
 
-/** Represent a Texture of the engine
+/// ImageParameters stores the infomation needed to upload a image to the GPU as a texture.
+/// The actual data is stored as a std::variant, so this image parameter can storke either a single
```
MathiasPaulin, line 36:
```suggestion
/// The actual data is stored as a std::variant, so this image parameter can store either a single
```

src/Engine/Data/Texture.hpp
```diff
@@ -16,230 +17,288 @@ class Texture;
...
+/** @brief Describes the sampler and image of a texture.
+ *
+ * These parameters describe the image data of the texture :
+ *    - ImageParameters: target, width, height, depth, format, internalFormat, type and texels for
+ * describing image data
+ *    - SampleParameters: wrapS, wrapT, wrapP, minFilter and magFilter for describing the sampler of
+ * the texture.
+ *
+ * MipMap representation of the texture is automatically generated as soon as the minFilter
+ * parameter is something else than GL_LINEAR or GL_NEAREST
```
MathiasPaulin, line 93:
```suggestion
 * parameter is something else than GL_LINEAR or GL_NEAREST.
 * @todo allow to use another mipmap building function than glGenerateMipmap
```
Mipmap could be built in another way then just averaging on pixels as done by glGenerateMipmap. Delay this to a future use case (I've some) but remind here this future evolution.

src/Headless/CLIViewer.cpp
```diff
@@ -27,6 +27,9 @@ CLIViewer::CLIViewer( std::unique_ptr<OpenGLContext> context ) :
         ->check( CLI::ExistingFile );
     addOption( "-s,--size", m_parameters.m_size, "Size of the computed image." )->delimiter( 'x' );
     addFlag( "-a,--animation", m_parameters.m_animationEnable, "Enable Radium Animation system." );
+
+    m_tasks = std::make_unique<Core::TaskQueue>(
+        std::min( std::thread::hardware_concurrency(), 2u ) - 1u );
```
MathiasPaulin, line 32: Why std::min ? Should it be std::max ?

src/Headless/CLIViewer.hpp
```diff
@@ -70,11 +73,12 @@ class HEADLESS_API CLIViewer : public CLIBaseApplication
 
     /// is the window shown ?
     bool m_exposedWindow { false };
+    std::unique_ptr<Ra::Core::TaskQueue> m_tasks { nullptr };
```
MathiasPaulin, line 76: Add some comment to explain the need for this and is usage.

src/PluginBase/RadiumPluginInterface.hpp
```diff
@@ -146,7 +147,7 @@ class RadiumPluginInterface
      * MUST not be destroyed by the plugin
      */
     virtual void
-    addRenderers( std::vector<std::shared_ptr<Ra::Engine::Rendering::Renderer>>* /*rds*/ ) {}
+    addRenderers( std::vector<std::shared_ptr<Ra::Engine::Rendering::Renderer>>* rds ) {}
```
MathiasPaulin, line 150: If a parameter is not used in a function, the name should be omitted (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused). 
If the name is given but the parameter is not used, several compilers generates a warning. 
So, either name the parameter in a comment to remind it or do not name the parameter. 
I prefer here the former as this declaration introduce a virtual function that should be overridden by derived classes.

MathiasPaulin: So, we should use the attribute `[[maybe_unused]]`. 
i.e.
```c++
virtual void
    addRenderers( [[maybe_unused]] std::vector<std::shared_ptr<Ra::Engine::Rendering::Renderer>>* rds ) {}
```
This will remove the compilation warning while allowing to document the parameter in the current doxygen version.
I prefer to not remove any warning but write the code in a proper way if there is a way to do so.

src/Core/Tasks/TaskQueue.cpp
```diff
@@ -178,82 +184,92 @@ void TaskQueue::detectCycles() {
...
+        }
     }
-    while ( !m_taskQueue.empty() ) {
-        TaskId task { m_taskQueue.back() };
-        m_taskQueue.pop_back();
+    {
+        while ( !taskQueue.empty() ) {
+            TaskId task;
+            {
+                wlock lock( m_mutex );
```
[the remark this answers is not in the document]
MathiasPaulin: is the local variable `std::deque<TaskId> taskQueue;` shared among threads ? I don't think so. And if this is the case, why to not lock the `push_front` at line 230 ...

src/Core/Tasks/TaskQueue.cpp
```diff
@@ -178,82 +184,92 @@ void TaskQueue::detectCycles() {
...
-        m_tasks[task]->process();
-        m_timerData[task].end = Utils::Clock::now();
-
-        for ( auto t : m_dependencies[task] ) {
-            uint& nDepends = m_remainingDependencies[t];
-            CORE_ASSERT( nDepends > 0, "Inconsistency in dependencies" );
-            --nDepends;
-            if ( nDepends == 0 ) { queueTask( t ); }
+            for ( auto t : m_dependencies[task] ) {
+                wlock lock( m_mutex );
```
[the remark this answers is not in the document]
MathiasPaulin: `taskQueue` is still a local variable in the function (declared at line 216). So, not shared among threads but allocated on the stack in each thread. Are local variables declared in functions executed by several threads shared among those ? I don't think so but perhaps I missed something.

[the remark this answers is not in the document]
dlyr: My bad, I mixed up between runTask and runTasksInThisThread.
I'll fix it.
The local taskQueue is the most direct solution I find to prevent thread taking tasks during "this thread" run.

When I think about it, the solution might be to have two tasks lists, one for multi-thread, the other for this thread, and they might have inter dependencies. 
Hence GPU tasks in main thread and other tasks could be scheduled by the same taskQueue. Let's postpone for a future PR

src/PluginBase/RadiumPluginInterface.hpp
```diff
@@ -181,7 +182,7 @@ class RadiumPluginInterface
      * and realease it after usage (viewer->doneCurrent())
      * @see Ra::Gui::Viewer
      */
-    virtual void openGlInitialize( const Context& /*context*/ ) {}
+    virtual void openGlInitialize( const Context& context ) {}
```
MathiasPaulin, line 185:
```suggestion
    virtual void openGlInitialize( [[maybe_unused]] const Context& context ) {}
```

dlyr:
To ease conversation, rebase and force push will be done once accepted.

dlyr:
Since changes done after first acceptance are now accepted in #1070 and #1071, I'll rebase/force push this one (and merge fix commits)

MathiasPaulin (approved):

dlyr:
To ease rebase (since there were a lot of conflicts due to doxygen commands changes) I have merged bunch of commits, keeping all commits comments.

PR #1059, opened 2023-07-10
Small fixes

dlyr:
Small fixes, see commits.
Mostly warnings.


src/Core/Containers/VariableSet.hpp
```diff
@@ -513,7 +513,7 @@ class RA_CORE_API VariableSet
     /// Type of the constructed visit function that returns a lambda
 
     /// Shared VariableSetFunctions among VariableSet instances.
-    class VariableSetFunctions
+    class RA_CORE_API VariableSetFunctions
```
MathiasPaulin, line 516: I do not understand why to export this inner-class. As this is private implementation of the variable set functions management, exporting the symbols is not needed as the service is only used internally.

MathiasPaulin (approved):

PR #1067, opened 2023-10-06
Format new proposal

dlyr:
AFTER #1066 

Try slight changes on clang format configuration to have more oneline, and inline template.
Inline template are specifically usefull for using declaration

```{.cpp}
template<typname T> 
using type1 = T*;
template<typname T> 
using t2 = T**;
```
vs.

```{.cpp}
template<typname T> using type1 = T*;
template<typname TA> using t2   = TA**;
```

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,11 +1,11 @@
 ---
 BasedOnStyle: LLVM
 Language: Cpp
-Standard: Cpp11
+Standard: c++11
 AllowShortFunctionsOnASingleLine: Inline
 AllowShortBlocksOnASingleLine: true
 AllowShortIfStatementsOnASingleLine: true
-AlwaysBreakTemplateDeclarations: true
+AlwaysBreakTemplateDeclarations: Yes
 BreakBeforeBraces: Custom
 BraceWrapping:
   AfterClass: true
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,12 +26,12 @@ repos:
 #      - id: cppcheck
 #        args: [--enable=all]
 - repo: https://github.com/igorshubovych/markdownlint-cli
-  rev: v0.33.0
+  rev: v0.37.0
   hooks:
   - id: markdownlint
     args: [-f]
 - repo: https://github.com/pre-commit/mirrors-clang-format
-  rev: v15.0.6
+  rev: v16.0.6
   hooks:
   - id: clang-format
     types_or: [c++, c, cuda]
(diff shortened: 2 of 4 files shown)


src/Core/Animation/KeyFramedValue.hpp
```diff
@@ -58,8 +58,7 @@ class RA_CORE_API KeyFramedValueBase
  *
  * \note There is always at least one keyframe defined.
  */
-template <typename VALUE_TYPE>
```
nmellado, line 61: A titre perso je préfère garder la séparation entre le type template et le reste de la déclaration.

dlyr: For the record 
> "Personally, I prefer to keep the template type separate from the rest of the declaration."

Ok, my concern raises when I have some `using` def with `template`. But there is no way currently to have different behavior of the template break for that in clang-format. But fine to stick with the always break.

I'll split other proposed style changes to check if there meaningful.

```{.cpp}
    template <typename Port>
    using PortPtr = std::shared_ptr<Port>;
    template <typename Port>
    using PortRawPtr = typename PortPtr<Port>::element_type*;

    template <typename Type>
    using PortInPtr = PortPtr<PortIn<Type>>;
    template <typename Type>
    using PortInRawPtr = typename PortInPtr<Type>::element_type*;

    template <typename Type>
    using PortOutPtr = PortPtr<PortOut<Type>>;
    template <typename Type>
    using PortOutRawPtr = typename PortOutPtr<Type>::element_type*;
````
vs
```{.cpp}

    template <typename Port> using PortPtr    = std::shared_ptr<Port>;
    template <typename Port> using PortRawPtr = typename PortPtr<Port>::element_type*;

    template <typename Type> using PortInPtr    = PortPtr<PortIn<Type>>;
    template <typename Type> using PortInRawPtr = typename PortInPtr<Type>::element_type*;

    template <typename Type> using PortOutPtr    = PortPtr<PortOut<Type>>;
    template <typename Type> using PortOutRawPtr = typename PortOutPtr<Type>::element_type*;
```

dlyr: You can check 94deb27 concerning oneliner.

also I vote to switch to c++17 (or c++11, but not Cpp11 since "Cpp11 is a deprecated alias for Latest" [clang-format-doc](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#standard)

MathiasPaulin: I prefer :
```c++
template <typename Port>
using PortPtr = std::shared_ptr<Port>;

template <typename Port>
using PortRawPtr = typename PortPtr<Port>::element_type*;
```

dlyr:
I'have added clang-format update to #1066 , and I close this PR.

PR #1069, opened 2023-10-09
Render parameters variable set

dlyr:
# Pull Request Desription

RenderParameters had VariableSet as data member and defined aliases method.
This PR make RenderParameters inherit from VariableSet and remove duplicated code. 
Enum management is also moved to VariableSet. 

The objective is to allow the creation of Gui using VariableSet directly.
Gui automatic creation handle predefined types only (i.e. gui has to know which type are possibly present and how to display them). 
That's why enum can either be treated as integral (enum underlying types, automatically handled by gui if it take care of integral types), with enum <> string conversion for display, or either as "plain" enum, but it's up to the user to specify how to manage them for gui or other usage.

  - [x] feature

RenderParameters public interface is changed (and now is mostly VariableSet.

### DIFF
--- a/src/Core/Containers/VariableSet.hpp
+++ b/src/Core/Containers/VariableSet.hpp
@@ -170,6 +170,12 @@ class RA_CORE_API VariableSet
     template <typename T>
     auto getVariable( const std::string& name ) -> T&;
 
+    /// \todo add enable_if
+    template <typename T>
+    auto getEnumVariable( const std::string& name ) const -> const T&;
+    template <typename T>
+    auto getEnumVariable( const std::string& name ) -> T&;
+
     /// \brief get the handle on the variable with the given name
     /// \tparam T the type of the variable
     /// \param name the name of a variable
@@ -678,15 +684,19 @@ auto VariableSet::getVariable( const std::string& name ) -> T& {
 
 template <typename T>
 auto VariableSet::getVariable( const std::string& name ) const -> const T& {
-
-    if constexpr ( std::is_enum<T>::value ) {
-        // need to cast to take into account the way enums are managed
-        return reinterpret_cast<const T&>(
-            getVariable<typename std::underlying_type<T>::type>( name ) );
-    }
-    //    assert( existsVariable<T>( name ) );
-
     return getVariableHandle<T>( name )->second;
+}
+
+template <typename T>
+auto VariableSet::getEnumVariable( const std::string& name ) -> T& {
+    return const_cast<T&>( const_cast<const VariableSet*>( this )->getEnumVariable<T>( name ) );
+}
+
+template <typename T>
+auto VariableSet::getEnumVariable( const std::string& name ) const -> const T& {
+    static_assert( std::is_enum<T>::value );
+    return reinterpret_cast<const T&>(
+        getVariable<typename std::underlying_type<T>::type>( name ) );
 }
 
 template <typename T>
@@ -734,9 +744,6 @@ template <typename T>
 auto VariableSet::existsVariable( const std::string& name ) const
     -> Utils::optional<VariableHandle<T>> {
 
-    if constexpr ( std::is_enum<T>::value ) {
-        return existsVariable<typename std::underlying_type<T>::type>( name );
-    }
     if ( auto typeAccess = existsVariableType<T>(); typeAccess ) {
         auto itr = ( *typeAccess )->find( name );
         if ( itr != ( *typeAccess )->cend() ) { return itr; }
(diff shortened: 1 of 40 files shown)


src/Core/Containers/VariableSet.hpp
```diff
@@ -613,13 +675,25 @@ auto VariableSet::insertVariable( const std::string& name, const T& value )
...
-auto VariableSet::getVariable( const std::string& name ) const -> T& {
-    assert( existsVariable<T>( name ) );
+auto VariableSet::getVariable( const std::string& name ) -> T& {
+    return const_cast<T&>( const_cast<const VariableSet*>( this )->getVariable<T>( name ) );
+}
+
+template <typename T>
+auto VariableSet::getVariable( const std::string& name ) const -> const T& {
+
+    if constexpr ( std::is_enum<T>::value ) {
```
MathiasPaulin, line 685: This test should not be on this condition but on the existance of an enumConverter, then should not be constexpr. 
Testing on `is_enum` impose to store enumeration variable as values of the enumeration's underlying type and not as  typed enumeration. So, this propagates GLSL limitations about enumerations to VariableSet and I think this is not a good thing.
Enum management through EnumConverters, and the storage using underlying type is welcome for several operations (GLSL binding, edition gui, ... ) but should not be imposed as user should store enumeration without a converter and with the real enumeration type.

src/Core/Containers/VariableSet.hpp
```diff
@@ -662,6 +736,10 @@ bool VariableSet::deleteVariable( H& handle ) {
 template <typename T>
 auto VariableSet::existsVariable( const std::string& name ) const
     -> Utils::optional<VariableHandle<T>> {
+
+    if constexpr ( std::is_enum<T>::value ) {
```
MathiasPaulin, line 740: Why considering enum variable only through the enum underlying type ?

src/Core/Containers/VariableSet.hpp
```diff
@@ -739,6 +817,12 @@ auto VariableSet::addVariableType() -> Utils::optional<VariableContainer<T>*> {
 
 template <typename T>
 auto VariableSet::existsVariableType() const -> Utils::optional<VariableContainer<T>*> {
+    if constexpr ( std::is_enum<T>::value ) {
```
MathiasPaulin, line 820: Enum management depends on the usage of the enum, not on the VariableSet management. I think we do not have to impose some restrictions on the variable stored on VariableSet. We should definitly discuss about enum management as an integrated VariableSet service ...

MathiasPaulin (dismissed):
I've several concerns about this PR, mainly about the integration of GLSL or gui restriction into VariableSet (for enum management). This should be discussed deeper.

src/Core/Containers/VariableSet.hpp
dlyr: Ok, I'll think of an alternative.
A first guess is to store the enums directly in the VariableSet (always), and have a dynamic visitor that convert to unerlying types in the ShaderParameters class.
I let the conversation open until a solution is found.

dlyr: (I close the other related conversations to have a simpler workflow)

dlyr: > Enum management depends on the usage of the enum, not on the VariableSet management. I think we do not have to impose some restrictions on the variable stored on VariableSet. We should definitly discuss about enum management as an integrated VariableSet service ...
> Why considering enum variable only through the enum underlying type ?

dlyr: new proposal
one can add/set/get plain enum with [insert/set/get}Variable
or set get underlying with [set/get]EnumVariable, this also support string setters thru enumConverters.
TODO
maybe change the name of the second set of method, and also add enumConverters to plain enum type.
add enable_if to ensure template type compatibility

dlyr: Another solution is to move enum management to funtion, in a specific namespace, e.g. `VariableSetEnumManagement`

Thu
`vs.setEnumVariable(...)` becomes `VariableSetEnumManagement::setEnumVariable(vs, ...)`

MathiasPaulin: I prefer this way that should allow both adding/retrieving an enum through its real type or through its underlying type but also, as you mentionned, to envision other helpers to extend VariableSetManagement without modifying VariableSet.

src/Core/Containers/VariableSetEnumManagement.hpp
```diff
@@ -0,0 +1,135 @@
...
+    // typed converter
+    auto converterHandle = vs.setVariable( name, converter );
+
+    // string string converter/setter for setEnumVariable
+    std::function<void( Core::VariableSet&, const std::string&, const std::string& )>
+        convertingFunction = [converter = converterHandle.first]( Core::VariableSet& vs,
+                                                                  const std::string& nm,
+                                                                  const std::string& vl ) {
+            vs.setVariable( nm, converter->second->getEnumerator( vl ) );
+        };
```
MathiasPaulin, line 101: This will remove a compilation warning about variable shadowing ...
```suggestion
    std::function<void( Core::VariableSet&, const std::string&, const std::string& )>
        convertingFunction = [converter = converterHandle.first]( Core::VariableSet& st,
                                                                  const std::string& nm,
                                                                  const std::string& vl ) {
            st.setVariable( nm, converter->second->getEnumerator( vl ) );
        };
```

MathiasPaulin (commented):
A few comments before resolving unittest on macos and windows

MathiasPaulin (approved):

PR #1070, opened 2023-10-12
Ci update doxygen

dlyr:
This PR update doxygen version, it also adds a new github action to test doc creation.

It fixes doxygen missing TOC in markdown doc.

This workflow might be run manually on branches and create an artifact with the produced doc.

I'm not sure it will be ok to be run on PR branch, but is should be, and also can be modified when needed with 
```{.yml}
on:
  pull_request:
    branches:
      - release-candidate
```

see https://build5nines.com/configuring-manual-triggers-in-github-actions-with-workflow_dispatch/
https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow

### DIFF
--- a/.github/workflows/deploy-doc.yml
+++ b/.github/workflows/deploy-doc.yml
@@ -39,10 +39,10 @@ jobs:
       - name: Install packages
         run : |
           sudo apt-get install graphviz
-      - name: Fetch recent doxygen (1.9.7)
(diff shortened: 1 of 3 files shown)


nmellado (approved):

dlyr:
As I suspected, the manual workflow is only available when merged on master. It might be callable on any branch to test doc of this branch, but not usefull to test new doxygen version.

Issue #1065, opened 2023-10-05
Concurrent GPU / multithread tasks.


dlyr:
My bad, I mixed up between runTask and runTasksInThisThread.
I'll fix it.
The local taskQueue is the most direct solution I find to prevent thread taking tasks during "this thread" run.

When I think about it, the solution might be to have two tasks lists, one for multi-thread, the other for this thread, and they might have inter dependencies. 
Hence GPU tasks in main thread and other tasks could be scheduled by the same taskQueue. Let's postpone for a future PR

_Originally posted by @dlyr in https://github.com/STORM-IRIT/Radium-Engine/pull/1056#discussion_r1347532722_

PR #1071, opened 2023-10-12
Fix tasks

dlyr:
# Pull Request Desription

Split from #1056 to ease review.

# bug fix

# What is the current behavior? (You can also link to an open issue here)

#1043 

see #1056 for some discussion

### DIFF
--- a/src/Core/Tasks/TaskQueue.cpp
+++ b/src/Core/Tasks/TaskQueue.cpp
@@ -258,16 +258,9 @@ void TaskQueue::runTasksInThisThread() {
 }
 
 void TaskQueue::waitForTasks() {
-    bool isFinished = false;
-    while ( !isFinished ) {
-
-        // TODO : use a notifier for task queue empty.
-        {
-            rlock lock( m_mutex );
-            isFinished = ( m_taskQueue.empty() && m_processingTasks == 0 );
-        }
-        if ( !isFinished ) { std::this_thread::yield(); }
-    }
+    rlock lock( m_mutex );
+    m_waitForTasksNotifier.wait(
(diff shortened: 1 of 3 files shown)


src/Core/Tasks/TaskQueue.cpp
```diff
@@ -178,82 +184,91 @@ void TaskQueue::detectCycles() {
...
 
     // Do a debug check
     detectCycles();
-
-    // Enqueue all tasks with no dependencies.
-    for ( uint t = 0; t < m_tasks.size(); ++t ) {
-        // only queue non null m_tasks
-        if ( m_tasks[t] && m_remainingDependencies[t] == 0 ) { queueTask( TaskId { t } ); }
+    {
+        // Enqueue all tasks with no dependencies.
```
nmellado, line 227: Don't we need a `rlock` here ?

MathiasPaulin: I don't think so. See previous discussion about `TaskQueue::runTasksInThisThread()` on the PR [Refactor textures](https://github.com/STORM-IRIT/Radium-Engine/pull/1056#top) #1056

MathiasPaulin (approved):

src/Core/Tasks/TaskQueue.cpp
dlyr: Right, there is the comment atop 
 // this method should not be called between startTasks/waitForTasks, we do not lock anything
    // here.
 
This method also calls  resolveDependencies();  and detectCycles(); which locks, so we can't lock at the beginning of the function.

rlock is an option, but do not mean anything there. A futur version may resolve concurrent/mono thread better, follow up issue #1065

PR #1073, opened 2023-10-14
[headless] Fix camera possible leak.

dlyr:
# Pull Request Desription

## PR details

  - [x] bug fix

Neked new is used for camera in CLI viewer, use unique_ptr instead.
Also fix 0 thread queue.

### DIFF
--- a/src/Headless/CLIViewer.cpp
+++ b/src/Headless/CLIViewer.cpp
@@ -100,7 +100,8 @@ int CLIViewer::oneFrame( float timeStep ) {
         m_engine->step();
     }
 
-    Ra::Core::TaskQueue tasks( std::thread::hardware_concurrency() - 1 );
+    Core::TaskQueue tasks { std::max( std::thread::hardware_concurrency(), 2u ) - 1u };
+
     m_engine->getTasks( &tasks, Scalar( timeStep ) );
     tasks.startTasks();
     tasks.waitForTasks();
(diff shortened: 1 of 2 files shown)


src/Headless/CLIViewer.cpp
```diff
@@ -163,7 +164,9 @@ void CLIViewer::setCamera( Ra::Core::Utils::Index camIdx ) {
         Ra::Engine::RadiumEngine::getInstance()->getSystem( "DefaultCameraManager" ) );
 
     if ( camIdx.isInvalid() || cameraManager->count() <= size_t( camIdx ) ) {
-        m_camera = new Ra::Core::Asset::Camera( m_parameters.m_size[0], m_parameters.m_size[1] );
+        m_ownCamera = std::make_unique<Ra::Core::Asset::Camera>( m_parameters.m_size[0],
```
MathiasPaulin, line 167: Why not adding the created camera to the  `CameraManager` that manage ownership of cameras ? Even if there are some deficiencies in the `CameraManager` interface and the way it is used (whose fix is out of scope of this PR), this could be done as follow : 
```c++
{
        auto e      = m_engine->getEntityManager()->createEntity( "Default Camera" );
        auto camera = new Scene::CameraComponent( e, "Camera" );
        camera->initialize();
        cameraManager->addCamera( camera );
        cameraManager->activate( cameraManager->getCameraIndex( camera ) );
        m_camera = cameraManager->getActiveCamera();
        m_camera->setFOV( 60.0_ra * Ra::Core::Math::toRad );
        /// the same lines as below  to initialize the camera ...
        m_camera->setZFar( zfar );
    }
```
The method should then return the camera index of the active (the one created here) camera so that client can use it in other call to this method.

src/Headless/CLIViewer.hpp
```diff
@@ -61,6 +61,7 @@ class HEADLESS_API CLIViewer : public CLIBaseApplication
 
     /// The camera for rendering
     Ra::Core::Asset::Camera* m_camera { nullptr };
+    std::unique_ptr<Ra::Core::Asset::Camera> m_ownCamera { nullptr };
```
MathiasPaulin, line 64: Not needed if m_camera is fetched from the CameraManager.

src/Headless/CLIViewer.cpp
dlyr: Maybe, since I'm not sure what this naked new is supposed to do and why it arrives here, my solution is to make it smart, but another solution is surely better. 
Commit/fix welcome.

MathiasPaulin: OK, I'll do that ASAP

nmellado (approved):

PR #1091, opened 2025-03-26
Radium dataflow dlyr

dlyr:
Replace #1007 

# Pull Request Desription
Add dataflow programming capabilities to the Radium engine (draft)

TODO
 - [x] clarify ownership transfer of the nodes to the graph 
 - [x] clearly state that naked pointers got from the graph are non owning pointers.
 - [x] add documentation and programming guide (WIP)
 - [x] transform demos to unittests


src/Core/Containers/DynamicVisitor.hpp
```diff
@@ -0,0 +1,163 @@
...
+                     f( name, p );
+                 } };
+    }
+};
+
+template <class T, class F>
+inline auto DynamicVisitor::makeVisitorOperator( F& f ) -> OperatorsStorageType::value_type {
+    auto opBuilder = MakeVisitOperatorHelper < T, F,
+         std::is_invocable<F, const std::string&, T&, std::any&&>::value ||
+             std::is_invocable<F, const std::string&, T&, std::any&&>::value > {};
```
MathiasPaulin, line 158: it seems the two arguments of the `||` operator are the same.

src/Dataflow/Core/Node.hpp
```diff
@@ -0,0 +1,546 @@
...
+     *
+     * \param type either "in" or "out", the directional type of the port
+     * \param name
+     * \return the index to access the port and a raw ptr to the port.
+     */
+    auto port_by_name( const std::string& type, const std::string& name ) const
+        -> IndexAndPort<PortBaseRawPtr>;
+
+    /// Convenience alias to port_by_name("in", name)
+    IndexAndPort<PortBaseInRawPtr> input_by_name( const std::string& name ) const;
```
MathiasPaulin, line 148: Should we choose only one type of function declaration, i.e `auto f(...)-> returnType;` or `returnType f(...);`. I vote the first but agree with the other also.

src/Core/Containers/DynamicVisitor.hpp
dlyr: I don't remember how/why I added `&` to the first T, compiles and passes tests without it. It happens when I move this code to own source file. Since I'm not sure what is expected... is it
```suggestion
```
or
```suggestion
             std::is_invocable<F, const std::string&, T, std::any&&>::value > {};
```

src/Dataflow/Core/Node.hpp
dlyr: I think we can use auto for complex/templated/nested return value. and keep plain return for simple/std types. I'll change it.

src/Core/Containers/DynamicVisitor.hpp
MathiasPaulin: From line 891 of the actual src/Core/Containers/VariableSet.hpp, it is the second. This test for a callable with userParam(3rd parameter)  with the second parameter pass as value or as reference. This is to relate to the partial specialization of the struct `struct DynamicVisitor::MakeVisitOperatorHelper<T, F, [true | false]>` some lines above

dlyr: My bad, it's commit 75d69781d7c2a3c0ae84e8eb8e0db8dc17219611 which introduce this modification. 
1. Compiles and pass tests
```
std::is_invocable<F,` const std::string&, T&, std::any&&>::value
```

2. Does not compile
```
std::is_invocable<F,` const std::string&, T, std::any&&>::value
```

3. Compiles and pass tests.
```
std::is_invocable<F, const std::string&, T, std::any&&>::value ||
std::is_invocable<F, const std::string&, T&, std::any&&>::value
```


So it seems that we haven't a test that require non ref `T`. If such case is not needed, it's ok to remove the first part of the "or" (and go to case 1.). But if there is any case one need to have both type of functor, we need to add the corresponding test and go to case 3.

dlyr: I removed non ref `T` version and also moved to enable_if instead of helper struct.

MathiasPaulin (approved):
Works fine on Linux, not tested on macos nor windows

PR #1109, opened 2025-07-15
Build script

dlyr:
Add a script `build.sh` to ease workflow run, to be used in Radium-Apps and Radium-Release as well.
Highly configurable, I hope for most situation.

Also fixes workflow (windows build missed glfw due to bad path configuration).


nmellado (approved):
Did not test, but changes look ok

